PR #1, opened 2021-11-02
add CI and docs deployment

KristofferC:


KristofferC:
```
explicit zeros: Error During Test at /home/runner/work/SparseArrays.jl/SparseArrays.jl/test/sparse.jl:1648
  Test threw exception
  Expression: cholesky(a) \ [2.0, 3.0] ≈ [2.0, 3.0]
  MethodError: no method matching cholesky(::SparseArrays.SparseMatrixCSC{Float64, Int64})
```

Cholmod also does type piracy: https://github.com/JuliaLang/SuiteSparse.jl/blob/e4df734c3e0b54cd2275adbd923b5afaf0f7e3d0/src/cholmod.jl#L1170-L1176

:(

ViralBShah:
Those tests should just be moved to SuiteSparse.jl, and disabled in here for now.

ViralBShah:
The command line for running the CI doesn't seem to work on Windows.

ViralBShah:
Why are the ambiguity tests failing?

KristofferC:
Don't know. Seems to be the method in

https://github.com/JuliaLang/SparseArrays.jl/blob/c6b613c288b88d44acda7000dc56901228e38e2e/src/sparsevector.jl#L1125

All the concatenation stuff is filled with type piracy and is just a mess,

https://github.com/JuliaLang/SparseArrays.jl/blob/c6b613c288b88d44acda7000dc56901228e38e2e/src/sparsevector.jl#L1094-L1140

Like `vcat` on a vector is defined in SparseArrays!

https://github.com/JuliaLang/SparseArrays.jl/blob/c6b613c288b88d44acda7000dc56901228e38e2e/src/sparsevector.jl#L1131

ViralBShah:
Yeah - would be nice to separate these out, and move dense only methods to base. Easier said than done to detangle.

ViralBShah:
Is it possible these tests are failing because SparseArrays is also currently included in the system image?

ViralBShah:
fyi @wimmerer

Wimmerer:
Let me try to figure out the UMFPACK issue today, and then I'll take a look at this. This is just a repo change for now right?

ViralBShah:
Yes, that's a good plan. I just thought it would be good to let you know. For reasons that @KristofferC mentions above, separating out this repo is going to be quite a bit work - but we need to do what we can without breaking in order to pave the way for doing more interesting things going forward.

ViralBShah:
@KristofferC Any ideas why the windows CI is not running? It seems like the same command in SuiteSparse.jl works fine.

ViralBShah:
Ok so the ambiguity tests can be addressed separately as a next step, and the linalg tests should anyways move to SuiteSparse.jl. The rest of the testsuite passes, and once we can fix the windows issue, we can get this merged.

.ci/test_and_change_uuid.jl
```diff
@@ -0,0 +1,28 @@
...
+    using TOML
+    using Test
+else
+    using Pkg: TOML
+    using Test
+end
+
+# To generate the new UUID, we simply modify the first character of the original UUID
+const original_uuid = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
+const new_uuid      = "1f01184e-e22b-5df5-ae63-d93ebab69eaf"
```
DilumAluthge:
```suggestion
const original_uuid = "3f01184e-e22b-5df5-ae63-d93ebab69eaf"
const new_uuid      = "4f01184e-e22b-5df5-ae63-d93ebab69eaf"
```

ViralBShah: Trying this.

Project.toml
```diff
@@ -1,5 +1,5 @@
 name = "SparseArrays"
-uuid = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
+uuid = "3f01184e-e22b-5df5-ae63-d93ebab69eaf"
```
KristofferC, line 2: Eeeh? This need to be changed back.

ViralBShah: That wasn't working.

ViralBShah: I don't understand how this works. Whatever was originally there didn't work and the new thing did. Would be nice if you can fix it to what it needs to be.

fredrikekre: It is wrong here: https://github.com/JuliaLang/SparseArrays.jl/blob/af053194a98848c31ae602cd5ce7f40265b84a8c/.ci/test_and_change_uuid.jl#L10

PR #70, opened 2022-01-17
Add the tests from base

ViralBShah:
Move various sparse tests from julia base tests into basetests.jl here.

This leaves mentions of SparseArrays in hashing.jl, ambiguous.jl (sparse tests are excluded there), and precompile.jl. I believe these need to stay in base.

```
➜  test git:(vs/sparsetests) ✗ git grep SparseArrays | cat
ambiguous.jl:using LinearAlgebra, SparseArrays, SuiteSparse
ambiguous.jl:    STDLIBS = filter!(x -> x != "LinearAlgebra" && x != "SparseArrays" && # Some packages run this test themselves
hashing.jl:using Random, LinearAlgebra, SparseArrays
precompile.jl:                 :SparseArrays, :Statistics, :SuiteSparse, :TOML, :Tar, :Test, :UUIDs, :Unicode,
```

### DIFF
--- a/test/basetests.jl
+++ b/test/basetests.jl
@@ -1,4 +1,10 @@
+# As part of the migration of SparseArrays.jl into its own repo,
+# these tests have been moved from other files in julia tests to
+# the SparseArrays.jl repo
+
 using Random, LinearAlgebra, SparseArrays
+
+# From arrayops.jl
 
 @testset "copy!" begin
     @testset "AbstractVector" begin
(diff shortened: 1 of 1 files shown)


KristofferC (approved):

ViralBShah:
Also, do we really need `using SparseArrays, LinearAlgebra` in ambiguous.jl?

ViralBShah:
We should bump stdlibs for this package once we merge, since these tests will be removed from julia base.

PR #77, opened 2022-02-14
Change type signature of _checkbuffers and _goodbuffers

### LINKED ISSUE #9: defining a new AbstractSparseMatrixCSC requires defining `_checkbuffers` (internal function)
i was in need to define my own type <: `AbstractSparseMatrixCSC` (basically i needed the indexing behaviour of a `SparseMatrixCSC`, but with another storage type). The (not written in the docs) interface for an  `AbstractSparseMatrixCSC` requires:
- `getcolptr`
- `rowvals`
- `nonzeros`
- `size`
the problem is that an internal function `_checkbuffers` is required, mainly for the display machinery. this function does not require new methods, in fact, is defined in terms of the functions mentioned above (`getrowval` gets dispatched to `rowvals`):

https://github.com/JuliaLang/julia/blob/3b5f98510368d757f468d5ea0674229a27a7c526/stdlib/SparseArrays/src/sparsematrix.jl#L82-L90

`_checkbuffers` (and `_goodbuffers`) should be defined in terms of `AbstractSparseMatrixCSC` instead of the concrete type.

abraemer:
This PR only changes 2 lines - the signatures of `_checkbuffers` and `_goodbuffers` as mentioned in the linked issue.

These methods check the assumptions on colptr, rowval and nonzeros and thus should work for all AbstractSparseMatrixCSC. Changing the signature also removes the need for every subtype to implement it, which will fix the packages broken by [JuliaLang/julia/pull/40523](https://github.com/JuliaLang/julia/pull/40523) such as [ThreadedSparseArrays.jl/issues/9](https://github.com/jagot/ThreadedSparseArrays.jl/issues/9)

Fixes #9

### DIFF
--- a/src/sparsematrix.jl
+++ b/src/sparsematrix.jl
@@ -79,8 +79,8 @@ end
 
 size(S::SparseMatrixCSC) = (getfield(S, :m), getfield(S, :n))
 
-_goodbuffers(S::SparseMatrixCSC) = _goodbuffers(size(S)..., getcolptr(S), getrowval(S), nonzeros(S))
-_checkbuffers(S::SparseMatrixCSC) = (@assert _goodbuffers(S); S)
+_goodbuffers(S::AbstractSparseMatrixCSC) = _goodbuffers(size(S)..., getcolptr(S), getrowval(S), nonzeros(S))
+_checkbuffers(S::AbstractSparseMatrixCSC) = (@assert _goodbuffers(S); S)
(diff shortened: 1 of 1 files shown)


abraemer:
After the split, how is compatibility with Julia handled? Do we want/need to merge this somewhere else as well?

Are there basically multiple versions of SparseArrays.jl now? This external one and for each (older) LTS Julia version the one still in the main repository?

PR #80, opened 2022-02-17
Fix syntax in basetests.jl

c42f:
Here's a tiny fix for some broken syntax I noticed in test/basetests.jl

By the way, do people realize that the tests in `test/basetests.jl` are not currently being included?

### DIFF
--- a/test/basetests.jl
+++ b/test/basetests.jl
@@ -2,6 +2,7 @@
 # these tests have been moved from other files in julia tests to
 # the SparseArrays.jl repo
 
+using Test
 using Random, LinearAlgebra, SparseArrays
 
 # From arrayops.jl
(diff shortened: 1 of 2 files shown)


DilumAluthge:
> By the way, do people realize that the tests in `test/basetests.jl` are not currently being included?

That seems... bad. Can you open an issue so we don't forget?

DilumAluthge:
Or make a PR to add it to https://github.com/c42f/SparseArrays.jl/blob/cjf/fix-basetests-syntax/test/testgroups

c42f:
I don't want to dig too deep here, but I added it to `testgroups` in this PR. Let's see what CI does.

DilumAluthge:
Stick a `using Test` at the top of `test/runtests.jl`?

c42f:
Looks like it. I thought that'd be handled by the runner but the other tests are in their own modules for whatever reason.

DilumAluthge (approved):

PR #81, opened 2022-02-18
Add version number

ViralBShah:
Since this is now an external package, it was suggested we have a version number.

cc @staticfloat @KristofferC


ViralBShah:
Let's worry about this later.

PR #85, opened 2022-03-24
remove invalid test for failing sprand(BigFloat)

stevengj:
It doesn't make sense to test that `sprand(BigFloat, ...)` throws a `MethodError`, as this function will work as soon as https://github.com/JuliaLang/julia/issues/17629 is fixed.

In particular, this test is blocking https://github.com/JuliaLang/julia/pull/44714


DilumAluthge (approved):

DilumAluthge:
@ViralBShah After this PR is merged, can we bump Julia master to use the latest commit of SuiteSparse#master? Or should we use a different branch of SuiteSparse, e.g. a release branch (to which we would backport this PR)?

ViralBShah:
For now we are still just using `master`, since there haven't been major changes we are implementing just yet.

PR #92, opened 2022-04-20
use more type stable `cat_t` implementation

aviatesk:
After JuliaLang/julia#45028, it will be more recommended to use this
more type stable version of `cat_t`.


dkarrasch:
@aviatesk Do you think you can cook up a little test for that (now that the julia PR is merged)? Inferability of `cat` is a rather fragile thing.

aviatesk:
Done, see the added test cases.

dkarrasch (approved):

PR #93, opened 2022-04-22
fix findmin/findmax

Moelf:
fixes https://github.com/JuliaLang/julia/issues/44978

### DIFF
--- a/src/sparsevector.jl
+++ b/src/sparsevector.jl
@@ -1433,11 +1433,12 @@ for (fun, comp, word) in ((:findmin, :(<), "minimum"), (:findmax, :(>), "maximum
         # we need to find the first zero, which could be stored or implicit
         # we try to avoid findfirst(iszero, x)
         sindex = findfirst(iszero, nzvals) # first stored zero, if any
-        zindex = findfirst(i -> i <= nzinds[i], eachindex(nzinds)) # first non-stored zero
-        if !isnothing(zindex) && zindex < lastindex(x)
-            zindex += 1
+        zindex = findfirst(i -> i < nzinds[i], eachindex(nzinds)) # first non-stored zero
+        index = if isnothing(zindex) && last(nzinds) < lastindex(x)
+            last(nzinds) + 1
+        else
+            isnothing(sindex) ? zindex : min(sindex, zindex)
         end
-        index = isnothing(sindex) ? zindex : min(sindex, zindex)
         return zeroval, index
     end
 end
--- a/src/sparsevector.jl
+++ b/src/sparsevector.jl
@@ -1434,6 +1434,9 @@ for (fun, comp, word) in ((:findmin, :(<), "minimum"), (:findmax, :(>), "maximum
         # we try to avoid findfirst(iszero, x)
         sindex = findfirst(iszero, nzvals) # first stored zero, if any
         zindex = findfirst(i -> i <= nzinds[i], eachindex(nzinds)) # first non-stored zero
+        if !isnothing(zindex) && zindex < lastindex(x)
+            zindex += 1
+        end
         index = isnothing(sindex) ? zindex : min(sindex, zindex)
         return zeroval, index
     end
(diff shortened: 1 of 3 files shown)


Moelf:
Error seems unrelated

src/sparsevector.jl
```diff
@@ -1433,7 +1433,7 @@ for (fun, comp, word) in ((:findmin, :(<), "minimum"), (:findmax, :(>), "maximum
         # we need to find the first zero, which could be stored or implicit
         # we try to avoid findfirst(iszero, x)
         sindex = findfirst(iszero, nzvals) # first stored zero, if any
-        zindex = findfirst(i -> i < nzinds[i], eachindex(nzinds)) # first non-stored zero
+        zindex = findfirst(i -> iszero(x[i]), eachindex(x)) # first non-stored zero
```
KristofferC, line 1436: The comment above says ` # we try to avoid findfirst(iszero, x)`. But isn't this what is done now?

Moelf: I read that comment and I don't know how to avoid that if the objective is to find the first non-stored zero.

I think the original approach is to find the first "hole" between stored value, which would be the first non-stored zero.

I will give it some thought

dkarrasch: The problem is that indexing into a sparse vector is slow: each time, you need to figure out if the value you're asking for is stored or not. Would be good to check that the fix doesn't spoil the purpose of the specialization, see https://github.com/JuliaLang/julia/issues/42823. If you don't get it "right", it will take a fair amount of time to compute.

Moelf:
404 error is unrelated

src/sparsevector.jl
```diff
@@ -1434,7 +1434,11 @@ for (fun, comp, word) in ((:findmin, :(<), "minimum"), (:findmax, :(>), "maximum
         # we try to avoid findfirst(iszero, x)
         sindex = findfirst(iszero, nzvals) # first stored zero, if any
         zindex = findfirst(i -> i < nzinds[i], eachindex(nzinds)) # first non-stored zero
-        index = isnothing(sindex) ? zindex : min(sindex, zindex)
+        index = if isnothing(zindex) && last(nzinds) < lastindex(x)
+            last(nzinds) + 1
+        else
+            isnothing(sindex) ? zindex : min(sindex, zindex)
+        end
```
stevengj, line 1441:
```suggestion
        zindex = isnothing(zindex) && last(nzinds) < lastindex(x) ? last(nzinds) + 1 : zindex
        return isnothing(sindex) ? zindex : isnothing(zindex) ? sindex : min(sindex, zindex)
```

Moelf:
looks like that logic is flawed

src/sparsevector.jl
```diff
@@ -1434,7 +1434,14 @@ for (fun, comp, word) in ((:findmin, :(<), "minimum"), (:findmax, :(>), "maximum
         # we try to avoid findfirst(iszero, x)
         sindex = findfirst(iszero, nzvals) # first stored zero, if any
         zindex = findfirst(i -> i < nzinds[i], eachindex(nzinds)) # first non-stored zero
-        index = isnothing(sindex) ? zindex : min(sindex, zindex)
+        index = if isnothing(zindex) && isnothing(sindex) && last(nzinds) < lastindex(x)
+            # non-stored zero are contiguous and at the end
+            last(nzinds) + 1
+        elseif isnothing(sindex)
+            zindex
```
stevengj, line 1441:
```suggestion
        index = if isnothing(sindex)
            # non-stored zero are contiguous and at the end
            isnothing(zindex) && last(nzinds) < lastindex(x) ? last(nzinds) + 1 : zindex
```
(Avoid branching twice on `isnothing(sindex)`.)

dkarrasch (approved):

KristofferC:
@dkarrasch, I didn't realize you did not have merge access here. I think you should have gotten an invite now.

dkarrasch:
@KristofferC I think I already had merge access, I was just a bit hesitant to push the button. Tests pass, which is good, but I haven't completely understood the originally missing case here.

Moelf:
The missing case is when the sparse vector is filled with value from the beginning but not all the way to the end.

The original "find first non-stored zero" rely on there existing a gap in the stored values

dkarrasch:
Thanks for the fix! Now, how do we make sure this enters julia v1.8?

Moelf:
something like this: https://github.com/JuliaLang/julia/pull/44142

dkarrasch:
Alright, so the bot will jump in at some point.

Moelf:
idk if the backport to 1.8 will have to be manual though

oscardssmith:
did we ever back port this?

PR #94, opened 2022-04-23
Add dispatch for single-arg findmin/findmax

andrewjradcliffe:
Arose during PR to JuliaLang:master (Updates to findmin/findmax:
function application over multidims #45061)
[https://github.com/JuliaLang/julia/pull/45061](url)

### DIFF
--- a/src/sparsevector.jl
+++ b/src/sparsevector.jl
@@ -1437,6 +1437,7 @@ for (fun, comp, word) in ((:findmin, :(<), "minimum"), (:findmax, :(>), "maximum
         index = isnothing(sindex) ? zindex : min(sindex, zindex)
         return zeroval, index
     end
+    @eval $fun(x::AbstractSparseVector{T}) where {T} = $fun(identity, x)
 end
 
 norm(x::SparseVectorUnion, p::Real=2) = norm(nonzeros(x), p)


stevengj:
LGTM, but I'm not sure why this fallback isn't in Base?  cc @Moelf

Moelf:
Performance issue, someone pointed out to me getindex is super slow so this is one of those cases it's deliberately not implemented?

andrewjradcliffe:
Current Base interface + dispatches for findmin/findmax:
```
findmax(A::AbstractArray; dims=:) = _findmax(A, dims)
findmax(f, domain) = mapfoldl( ((k, v),) -> (f(v), k), _rf_findmax, pairs(domain) )
findmax(itr) = _findmax(itr, :)
_findmax(a, ::Colon) = findmax(identity, a)
_findmax(A, region) # region being Union{Integer, NTuple{M, Integer}} where {M}
```
In the above scheme, the call to `findmax(A)` (with A::AbstractSparseVector) would have first been through `findmax(A::AbstractArray; dims=:)`, which would redirect to `_findmin(a, ::Colon)`, thereby calling `findmax(identity, a)`, which then dispatches to `findmax(f, A::AbstractSparseVector{T}) where {T}`
While this works, it precludes the use of a function argument for multidimensional findmin/findmax. Hence, I was proposing the interface + dispatches described below:

Option 1 of new Base interface + dispatches for findmin/findmax:
```
findmax(A::AbstractArray; dims=:) = _findmax(identity, A, dims)
findmax(f, A::AbstractArray; dims=:) = _findmax(f, A, dims)
findmax(f, domain) = _findmax(f, domain, :)
findmax(itr) = findmax(identity, itr)
_findmax(f, domain, ::Colon) = mapfoldl( ((k, v),) -> (f(v), k), _rf_findmax, pairs(domain) )
_findmax(f, A, region) # region being Union{Integer, NTuple{M, Integer}} where {M}
```

Under the proposed changes, the call to `findmax(A)` (with A::AbstractSparseVector) never has the chance to dispatch on `findmax(f, A::AbstractSparseVector`. This PR adds that dispatch, but I must admit, the whole situation is awkward. If we consider a hypothetical new user, adding a dispatch which provides an optimized method for `findmax` for some subtype of AbstractArray, they would probably think that adding `findmax(f, A::SomeSpecialArray)` would be sufficient. My change forces them to also add `findmax(A::SomeSpecialArray) = findmax(identity, A)`.

There are two design options:
1. the above
2. keep the original interface + dispatch intact, while incorporating the new ideas. 

Option 2 of new Base interface + dispatches for findmin/findmax:
```
findmax(A::AbstractArray; dims=:) = _findmax(A, dims)
findmax(f, A::AbstractArray; dims=:) = _findmax(f, A, dims)
findmax(f, domain) = _findmax(f, domain, :)
findmax(itr) = _findmax(itr, :) # or, equivalently, findmax(identity, itr)
_findmax(a, ::Colon) = findmax(identity, a)
_findmax(A, dims) = _findmax(identity, A, dims)
_findmax(f, A, ::Colon) = mapfoldl( ((k, v),) -> (f(v), k), _rf_findmax, pairs(domain) )
_findmax(f, A, region) # region being Union{Integer, NTuple{M, Integer}} where {M}
```

This approach allows our hypothetical user to get the behavior they expect by defining just `findmax(f, A::SomeSpecialArray)`.

Thoughts:
- Option 1 will be more disturbing to users which are already relying on the extant behavior to define new dispatches.
- Deducing that Option 1 necessitates `findmax(A::SomeSpecialArray) = findmax(identity, A)` is probably non-trivial for a newer user. For experienced users, it begs the question: why is this not handled through a generic?

Having worked through this with a clear head, I don't see much reason to consider anything other than Option 2. Consequently, I believe that this PR is actually unneeded.

andrewjradcliffe:
As the above discussion justifies, and as a full test confirms, this PR is unnecessary. Thus, I close it. Thank you for your time/input Steven and Jerry.

PR #98, opened 2022-05-31
Umfpack Thread safety fix

SobhanMP:
Is this what i was supposed to do? with some luck i managed to "transfer" the commits...

### DIFF
--- a/src/solvers/umfpack.jl
+++ b/src/solvers/umfpack.jl
@@ -138,17 +138,34 @@ function show_umf_info(level::Real = 2.0)
     umf_ctrl[1] = old_prt
 end
 """
-Working space for Umfpack so ldiv! doesn't reallocate
+Working space for Umfpack so `ldiv!` doesn't allocate.
+
+This is allocated once per lu factorization so that Umpfack does not allocated.
+However, this makes `ldiv!` edit the the workspace field of the lu factorization. 
+This structure is thread safe as it's not obvious that `ldiv!` modifies the lu factorization.
+If the lock cannot be aquired, it will send a warning. 
+
+To use multiple threads, each thread should have their own workspace that can be allocated using `Base.similar(::UmfpackWS)` and passed as a kwarg to `ldiv!`.
 """
 struct UmfpackWS{T<:UMFITypes}
     Wi::Vector{T}
     W::Vector{Float64}
+    l::ReentrantLock
 end
-
-Base.similar(w::UmfpackWS) = UmfpackWS(similar(w.Wi), similar(w.W))
+UmfpackWS(Wi, W) = UmfpackWS(Wi, W, ReentrantLock())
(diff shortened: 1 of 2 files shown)


SobhanMP:
@ViralBShah @Wimmerer

SobhanMP:
hmmm

SobhanMP:
btw, i still think that adding a lock would be a very good idea. it's not costly (at least not to anything we do with suitesparse).

Wimmerer:
Okay you can go ahead and add

SobhanMP:
i am somehow not running the tests locally correctly :/

ViralBShah:
Seems like just not finding `@thread`: https://github.com/JuliaSparse/SparseArrays.jl/runs/6678759287?check_suite_focus=true#step:7:98

SobhanMP:
no, i understand that. but why is `Pkg.test()`
 not giving me any errors?

ViralBShah:
`Pkg.test("SparseArrays")` does not give an error on that?

SobhanMP:
yeah, it's not running the tests locally for me. it skips umfpack

SobhanMP:
ah, i forgot to change the hash...

Wimmerer:
That always gets me, I just keep the hash changed and uncommitted.

SobhanMP:
i guess that's it

Wimmerer:
@SobhanMP I forgot the main problem with the lock. Right after this is merged I will move both the Control and Info structs into the `UmfpackLU` struct. `Info` is not thread safe, and I'm not sure it makes a ton of sense to have a lock on just the workspace anymore at that point since `Info` will also make it unsafe (and you don't want the `Info` structs to get clobbered).

We can extend what you had at one point, a function which created an aliased `UmfpackLU` but which allocated a new workspace to also allocate a new Info and Control.

Sorry about that.

SobhanMP:
@Wimmerer ah, damn 

yeah give me an hour

SobhanMP:
@Wimmerer welp that wasn't an hour, but i think this is what you meant, right?

Wimmerer:
Exactly! Let me read it over locally and test and I'll go ahead and merge.

SobhanMP:
btw will this go in 1.8?

Wimmerer:
No. It'll go in 1.9, and actually SparseArrays will no longer be a standard library, just a plain old package at that point.

SobhanMP:
i was looking forward to using this in stuff i'm working on...

ViralBShah:
1.8 is already into RCs - this would have had to be in months ago by feature freeze. Going forward, with 1.9, if we successfully remove SparseArrays from the julia base distribution, then it will just be like any other package.

SobhanMP:
i replaced the stdlib version in my installation and realized i had broken precompile

PR #104, opened 2022-06-05
make lu! allocate less if possible

SobhanMP:
While at this point it should make things a bit nicer, i have two questions:
1- is there any downside to using broadcase, should i write the loop? and,
2- can we reuse the numeric in umfpack?

Right now, it resizes the fields of the umfpacklu instead of reallocating.


SobhanMP:
can we change the
```
tmp = Vector{Ptr{Cvoid}}(undef, 1)
```
to
```
tmp = Ref{Ptr{Void}}()
```
in
https://github.com/JuliaSparse/SparseArrays.jl/blob/23bc7c94ca3169e4e33743fb3ccaf4eb9ec43b4c/src/solvers/umfpack.jl#L497
?

Also lu! isn't safe in the sense that calling an lu! on a matrix with a different sparsity pattern will result in an error since we reuse the symbolic
(see  https://github.com/JuliaSparse/SparseArrays.jl/blob/23bc7c94ca3169e4e33743fb3ccaf4eb9ec43b4c/src/solvers/umfpack.jl#L510)
but we also support changing the size of the matrix.

One solution would be to provide an argument to recalculate the symbolic. Another would be to just not let any one change the size of the the lu.

@Wimmerer what do you think?

Wimmerer:
We should probably check that the input is the same or mark it as an "expert" method. I'm fine with changing the signature of `tmp` provided it still works as before.

Wimmerer:
Also what do you mean reuse the numeric?

SobhanMP:
i said reuse the symbolic not numeric. Every time umfpack calculates the lu it first calculates the symbolic and then the numeric. The symbolic can be reused if the sparsity pattern is the same (not sure what the speed up is).

Wimmerer:
> While at this point it should make things a bit nicer, i have two questions:
> 1- is there any downside to using broadcase, should i write the loop? and,
> 2- can we reuse the numeric in umfpack?

Is what you wrote above, hence my confusion. 

Reusing is fine for symbolic, and I'm surprised it wasn't done already. I do it in KLU.jl.

SobhanMP:
Ah, you were referring to the comment above, yeah i don't think we can reuse the numeric. To be clear sparse arrays does reuse it. should i leave the error handling to umfpack? or should we also do the checks on the julia side?

what's a condition number estimates?

Wimmerer:
Leave what error handling you can to UMFPACK, we should be wrapping exceptions there correctly.

`rcond` is the function, but I'll look into that separately. I'll be home later to look at this PR

ViralBShah:
> can we change the
> 
> ```
> tmp = Vector{Ptr{Cvoid}}(undef, 1)
> ```
> 
> to
> 
> ```
> tmp = Ref{Ptr{Void}}()
> ```

Yes, should be ok to do so. Do you want to add that in this PR?

SobhanMP:
yes, i'm a bit swamped at the moment, hopefully will have some time to do these changes in a few days.

ViralBShah:
No urgency at all. I just have been going through all the issues and PRs today.

ViralBShah:
As a general rule of thumb, I believe symbolic is about 10-15% of the total computation time.

ViralBShah:
> i said reuse the symbolic not numeric. Every time umfpack calculates the lu it first calculates the symbolic and then the numeric. The symbolic can be reused if the sparsity pattern is the same (not sure what the speed up is).

As a general rule of thumb, I believe symbolic is about 10-15% of the total computation time.

SobhanMP:
good to know

SobhanMP:
@Wimmerer @ViralBShah i think this is good to go

Wimmerer:
We shouldn't need locks anymore now that Control, Info, and workspaces are local to each Factorization right?

SobhanMP:
i guess, should i remove them? i think that they don't really cost anything and save the headache from unsuspecting people as it's really not intuitive that ldiv! changes the factorization.

SobhanMP:
i guess, should i remove them? i think that they don't really cost anything and save the headache from unsuspecting people as it's really not intuitive that ldiv! changes the factorization.

SobhanMP:
also should i add a small footnote about parallel solving to the documentations? like it scales well and stuff i'd use something like
```julia
@floop for i in axes(b, 2)
@init w = duplicate(F)
ldiv!(view(x, :, i), w, view(b, :, i))
end
```

Wimmerer:
Well I wouldn't put Floops in the official docs yet. But Threads should be good yes.

SobhanMP:
@Wimmerer i'm still not sure, do i keep the locks? i'll put the parallel stuff in another commit.

Wimmerer:
Someone more experienced with the pros and cons of locks should probably chime in here :/

ViralBShah:
@Wimmerer Is this good to merge?

Wimmerer:
Do the locks make sense to you? It prevents mistakes I guess, even if the intended usage is a thread local view of each Factor.

ViralBShah:
Yes, the locks are fairly coarse grained and look ok. @SobhanMP Can we have some tests that do multi-threaded LUs?

Wimmerer:
I believe he mentioned including those in a separate PR. I may be thinking of docs though. We can merge this part whenever ready.

ViralBShah:
Yes, I see multi-threaded LU tests in the testsuite - but wasn't sure if more were needed as part of this PR.

ViralBShah:
Actually, let's hold on this one and get an stdlib bump first to get the other recent PRs into Julia nightly  - and then merge this one.

ViralBShah:
What's the reason to close? (out of curiosity).

SobhanMP:
gonna open a new one, if you don't mind

PR #107, opened 2022-06-07
update documentations

### LINKED ISSUE #105: halfperm! is not documented.
is this intentional? it has full documentation in the code.

SobhanMP:
closes #105

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -228,6 +228,12 @@ SparseArrays.dropzeros!
 SparseArrays.dropzeros
 SparseArrays.permute
 permute!{Tv, Ti, Tp <: Integer, Tq <: Integer}(::SparseMatrixCSC{Tv,Ti}, ::SparseMatrixCSC{Tv,Ti}, ::AbstractArray{Tp,1}, ::AbstractArray{Tq,1})
+SparseArrays.transpose!
+SparseArrays.adjoint!
+SparseArrays.halfperm!
+SparseArrays.ftranspose!
+SparseArrays.unchecked_noalias_permute!
+SparseArrays.fkeep!
 ```
 
 ```@meta


Wimmerer:
`SparseArrays.transpose!`, `SparseArrays.adjoint!` and `SparseArrays.ftranspose!` don't have docstrings. `franspose!` should probably have a new string added, but the other two shouldn't be documented here I don't believe, there's no additional information not found in the LinearAlgebra docs.

Finally, I'm not sure if we should put `unchecked_noalias_permute!` in the web docs. It's sort of an internal method, but perhaps I'm wrong.

SobhanMP:
yes but it's not intuitive that they exist. it would be great if there is list of functions in the docs

Wimmerer:
Would you say that the `SparseArrays.transpose!` and `SparseArrays.adjoint!` methods are different from the `LinearAlgebra` functions they extend? The docs for `transpose!`:

```julia
transpose!(dest,src)

  Transpose array src and store the result in the preallocated array dest, which should
  have a size corresponding to (size(src,2),size(src,1)). No in-place transposition is
  supported and unexpected results will happen if src and dest have overlapping memory
  regions.
```

Seem to describe what happens pretty well (unless perhaps we handle aliasing in SparseArrays).

For stuff like `unchecked_noalias_permute!` I think maybe you're right. An `Internals` or `Private API` section might be quite nice to have.

SobhanMP:
i still stand by my opinion that i was surprised to see those functions. i didn't think it was possible to have an inplace transpose

Wimmerer:
Okay! Then all we need is a docstring. One that explicitly mentions the in-place transpose.

Wimmerer (approved):
Lgtm

Wimmerer:
One optional thing *I personally* find cleaner is to put the constructor examples in a code block.

SobhanMP:
like this?

ViralBShah:
Let's merge when all green. Docs can always be updated in newer PRs.

fredrikekre:
`halfperm!`, `ftranspose!`, and `fkeep!` feels slightly internal, should they be exported and documented like in this PR?

ViralBShah:
Good point. Maybe just not export?

Wimmerer:
I will remove the exports, I missed those.

Perhaps we should add a section to the SA docs for internal functions once it is excised from the Manual?

Wimmerer:
I'm going to go to bat for `fkeep!` here. I can understand `ftranspose!` not being exported. But `fkeep!` is quite valuable, and something I'll likely re-export from my own packages if it's not exported here.

SobhanMP:
I added fkeep! to the exports, i agree that it's nice to have functions like these

SobhanMP:
also dropzero doc string references fkeep! so i don't think it's really "internal" at that point

ViralBShah:
Documenting is always good! But this stuff all ships in Julia and the APIs have to be supported for a long time if exported - so even if there is some hesitation, the safe thing is to not export. Let's collect some more comments - unexporting is easy.

PR #141, opened 2022-06-18
Add solver docs to SparseArrays docs

ViralBShah:
@Wimmerer Can you check to see if this is the right way to do this?


Wimmerer:
This passes docs build now @ViralBShah. Once it passes main tests and is merged (the new doctests imported from SS could fail perhaps?) we are ready to bump.

ViralBShah:
Note that this shouldn't affect julia itself - just the standalone docs that are presented for SparseArrays.jl through the README docs link.

PR #151, opened 2022-06-21
Split tests into a few different files

Wimmerer:
I'm not very happy with these splits, but they're all relatively small files, and have some logic to them.

This will be followed by a complete test re-org at some point. For now we're just aiming to reduce the sizes of each test file.

If there's a better split just let me know and I'll change it!


ViralBShah:
In the future, it would be better to make branches in this repo than forks since that would then allow others who have commit access to directly add to the branch.

ViralBShah:
One more thing that is useful is to move the `USE_GPL_LIBS` check to the individual solver files, and perhaps put all of them in the same directory - so that they can be added to `testgroups`. That way even those will run in parallel across workers. If the threading test in `runtests.jl` can also go into some other file, then `runtests.jl` will be a clean wrapper that just loads everything in `testgroups`.

test/sparsematrix_ops.jl
```diff
@@ -0,0 +1,445 @@
+# This file is a part of Julia. License is MIT: https://julialang.org/license
+
+module SparseTests
```
ViralBShah, line 3: The module name is same across test files. Perhaps should be different?

ViralBShah:
Figured it's easier to merge and do another PR for the solvers.

Wimmerer:
Oh that's why everyone has their own branches. Good to know!

PR #154, opened 2022-06-22
move USE_GPL_LIBS check to the solver files

ViralBShah:
include all the solvers tests in testgroups
move basetests.jl into issues.jl


Wimmerer (approved):
LGTM, I had some troubles with missing imports here and there, but if no tests fail on that this is good.

Issue #150, opened 2022-06-21
transpose! allocates?


SobhanMP:
```julia
julia> using SparseArrays, LinearAlgebra, BenchmarkTools
       let A = sprandn(100, 100, 0.1),
           X = deepcopy(A)
       
           @btime transpose!($X, $A)
           @time  transpose!(X, A)
           
           # @descend_code_warntype transpose!(X, A)
       end;


  1.229 μs (3 allocations: 128 bytes)
  0.000006 seconds (10 allocations: 368 bytes)
```
i was guessing type safety problems but couldn't find any other than size(A, 2) being any.
![Screenshot from 2022-06-21 12-14-01](https://user-images.githubusercontent.com/2301159/174849298-0f7d2c14-536d-4d87-9b14-cf8def30401c.png)

using @time i found that _distributevals_halfperm! is the culprit, not sure why.

Wimmerer:
Why is `size` Any? I would think that's type stable Int64.

SobhanMP:
i have no idea. Maybe it's a but warn_type. One solution would be to specialize everything in that chain on their functions. It makes sense since the default f is identity

SobhanMP:
i.e. ```function halfperm!(X::AbstractSparseMatrixCSC{Tv,Ti}, A::AbstractSparseMatrixCSC{TvA,Ti},
        q::AbstractVector{<:Integer}, f::F = identity) where {Tv,TvA,Ti,F}```

PR #157, opened 2022-06-23
fix allocations in transpose!

SobhanMP:


SobhanMP:
fixes https://github.com/JuliaSparse/SparseArrays.jl/issues/150

src/sparsematrix.jl
```diff
@@ -1143,7 +1143,7 @@ No additonal memory is allocated other than resizing the rowval and nzval of `X`
 
 See `halfperm!`
 """
-function ftranspose!(X::AbstractSparseMatrixCSC{Tv,Ti}, A::AbstractSparseMatrixCSC{Tv,Ti}, f::Function) where {Tv,Ti}
+function ftranspose!(X::AbstractSparseMatrixCSC{Tv,Ti}, A::AbstractSparseMatrixCSC{Tv,Ti}, f::F) where {Tv,Ti,F}
```
fredrikekre, line 1146:
```suggestion
function ftranspose!(X::AbstractSparseMatrixCSC{Tv,Ti}, A::AbstractSparseMatrixCSC{Tv,Ti}, f::F) where {Tv,Ti,F<:Function}
```

test/sparsematrix_ops.jl
```diff
@@ -442,4 +442,15 @@ end
     @test one(A) isa SparseMatrixCSC{Int}
 end
 
+@testset "transpose! does not allocate" begin
+    function f()
+        A = sprandn(10, 10, 0.1)
+        X = deepcopy(A)
```
fredrikekre, line 448:
```suggestion
        X = copy(A)
```

fredrikekre (commented):
> i don't know why

See https://docs.julialang.org/en/v1/manual/performance-tips/#Be-aware-of-when-Julia-avoids-specializing

test/sparsematrix_ops.jl
fredrikekre: No `copy` copies:
https://github.com/JuliaSparse/SparseArrays.jl/blob/7745517d8e6f22553af0732bb7b018ee6390c893/src/sparsematrix.jl#L422-L423
You should basically never use `deepcopy`.

SobhanMP: It seems weird that it's implemented for AbstractSparseMatrixCSC, but interesting.

`deepcopy` is avoided because of dynamic dispatch?

SobhanMP: > > i don't know why
> 
> See https://docs.julialang.org/en/v1/manual/performance-tips/#Be-aware-of-when-Julia-avoids-specializing

not sure i understand why. The allocations are from dynamic dispatch on the functions?

PR #158, opened 2022-06-23
fix Ti of uniform scaling addition closes #103

### LINKED ISSUE #103: Identity plus sparse matrix changes Ti
```julia
A = spzeros(Float64, Int32, 100, 100)
B = I - A
@assert typeof(A) == typeof(B)
```
fails since B is a `SparseMatrixCSC{Float64, Int64}`

SobhanMP:
Fix #103


SobhanMP:
btw should i implement a specialized method for sparse matrixcscs?


there might be some gains to be made by not converting the uniform scaling to a sparse matrix (that's 3*n allocations) and maybe an inplace method like add!(x, I, M), i have to calculate (at work) `(I-M)` alot and M is 20k*20k so i'd like having one. thoughts?

ViralBShah:
In-place doesn't work unless the sparse nonzero structure is the same. So if it is not the same, will you just error out? If so, that would make sense.

SobhanMP:
no it wont error out, it can resize if it doesn't fit but if the matrix has the right size, it would be allocation free. i'll try getting an implementation next week

SobhanMP:
nvm I + A is very performant already and what i had in mind would just introduce potential bugs and such. though, does an in place addition make sense? i don't see one for dense arrays because i guess it's handled by broadcast.

ViralBShah:
It's best not to touch it unless the gains are worthwhile.

PR #159, opened 2022-06-23
update `inv` error message

SobhanMP:
`inv` should hint that solving the linear system is also an option. it's usually a much faster alternative

### DIFF
--- a/src/linalg.jl
+++ b/src/linalg.jl
@@ -1450,7 +1450,7 @@ kron(A::SparseVectorUnion, B::AdjOrTransSparseVectorUnion) = A .* B
 
 ## det, inv, cond
 
-inv(A::AbstractSparseMatrixCSC) = error("The inverse of a sparse matrix can often be dense and can cause the computer to run out of memory. If you are sure you have enough memory, please either convert your matrix to a dense matrix, e.g. by calling `Matrix` or if `A` can be factorized, use `\` on the dense identity matrix, e.g. `A \ Matrix{eltype(A)}(I, size(A)...)` restrictions of `\` on sparse lhs applies.")
+inv(A::AbstractSparseMatrixCSC) = error("The inverse of a sparse matrix can often be dense and can cause the computer to run out of memory. If you are sure you have enough memory, please either convert your matrix to a dense matrix, e.g. by calling `Matrix` or if `A` can be factorized, use `\` on the dense identity matrix, e.g. `A \\ Matrix{eltype(A)}(I, size(A)...)` restrictions of `\\` on sparse lhs applies.")
 
 # TODO
 
--- a/src/linalg.jl
+++ b/src/linalg.jl
@@ -1450,7 +1450,7 @@ kron(A::SparseVectorUnion, B::AdjOrTransSparseVectorUnion) = A .* B
 
 ## det, inv, cond
 
-inv(A::AbstractSparseMatrixCSC) = error("The inverse of a sparse matrix can often be dense and can cause the computer to run out of memory. If you are sure you have enough memory, please convert your matrix to a dense matrix, e.g. by calling `Matrix`.")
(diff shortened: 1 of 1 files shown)


ViralBShah:
I don't think this is a good recommendation to ask people to do `A \ I` to compute the inverse. Usually you don't want people form the inverse at all, and directly use LU factorization through `\`.

ViralBShah:
I suggest closing this.

Wimmerer:
It may be good to have a greatly increased SparseArrays manual in the future. This info could perhaps go there, with examples of solves.

ViralBShah:
That just gives a dense inverse anyways, but slower. Why is it better than `inv(Matrix(A))`?

SobhanMP:
```julia
using SparseArrays, LinearAlgebra, BenchmarkTools

f(A) = inv(Matrix(A))
A = I + sprandn(10_000, 10_000, 0.0001)

@benchmark f($A)
g(A) = lu(A) \ Matrix{Float64}(I, size(A)...)
@benchmark g($A)
```

f:
BenchmarkTools.Trial: 1 sample with 1 evaluation.
 Single result which took 31.489 s (0.01% GC) to evaluate,
 with a memory estimate of 1.49 GiB, over 8 allocations.

g:
BenchmarkTools.Trial: 2 samples with 1 evaluation.
 Range (min … max):  3.561 s …  3.567 s  ┊ GC (min … max): 5.13% … 7.35%
 Time  (median):     3.564 s             ┊ GC (median):    6.24%
 Time  (mean ± σ):   3.564 s ± 3.984 ms  ┊ GC (mean ± σ):  6.24% ± 1.57%

  █                                                      █  
  █▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁█ ▁
  3.56 s        Histogram: frequency by time        3.57 s <

 Memory estimate: 6.71 GiB, allocs estimate: 20068.

the memory usage should be even better now  with the workspace management. Also the second one can be parallelized. it uses even less memory if the dense I is reused (or ran col by col like)

```
function c(A)
    o = Matrix{eltype(A)}(undef, size(A)...)
    c = Vector{eltype(A)}(undef, size(A, 1))
    c .= 0.0
    l = lu(A)
    for i in axes(A, 2)
        c[i] = 1
        if i > 1
            c[i - 1] = 0
        end
        ldiv!(view(o, :, i), l, c)
    end
    o
end
@benchmark c($A)
```

BenchmarkTools.Trial: 2 samples with 1 evaluation.
 Range (min … max):  2.505 s …    2.712 s  ┊ GC (min … max): 0.33% … 8.05%
 Time  (median):     2.609 s               ┊ GC (median):    4.34%
 Time  (mean ± σ):   2.609 s ± 146.828 ms  ┊ GC (mean ± σ):  4.34% ± 5.46%

  █                                                        █  
  █▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁█ ▁
  2.5 s          Histogram: frequency by time         2.71 s <

 Memory estimate: 5.22 GiB, allocs estimate: 20066.

SobhanMP:
image
with colors :)

ViralBShah:
@DilumAluthge Any idea why the docs CI is breaking. It's giving a 404 immediately in the setup phase.

DilumAluthge:
@saschamann and I recently made changes to `setup-Julia`. Sascha is it possible that we broke something?

SobhanMP:
i had missed a `\`

DilumAluthge:
> @DilumAluthge Any idea why the docs CI is breaking. It's giving a 404 immediately in the setup phase.

https://github.com/julia-actions/setup-julia/pull/115

SobhanMP:
ping.

ViralBShah:
Made a small update for a typo. Please merge when CI passes.

PR #164, opened 2022-06-24
Use conservative OSes

ViralBShah:


DilumAluthge:
Looks like your hypothesis (that 2016 is too recent to trigger the bug) is correct.

If necessary, @staticfloat and I can work with you to figure out how to use Buildkite to set up Windows 7 CI on this repository.

ViralBShah:
I don't think it is worthwhile to worry about Windows 7 CI. It would be good to just fix the Float16 issue.

Issue #116, opened 2020-07-17
`lu` support for custom permutation (like in `cholesky`)


jchen975:
It appears that [`lu`](https://github.com/JuliaLang/julia/blob/6a4793ae77490c7704967acb996902661279396e/stdlib/SuiteSparse/src/umfpack.jl#L154) doesn't support a custom permutation scheme (e.g., AMD) like in [`cholesky`](https://github.com/JuliaLang/julia/blob/6a4793ae77490c7704967acb996902661279396e/stdlib/SuiteSparse/src/cholmod.jl#L1352). 

Is this correct? If so, could the support be added? (Please see the [Discourse discussion](https://discourse.julialang.org/t/worse-performance-with-approximate-minimum-degree-reordering-of-sparse-matrices-lu-than/43212) on this and the resulted performance issue compared to Matlab.)

Thanks!

PR #165, opened 2022-06-25
Add support for qsymbolic to umfpack

SobhanMP:
i had two questions, 
frist do we have style? if so
```julia
function umfpack_numeric!(U::UmfpackLU{ComplexF64,$itype}; reuse_numeric = true, q = nothing)
```
or
```julia
function umfpack_numeric!(U::UmfpackLU{ComplexF64,$itype}; reuse_numeric=true, q=nothing)
```

another question is how i can test this? i don't seem to be able to call AMD.jl when i run julia with `--project`.

i get
```
ERROR: MethodError: no method matching colamd(::SparseMatrixCSC{Float64, Int64}, ::Colamd{Int64})
Closest candidates are:
  colamd(::SparseArrays.SparseMatrixCSC{F, Int64}, ::Colamd{Int64}) where F at ~/.julia/packages/AMD/aB8em/src/COLAMD.jl:81
  colamd(::Symmetric{F, SparseArrays.SparseMatrixCSC{F, Int64}}, ::Colamd{Int64}) where F at ~/.julia/packages/AMD/aB8em/src/COLAMD.jl:106
  colamd(::Hermitian{F, SparseArrays.SparseMatrixCSC{F, Int64}}, ::Colamd{Int64}) where F at ~/.julia/packages/AMD/aB8em/src/COLAMD.jl:108
Stacktrace:
 [1] top-level scope
```
when running 
```julia
using BenchmarkTools, Revise, SparseArrays, LinearAlgebra, AMD
meta = Colamd{Int}()
A = sprand(10, 10, .5)
p_amd = colamd(A, meta)
```
should close https://github.com/JuliaSparse/SparseArrays.jl/issues/116

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -5,9 +5,9 @@ DocTestSetup = :(using SparseArrays, LinearAlgebra)
 ```
 
 Julia has support for sparse vectors and [sparse matrices](https://en.wikipedia.org/wiki/Sparse_matrix)
-in the `SparseArrays` stdlib module. Sparse arrays are arrays that contain enough zeros
-that storing them in a special data structure leads to savings in space and execution time,
-compared to dense arrays.
+in the `SparseArrays` stdlib module. Sparse arrays are arrays that contain enough zeros that storing them in a special data structure leads to savings in space and execution time, compared to dense arrays.
+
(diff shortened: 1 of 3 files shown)


SobhanMP:
The issue with changing the UUID is that i don't think it's possible to call packages that use the original version of SparseArrays.

 Anyhow, hancoded an example. I think it's good to go.

Issue #50, opened 2018-03-26
Common abstractions for nonstructural indicies between Dense and Sparse Matrices


oxinabox:
`indices`  is our abstraction for getting the indices so one can iterate the index of a matrix.
It works great for dense matrices.

However, for sparse matrices, it is almost never what you want to do.
You instread only want to iterate the indicies of the nonstructural elements.
(because there are less of them)
That is done using `rowvals`, and `nzrange`

In a dense matrix one can say that all the elements are nonstructural.
So in a circumstance when you only want to iterate through the nonstructural elements,
you want to iterate through all the elements in a dense matrix.
More generally with a matrix of unknown type, you want to iterate through all the elements.

I thus propose that we should have an abstraction for getting the nonstructural indicies,
which falls back to getting all the indiices,
to make it easier to write code that works efficiently on spare matricies, and also works (as fast as is possible) on other types.

Some thing like

```
const SparseMatrix = AbstractSparseArray{<:Any,<:Any,2}
colinds(A::AbstractMatrix)  = indices(A,2)
colinds(A::SparseMatrix) = rowvals(A)

rowinds(A::AbstractMatrix, col::Integer) = indices(A,1)
rowinds(A::SparseMatrix, col::Integer) = nzrange(A, col)
```
(bikeshed on names pending)

There is also a similar relationship between `nonzeros` (sparse) and `vec` (dense/fallback).

I was discussing this on slack with @mbauman  and @StefanKarpinski the other day,
and wanted to put it on GitHub before it was lost to the ages

Issue #83, opened 2022-02-22
dot with Symmetric view of sparse matrix is very slow


goulart-paul:
Calling three argument `dot` with a Symmetric view of a sparse matrix is very slow.  

Example: 

```julia
using LinearAlgebra, SparseArrays, BenchmarkTools
A = sprandn(1000,1000,0.1)
A = A+A'   #symmetric 
B = Symmetric(triu(A))
x = randn(1000)
```
The quadratic form `x'*B*x` is about 100x slower using `dot`, although fine when computed directly.    The same operation is about the same either way using `A`.
```julia
julia> @btime ($x'*$A*$x);
  173.777 μs (1 allocation: 7.94 KiB)

julia> @btime ($x'*$B*$x);
  97.280 μs (1 allocation: 7.94 KiB)

julia> @btime dot($x,$A,$x);
  174.176 μs (0 allocations: 0 bytes)

julia> @btime dot($x,$B,$x);
  9.559 ms (0 allocations: 0 bytes)
```

The final case above ends up appears to be falling back on the general symmetric `dot` implementation [here](https://github.com/JuliaLang/julia/blob/e4920250cea633e5460384b7529c7554d96b73a5/stdlib/LinearAlgebra/src/symmetric.jl#L578).   

This might be related to #23.

ViralBShah:
Perhaps just need to use `parent` or something.

Issue #23, opened 2021-03-02
dot(::SparseMatrixCSC, ::Array) should be specialized


matbesancon:
```
using LinearAlgebra, SparseArrays
using BenchmarkTools

julia> M = rand(500, 300);
julia> S = spzeros(size(M)...);
julia> @btime dot($M, $S);
  521.241 μs (0 allocations: 0 bytes)
```

In comparison, even materializing `S` as dense is better time-wise at this scale:
```
julia> @btime dot($S, $S);
  236.493 ns (0 allocations: 0 bytes)

julia> @btime dot($M, $M);
  42.695 μs (0 allocations: 0 bytes)

julia> @btime dot(collect($S), $M);
  248.822 μs (2 allocations: 1.14 MiB)
```

The specialized version should iterate only on the nonzeros of the sparse matrix

ViralBShah:
In 1.12.7

```
julia> @btime dot($M, $S);
  173.133 ns (0 allocations: 0 bytes)

julia> @btime dot($M, $M);
  4.226 μs (0 allocations: 0 bytes)

julia> @btime dot(collect($S), $M);
  18.625 μs (3 allocations: 1.16 MiB)
```

PR #167, opened 2022-06-26
add the iteratenz API

SobhanMP:
I couldn't find anything like this so i thought i'd make a small implementation. the idea is that this makes iteration on sparsearrays much easier, no need to deal with colptr & friends. It's still pretty fast. But this begs the question, has this been in the api and did i miss it?

This should make adding special matrice operations easy.

```julia
julia> using BenchmarkTools, SparseArrays, LinearAlgebra

julia> A = I + sprandn(10_000, 10_000, 0.001);

julia> function f(A)
           c = zero(eltype(A))
           for (i, j, v) in SparseArrays.iternz(A)
               if (i + j) % 2 == 0
                   c += v 
               end
           end
           return c
       end
f (generic function with 1 method)

julia> function g(A)
           c = zero(eltype(A))
           for (i, j, v) in zip(findnz(A)...)
               if (i + j) % 2 == 0
                   c += v 
               end
           end
           return c
       end
g (generic function with 1 method)

julia> function h(A)
           c = zero(eltype(A))
           @inbounds for j in axes(A, 2),
               ind in A.colptr[j]: A.colptr[j + 1] - 1
               i = A.rowval[ind]
               if (i + j) % 2 == 0
                   c += A.nzval[ind]
               end
           end
       
           c
       end
h (generic function with 1 method)

julia> function k(A)
           c = zero(eltype(A))
           cptr = A.colptr
           rv = A.rowval
           nzv = A.nzval
           @inbounds for j in axes(A, 2),
               ind in cptr[j]: cptr[j + 1] - 1
               i = rv[ind]
               if (i + j) % 2 == 0
                   c += nzv[ind]
               end
           end
       
           c
       end 
k (generic function with 1 method)

julia> function l(A)
           c = zero(eltype(A))
           @inbounds for j in axes(A, 2),
               ind in SparseArrays.getcolptr(A)[j]:SparseArrays.getcolptr(A)[j + 1] - 1
               i = SparseArrays.getrowval(A)[ind]
               if (i + j) % 2 == 0
                   c += SparseArrays.getnzval(A)[ind]
               end
           end
           c
       end 
l (generic function with 1 method)

julia> @assert f(A) == g(A) == h(A) == k(A) == l(A)

julia> @benchmark f($A)
BenchmarkTools.Trial: 10000 samples with 1 evaluation.
 Range (min … max):  149.655 μs … 272.489 μs  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     156.244 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   162.705 μs ±  18.228 μs  ┊ GC (mean ± σ):  0.00% ± 0.00%

  ▇█▃▁▁▃▄▂▂▂▂▇▃▂▁▁▁▃▁▁▁▁▁                           ▁▄          ▂
  █████████████████████████▇▇▇██▆▅▆▆▆▆▆▅▆▆▆▅▆▅▅▄▅▅▆▆██▇▄▅▆▅▆█▆▆ █
  150 μs        Histogram: log(frequency) by time        227 μs <

 Memory estimate: 0 bytes, allocs estimate: 0.

julia> @benchmark g($A)
BenchmarkTools.Trial: 9411 samples with 1 evaluation.
 Range (min … max):  301.049 μs …   4.224 ms  ┊ GC (min … max): 0.00% … 55.37%
 Time  (median):     393.407 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   527.771 μs ± 346.183 μs  ┊ GC (mean ± σ):  8.61% ± 12.64%

  █▆▅▅▅▄▄▄▄▃▂▃▂▂▂▂▂▁▁▁▁▁▁▁      ▁▁   ▁▂                         ▂
  ██████████████████████████▇██████▇█████████▇▆▇▆▅▆▅▆▆▆▅▅▄▄▄▅▅▄ █
  301 μs        Histogram: log(frequency) by time        1.8 ms <

 Memory estimate: 2.52 MiB, allocs estimate: 6.

julia> @benchmark h($A)
BenchmarkTools.Trial: 10000 samples with 1 evaluation.
 Range (min … max):  370.690 μs … 559.212 μs  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     389.377 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   397.357 μs ±  21.079 μs  ┊ GC (mean ± σ):  0.00% ± 0.00%

       ▃▇█▆▅▄▃▁                                                  
  ▁▁▂▃▇████████▇▅▄▃▃▃▄▄▅▄▄▄▄▃▃▃▃▂▂▂▂▂▂▂▂▂▂▂▂▁▁▁▁▁▁▁▂▂▂▁▁▁▁▁▁▁▁▁ ▃
  371 μs           Histogram: frequency by time          472 μs <

 Memory estimate: 0 bytes, allocs estimate: 0.

julia> @benchmark k($A)
BenchmarkTools.Trial: 10000 samples with 1 evaluation.
 Range (min … max):  380.104 μs … 556.205 μs  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     394.372 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   401.398 μs ±  17.597 μs  ┊ GC (mean ± σ):  0.00% ± 0.00%

        ▃▆█▇▃                                                    
  ▁▁▂▃▄▇█████▇▃▃▂▂▃▃▃▃▃▃▃▂▂▂▂▂▂▂▂▂▂▂▂▂▁▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁ ▂
  380 μs           Histogram: frequency by time          470 μs <

 Memory estimate: 0 bytes, allocs estimate: 0.

julia> @benchmark l($A)
BenchmarkTools.Trial: 10000 samples with 1 evaluation.
 Range (min … max):  377.320 μs … 619.899 μs  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     388.873 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   395.194 μs ±  16.974 μs  ┊ GC (mean ± σ):  0.00% ± 0.00%

       ▄█▇▄▁                                                     
  ▁▁▂▅██████▆▄▃▃▂▂▂▂▃▂▂▂▃▂▂▂▂▂▂▂▂▁▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁ ▂
  377 μs           Histogram: frequency by time          463 μs <

 Memory estimate: 0 bytes, allocs estimate: 0.

julia> @benchmark f($A)
BenchmarkTools.Trial: 10000 samples with 1 evaluation.
 Range (min … max):  145.030 μs … 290.142 μs  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     156.375 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   163.418 μs ±  18.701 μs  ┊ GC (mean ± σ):  0.00% ± 0.00%

     ▇█▆▃▁▄▄▃▃▃▂▇▃▃▂▁▃▃▂▂▁▂▁▁▂▁▁▁   ▁                 ▄▃        ▂
  ▃▃▁█████████████████████████████████████▇█▇█▇▇█▇▆▇▆▇██▇▆▆▇▇██ █
  145 μs        Histogram: log(frequency) by time        225 μs <

 Memory estimate: 0 bytes, allocs estimate: 0.
```
tho my problem that it makes no sense that it's faster than manual iteration :/

### DIFF
--- a/src/abstractsparse.jl
+++ b/src/abstractsparse.jl
@@ -124,4 +124,12 @@ julia> findnz(A)
 """
 function findnz end
 
+"""
+    iternz(A::AbstractSparseArray)
+
+Equivalent to `zip(findnz(A)...)` but does not allocated
+```
+"""
+function iternz end
+
 widelength(x::AbstractSparseArray) = prod(Int64.(size(x)))
--- a/src/sparsematrix.jl
+++ b/src/sparsematrix.jl
@@ -1700,6 +1700,37 @@ function sparse_sortedlinearindices!(I::Vector{Ti}, V::Vector, m::Int, n::Int) w
     return SparseMatrixCSC(m, n, colptr, I, V)
 end
 
+
+abstract type SparseIndexIterate end
+@inline getcolptr(x::SparseIndexIterate) = getcolptr(x.m)
+@inline getrowval(x::SparseIndexIterate) = getrowval(x.m)
+@inline getnzval(x::SparseIndexIterate) = getnzval(x.m)
+@inline nonzeroinds(x::SparseIndexIterate) = nonzeroinds(x.m)
+@inline nonzeros(x::SparseIndexIterate) = nonzeros(x.m)
+@inline nnz(x::SparseIndexIterate) = nnz(x.m)
+
+Base.length(x::SparseIndexIterate) = nnz(x.m)
+Base.size(x::SparseIndexIterate) = size(x.m)
+Base.size(x::SparseIndexIterate, i) = size(x.m)[i]
+struct IterateNZCSC{T<: AbstractSparseMatrixCSC} <: SparseIndexIterate
+    m::T
+end
+
+Base.eltype(::IterateNZCSC{T}) where {Ti, Tv, T <: AbstractSparseMatrixCSC{Tv, Ti}} = Tuple{Ti, Ti, Tv}
+Base.iterate(x::IterateNZCSC, state=(1, 0)) = @inbounds let (j, ind) = state
+    ind += 1
+    while (j < size(x, 2)) && (ind > getcolptr(x)[j + 1] - 1)
+        j += 1
+    end
+    (j > size(x, 2) || ind > getcolptr(x)[end] - 1) && return nothing
+
+    (getrowval(x)[ind], j, getnzval(x)[ind]), (j, ind)
+end
+
+iternz(S::AbstractSparseMatrixCSC) = IterateNZCSC(S)
+
+
+
 """
     sprand([rng],[type],m,[n],p::AbstractFloat,[rfn])
 
--- a/src/sparsevector.jl
+++ b/src/sparsevector.jl
@@ -829,6 +829,26 @@ function _sparse_findprevnz(v::SparseVector, i::Integer)
     end
 end
 
+
+
+struct IterateSparseVec{T<: AbstractSparseVector} <: SparseIndexIterate
+    m::T
+end
+
+Base.eltype(::IterateSparseVec{T}) where {Ti, Tv, T <: AbstractSparseVector{Tv, Ti}} = Tuple{Ti, Tv}
+
+Base.iterate(x::IterateSparseVec, state=0) = @inbounds begin
+    state += 1
+    if state > nnz(x)
+        nothing
+    else
+        (nonzeroinds(x)[state], nonzeros(x)[state]), state
+    end
+end
+
+iternz(S::AbstractSparseVector) = IterateSparseVec(S)
+
+
 ### Generic functions operating on AbstractSparseVector
 
 ### getindex
--- a/test/sparsematrix_constructors_indexing.jl
+++ b/test/sparsematrix_constructors_indexing.jl
@@ -1578,4 +1578,10 @@ end
     @test_throws ArgumentError SparseArrays.expandptr([2; 3])
 end
 
+@testset "iteratenz" begin
+    for i in 1:20
+        A = sprandn(100, 100, 1 / i)
+        @test collect(SparseArrays.iternz(A)) == collect(zip(findnz(A)...))
+    end
+end
 end
--- a/test/sparsevector.jl
+++ b/test/sparsevector.jl
@@ -338,6 +338,13 @@ end
         @test findnz(xc) == ([2, 3, 5], [1.25, 0, -0.75])
     end
 end
+
+@testset "iteratenz (vector)" begin
+    for i in 1:10
+        A = sprandn(100, i / 100)
+        @test collect(SparseArrays.iternz(A)) == collect(zip(findnz(A)...))
+    end
+end
 ### Array manipulation
 
 @testset "copy[!]" begin


Wimmerer:
This is great, I have a prototype version based on ArrayIteration.jl, but having a version with no dependencies is also great.

SobhanMP:
@ViralBShah what do you think about implementing the same for all special matrices? this could give rise to fast generic implementations for many of the functions like dot. if you think it's a good idea, where should i submit a pull request?

Wimmerer:
That should almost certainly be experimented with in a separate package first.

ViralBShah:
Agree. I would put that in a separate package.

SobhanMP:
i made https://github.com/SobhanMP/SparseExtra.jl for now, guess i'll add a bunch of other stuff while i'm at it...

ViralBShah:
I would love to understand why this is faster? If we want to use this iterator across other parts of SparseArrays, and see if the pattern makes sense (it should), then we can go ahead and merge this.

SobhanMP:
note that it's not faster than the last function (the pattern used almost everywhere in Sparsearrays). accessing via the getcolptr is faster than naming a variable (i don't know why). it could be used to simplify a lot of code. for instance
```julia
const S_or_SS{Tv, Ti} = Union{
    AbstractSparseMatrixCSC{Tv, Ti},
    [i{AbstractSparseMatrixCSC{Tv, Ti}} 
        for i in [Symmetric, LowerTriangular, UpperTriangular, Transpose, Adjoint, Hermitian]]...}
function dot(x::AbstractVector, 
    A::S_or_SS{Tv},
    y) where Tv  
    acc = zero(promote_type(eltype(x), eltype(A), eltype(y)))
    for (v, i, j) in iternz(A)
        acc += x[i] * v * y[j]
    end
end
```
would be a correct implementation for 1 indexed x and y and all sort of special matrix. It could also be used for the dense version but the dense twice as slow.

SobhanMP:
@ViralBShah i've updated the readme of my repo, feel free to check it out

SobhanMP:
@ViralBShah should this stay in SparseExtra.jl (and close this) or should i make this a pull request (in contrast to a draft)?

this would make closing https://github.com/JuliaSparse/SparseArrays.jl/issues/23, https://github.com/JuliaSparse/SparseArrays.jl/issues/83, and possibly other `dot` implementation possible. Think transpose of diagonal matrix of sparse vector. It would also simplify every function that has a fast loop over sparse arrays.

would maybe close https://github.com/JuliaSparse/SparseArrays.jl/issues/50 
ref https://sobhanmp.github.io/SparseExtra.jl/iternz/

ViralBShah:
If the design is ready, let's bring it in. I will look at the SparesExtra package soon.

ViralBShah:
@Wimmerer Can you review too?

ViralBShah:
I am persuaded enough by the real nice documentation. Let's open the PR. 1.9 feature freeze may be soon, so let's get it in. Let's also get the design doc in. The real moment of truth will be when we start using it throughout the library. 

@fredrikekre @KristofferC @dkarrasch Any comments on the API in https://sobhanmp.github.io/SparseExtra.jl/iternz/

Wimmerer:
I like it. A lite version of ArrayIteration.jl, should simplify the code base a lot. 

Thoughts on exporting? I lean towards no for now.

SobhanMP:
in term of API it's done (at least the outward facing part) but there are three things worth considering.
- skip_col/skip_row_to skip to the next element - 1 so that calling next on their returned state gives the desired element.
- how about some infrastructure for matrices that are not column major
- does it need to support arrays that don't start at one? (for algebraic ones)

ViralBShah:
We can keep it internal for a whole release cycle and export it later. My thought is to get something in that is small and addresses the current data structure, use it widely in the package, and then expand it to be more general.

SobhanMP:
@fredrikekre @KristofferC @dkarrasch Any comments on the API in https://sobhanmp.github.io/SparseExtra.jl/iternz/, which we are discussing bringing into SparseArrays.jl?

SobhanMP:
@ViralBShah @dkarrasch updates?

ViralBShah:
In the link, the API just shows, which feels nice enough. Where's the `skip_col` stuff?

```
all(iternz(x)) do (v, k...)
    x[k...] == v
end
```

SobhanMP:
> I hadn't seen the rest of the API here

i don't think it makes sense to bring that in SparseArrays. the parallel `\` is nice but i'm not sure there is any merit in bringing it here.

dkarrasch:
So I finally took a look at SparseExtra.jl. It's very nice that it composes with outer wrappers! What seems to be currently missing is that for `Symmetric` and `Hermitian`, the elements of the viewed triangle are `transpose`d and `adjoint`, respectively. That doesn't show up in cases with real eltypes, but it will show up, for instance, in 3-arg `dot` with a (complex) `Hermitian` matrix. Moreover, when you read diagonal elements, you need to take real part in the Hermitian case, and, more generally, call `symmetric` or `hermitian` on the underlying element; see 

https://github.com/JuliaLang/julia/blob/690a5f67c13fd23c7b48e60c31bfa565c0eee861/stdlib/LinearAlgebra/src/symmetric.jl#L236-L255

Unfortunately, I wasn't able to follow all the details to propose how to fix this. I believe that these are the only two wrappers that have non-trivial manipulations, so when we get those right, this could be ported here and measured for performance.

Also, x-ref https://github.com/JuliaLang/julia/pull/40103.

SobhanMP:
@dkarrasch The links were very helpful. I think that this works as intended now.

```julia
julia> A = Symmetric([randn(20, 20) for _ in 1:20, _ in 1:20]);
julia> sum(sum(A[i, j] - v for (v, i, j) in iternz(A))) == 0
true
julia> A = Hermitian([randn(20, 20) + randn(20, 20) * 1im for _ in 1:20, _ in 1:20]);
julia> sum(sum(A[i, j] - v for (v, i, j) in iternz(A))) == 0
true
```

Are there still interesting in this? Should I update the pull request? Do we want more tests?

ViralBShah:
I thought the main issue here was the significant performance gap. I would be supportive of bringing the iterator in - it is a nice API, but perhaps we can't migrate the internals to it until it is on par or faster. However, it can still be made available to users (with an appropriate note).

@rayegun @dkarrasch @jishnub @fredrikekre Thoughts?

SobhanMP:
> I thought the main issue here was the significant performance gap.

I don't think that this is an issue look at the plots in 
https://sobhanmp.github.io/SparseExtra.jl/iternz/ 

The loss of performance is fairly small. IMHO, even if we do not replace existing code, we can at least make working fallbacks for all methods with this.

ViralBShah:
I am onboard with introducing it.

ViralBShah:
@SobhanMP Not sure if you are still paying attention to this repo - but it would be nice to get this in, and to get your current thoughts on the topic as well.

PR #171, opened 2022-06-30
remove solvers/lib form codecov

SobhanMP:
the code coverage has dropped drastically since suitesparse was merged. i don't think it's meaningful to calculate the code coverage on these files.


Wimmerer:
This seems reasonable to me.

PR #177, opened 2022-07-08
Determine correct eltype in sparse map

### LINKED ISSUE #101: Sparse map does not compute the correct element type
```
julia> a = SparseVector(2, [1,2], Union{Int, Missing}[1, missing])
2-element SparseVector{Union{Missing, Int64}, Int64} with 2 stored entries:
  [1]  =  1
  [2]  =  missing

julia> a + a
ERROR: TypeError: non-boolean (Missing) used in boolean context
Stacktrace:
 [1] _binarymap_mode_1!(f::typeof(+), mx::Int64, my::Int64, xnzind::Vector{Int64}, xnzval::Vector{Union{Missing, Int64}}, ynzind::Vector{Int64}, ynzval::Vector{Union{Missing, Int64}}, rind::Vector{Int64}, rval::Vector{Int64})
   @ SparseArrays ~/julia/usr/share/julia/stdlib/v1.9/SparseArrays/src/sparsevector.jl:1305
 [2] _binarymap(f::typeof(+), x::SparseVector{Union{Missing, Int64}, Int64}, y::SparseVector{Union{Missing, Int64}, Int64}, mode::Int64)
   @ SparseArrays ~/julia/usr/share/julia/stdlib/v1.9/SparseArrays/src/sparsevector.jl:1260
 [3] +(x::SparseVector{Union{Missing, Int64}, Int64}, y::SparseVector{Union{Missing, Int64}, Int64})
   @ SparseArrays ~/julia/usr/share/julia/stdlib/v1.9/SparseArrays/src/sparsevector.jl:1384
 [4] top-level scope
   @ REPL[18]:1
```
This needs to use the same logic as `map` for arrays, rather than just assuming that the first element has the correct array type.

dkarrasch:
Closes #101.

### DIFF
--- a/src/SparseArrays.jl
+++ b/src/SparseArrays.jl
@@ -32,6 +32,14 @@ export AbstractSparseArray, AbstractSparseMatrix, AbstractSparseVector,
     sprand, sprandn, spzeros, nnz, permute, findnz,  fkeep!, ftranspose!,
     sparse_hcat, sparse_vcat, sparse_hvcat
 
+# helper function needed in sparsematrix, sparsevector and higherorderfns
+@inline _iszero(x) = x == 0
+@inline _iszero(x::Number) = Base.iszero(x)
+@inline _iszero(x::AbstractArray) = Base.iszero(x)
+@inline _isnotzero(x) = (x != 0) !== false # like `x != 0`, but handles `x::Missing`
+@inline _isnotzero(x::Number) = !iszero(x)
+@inline _isnotzero(x::AbstractArray) = !iszero(x)
+
 include("abstractsparse.jl")
 include("sparsematrix.jl")
 include("sparseconvert.jl")
--- a/src/higherorderfns.jl
+++ b/src/higherorderfns.jl
@@ -9,7 +9,8 @@ import Base: map, map!, broadcast, copy, copyto!
 using Base: front, tail, to_shape
 using ..SparseArrays: SparseVector, SparseMatrixCSC, AbstractSparseVector, AbstractSparseMatrixCSC,
                       AbstractSparseMatrix, AbstractSparseArray, indtype, nnz, nzrange, spzeros,
-                      SparseVectorUnion, AdjOrTransSparseVectorUnion, nonzeroinds, nonzeros, rowvals, getcolptr, widelength
+                      SparseVectorUnion, AdjOrTransSparseVectorUnion, nonzeroinds, nonzeros,
+                      rowvals, getcolptr, widelength, _iszero, _isnotzero
 using Base.Broadcast: BroadcastStyle, Broadcasted, flatten
 using LinearAlgebra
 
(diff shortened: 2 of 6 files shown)


dkarrasch:
How do I test `SparseArrays.jl` locally? Do we have a little explanation somewhere? I'm getting that "`Serialization`" error, but missed how to fix it. If we don't yet have some explanation, could someone please write up such a, say, `CONTRIBUTING.md` file or something for `SparseArrays.jl`?

ViralBShah:
I have been only going by the CI here for testing myself - but @Wimmerer may have a better flow to ensure that the tests pass on Julia master as well.

SobhanMP:
on master calling resolve/instantiate fixes it

src/SparseArrays.jl
```diff
@@ -32,6 +32,14 @@ export AbstractSparseArray, AbstractSparseMatrix, AbstractSparseVector,
     sprand, sprandn, spzeros, nnz, permute, findnz,  fkeep!, ftranspose!,
     sparse_hcat, sparse_vcat, sparse_hvcat
 
+# helper function needed in sparsematrix, sparsevector and higherorderfns
+@inline _iszero(x) = x == 0
+@inline _iszero(x::Number) = Base.iszero(x)
+@inline _iszero(x::AbstractArray) = Base.iszero(x)
+@inline _isnotzero(x) = (x != 0) !== false # like `x != 0`, but handles `x::Missing`
+@inline _isnotzero(x::Number) = !iszero(x)
+@inline _isnotzero(x::AbstractArray) = !iszero(x)
```
ViralBShah, line 41: Should these methods be added to Base, rather than here?

ViralBShah:
Good to get a few more eyes here. @fredrikekre @KristofferC.

dkarrasch:
The result eltype computation is standard and consistent with dense (and even sparse, yet matrix) broadcast. The only potentially "controversial" aspect I see is that `_isnotzero(missing) == true`, where `!iszero(missing) == missing`. This is reasonable, I believe, since we do want to store `missing` entries and want to avoid the "using non-boolean value in a boolean context" error. Also, such behaviour is needed to write `missing`s into a `SparseArray{Union{T, Missing}}`.

I wonder if I should add another test loop such that only one of the two arguments is of union type, and the other has a concrete (number) type. OTOH, code coverage suggests that the corresponding `_getmode` functions (with only one `Missing`) are called already.

ViralBShah:
> How do I test `SparseArrays.jl` locally? Do we have a little explanation somewhere? I'm getting that "`Serialization`" error, but missed how to fix it. If we don't yet have some explanation, could someone please write up such a, say, `CONTRIBUTING.md` file or something for `SparseArrays.jl`?

https://github.com/JuliaSparse/SparseArrays.jl/pull/188 should be a good starting point.

ViralBShah:
@dkarrasch Is this good to merge?

dkarrasch:
It's been good to merge for two weeks. But you wanted to get a few more eyes on it, which is always advisable.

Wimmerer (approved):

PR #183, opened 2022-07-13
umfpack: initialize pointers to null

SobhanMP:
@ViralBShah that's all of them as far as i can see but should i also set the non pointer refs to zero?


SobhanMP:
that should be all of them

ViralBShah:
Running the SparseArrays bump.

ViralBShah:
BTW, @SobhanMP Now that you have write access, you can create branches here rather than working with forks. Hopefully that is a bit easier.

PR #185, opened 2022-07-13
umfpack: overload duplicate for transposed LUs

SobhanMP:
should make everyone's life one line easier

### DIFF
--- a/src/solvers/umfpack.jl
+++ b/src/solvers/umfpack.jl
@@ -209,23 +209,29 @@ workspace_W_size(F::UmfpackLU) = workspace_W_size(F, has_refinement(F))
 workspace_W_size(S::Union{UmfpackLU{<:AbstractFloat}, SparseMatrixCSC{<:AbstractFloat}}, refinement::Bool) = refinement ? 5 * size(S, 2) : size(S, 2)
 workspace_W_size(S::Union{UmfpackLU{<:Complex}, SparseMatrixCSC{<:Complex}}, refinement::Bool) = refinement ? 10 * size(S, 2) : 4 * size(S, 2)
 
+const ATLU{Tv, Ti} = Union{Transpose{Tv, UmfpackLU{Tv, Ti}}, Adjoint{Tv, UmfpackLU{Tv, Ti}}}
+has_refinement(F::ATLU) = has_refinement(F.parent)
 has_refinement(F::UmfpackLU) = has_refinement(F.control)
 has_refinement(control::AbstractVector) = control[JL_UMFPACK_IRSTEP] > 0
 
-# auto magick resize
+# auto magick resize, should this only expand and not shrink?
 getworkspace(F::UmfpackLU) = @lock F.lock begin
         resize!(F.workspace, F, has_refinement(F))
         F.workspace
     end
 
-UmfpackWS(F::UmfpackLU{Tv, Ti}, refinement::Bool=has_refinement(F)) where {Tv,Ti} = UmfpackWS{Ti}(
+UmfpackWS(F::UmfpackLU{Tv, Ti}, refinement::Bool=has_refinement(F)) where {Tv, Ti} = UmfpackWS(
         Vector{Ti}(undef, size(F, 2)),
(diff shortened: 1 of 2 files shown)


SobhanMP:
i finally managed to get a julia build with git SparseArrays and using the new patch in julia. Turns out that i needed both this and SuiteSparse git. it's very good i even update the parallel solver in sparse extra for this (and now CI fails :rofl: ) i think these make writing a parallel solve easy. If any brave soul could answer my two questions and green light this, i'd be grateful.

1- right now the workspace is automatically resized at every solve (it doesn't do anything unless refined is changed). maybe it makes sense to only expand?
2- should i add `duplicate` to the docs?

if this is done, there is only a small patch that i'll write after (to avoid conflicts) and i think the umfpack is ready for a bump.

ViralBShah:
> i finally managed to get a julia build with git SparseArrays and using the new patch in julia. Turns out that i needed both this and SuiteSparse git. it's very good i even update the parallel solver in sparse extra for this (and now CI fails 🤣 ) i think these make writing a parallel solve easy. If any brave soul could answer my two questions and green light this, i'd be grateful.

It would be nice to document this in the README.md of SparseArrays.jl, as @dkarrasch suggested in another issue. Not sure what parallel solve means here. Pointers?

> 1- right now the workspace is automatically resized at every solve (it doesn't do anything unless refined is changed). maybe it makes sense to only expand? 2- should i add `duplicate` to the docs?

It would have been nice to use an existing concept like `similar` or `copy`. I didn't quite follow why we can't use that. Is the workspace resized at every solve because the workspace is the size allocated in the last solve? Apologies for the silly questions, since it's been a while since I dived into the API. 

> if this is done, there is only a small patch that i'll write after (to avoid conflicts) and i think the umfpack is ready for a bump.
> 
> PS. these are just a few helpers to make adding workspace to UmfpackLU and their transposed easier.

I think once we iron out `duplicate`, we should be able to merge shortly.

src/solvers/umfpack.jl
```diff
@@ -209,39 +209,45 @@ workspace_W_size(F::UmfpackLU) = workspace_W_size(F, has_refinement(F))
 workspace_W_size(S::Union{UmfpackLU{<:AbstractFloat}, SparseMatrixCSC{<:AbstractFloat}}, refinement::Bool) = refinement ? 5 * size(S, 2) : size(S, 2)
 workspace_W_size(S::Union{UmfpackLU{<:Complex}, SparseMatrixCSC{<:Complex}}, refinement::Bool) = refinement ? 10 * size(S, 2) : 4 * size(S, 2)
 
+const ATLU{Tv, Ti} = Union{Transpose{Tv, UmfpackLU{Tv, Ti}}, Adjoint{Tv, UmfpackLU{Tv, Ti}}}
```
dkarrasch, line 212: Seems like you don't need the type parameters below, so

```suggestion
const ATLU = Union{Transpose{<:Any, UmfpackLU}, Adjoint{<:Any, UmfpackLU}}
```

might be enough? Not sure about `<:Any` vs. `Any` above.

SobhanMP: seems resonable

SobhanMP:
> parallel solve
https://github.com/SobhanMP/SparseExtra.jl/blob/main/src/solve.jl
(use multiple cpu cores for matrix lhs)
> resize
 the workspace is resized in case the refinement was changed/lu! changed the size of lu factorization (which i think should be made illegal by making the lu struct unmutable). now in most cases it's a noop. in the normal case, the allocation is done when the lu is created.
> similar/copy
> Create a shallow copy of x: the outer structure is copied, but not all internal values. For example, copying an array produces a new array with identically-same elements as the original.
seems like a reasonable idea to call it copy.

SobhanMP:
should i push when all green?

ViralBShah:
Ok from me. @Wimmerer Thoughts (although he is quite likely to be busy until JuliaCon finishes)? Let's give this a couple of days.

Wimmerer (approved):

PR #188, opened 2022-07-17
update readme

SobhanMP:


ViralBShah:
Why don't we just make the change in Julia upstream to set SuiteSparse to the right version? Can you make a PR? that should then also greatly simplify these comments.

Wimmerer:
Did we not merge one of the recent bumpstdlibs?

ViralBShah:
Yes, we have indeed bumped. I think this explanation can be greatly simplified.

SobhanMP:
it's not up to date in release-1.8 (what i usually use) https://github.com/JuliaLang/julia/blob/release-1.8/stdlib/SuiteSparse.version :shrug:

ViralBShah:
Fair enough, but maybe the title should be `Developing SparseArrays master with Julia master`. You really can't use the master of SparseArrays on 1.8 anyways, since there are lots of structural changes and 1.8 is almost released.

SobhanMP:
i've been using it, works like a charm

SobhanMP:
how does it look now? or should i also remove the last line?

ViralBShah:
This is fine, but if referring to the trick in Pkg.jl, we should link to a file that has the actual trick (rather than just a pointer to the Pkg.jl repo).

SobhanMP:
it becomes a bit long IMHO

PR #192, opened 2022-07-19
Tweak README instruction about running development version

SobhanMP:


SobhanMP:
ok so turns out i did not understand PKG_GIT... it's pointless for deps that aren't in Julialang/julia

SobhanMP:
@ViralBShah

fredrikekre:
Why doesn't `DEPS_GIT="SparseArrays"` work?

SobhanMP:
because it's checkouts the SHA1 commit set in SparseArrays.version. i tested it 3 times

ref:
https://github.com/JuliaLang/julia/blob/db570dfb8d7607d98e71d092a2933a44bf4e3f55/deps/tools/git-external.mk#L51

fredrikekre:
Yes, as noted in the Pkg note https://github.com/JuliaLang/Pkg.jl/blob/master/README.md#using-the-development-version-of-pkgjl, but it gives you a git repo which you can do whatever you want with, i.e. checkout master if you like.

SobhanMP:
but what's the point? it's going to give you a git repo no matter what?

SobhanMP:
ok i just saw your message on slack.

i guess it would be in the line of

```
make DEPS_GIT="SparseArrays" -C deps getall
```
checkout whatever commit manually in stdlib/SparseArrays
then
```
make DEPS_GIT="SparseArrays"
```

ViralBShah:
@fredrikekre What's the recommendation here for the README?

SobhanMP:
ok the instructions should be clear and correct.

README.md
```diff
@@ -22,11 +22,11 @@ SparseArrays.jl provides functionality for working with sparse arrays in Julia.
 
 ## Using newer version of this package with julia master
 
-You need to build Julia from scratch to use the git version (or other version) of this package. The build process is the same, but `DEPS_GIT` should be set when building i.e. `make DEPS_GIT="SparseArrays" ...`. The other option is to manually select the commit in `stdlib/SparseArrays.version`. 
+To use a newer version of this package, you need to build Julia from scratch. The build process is the same as any other build except that `DEPS_GIT="SparseArrays"` should be passed to `make` i.e. `make DEPS_GIT="SparseArrays"`. Then after the build is complete, the git repo in `stdlib/SparseArrays` can be used to check out the desired commit/branch. Alternatively, you need to change the commit used in `stdlib/SparseArrays.version`. The first method does not require rebuilding julia to update the library.
```
fredrikekre, line 25: What do you mean with 
> The first method does not require rebuilding julia to update the library.

What is the first method referring to?

SobhanMP: using `DEPS_GIT`

fredrikekre: Why doesn't that require a rebuild?

SobhanMP: because if you can checkout the git repo.

SobhanMP: i guess there is no need to rebuild it in both cases but it's more convenient in the latter case? (tho i haven't tested it)

fredrikekre: Yea okay, but only because SparseArrays is not currently in the sysimage.

SobhanMP: should i add that detail?

SobhanMP: if it was in the sysimage, how would the git route look like?

fredrikekre: Same, but you would always have to rebuild.

ViralBShah: I think we can merge this for now - but if we put SparseArrays back in the sysimage, then we'll need to update.

ViralBShah:
@dkarrasch Does this address your question about developing the package (unless I misunderstood it in another PR)?

ViralBShah:
I'm merging this for now. The README can always be improved quickly if necessary.

PR #196, opened 2022-07-25
Speed up `SparseMatrixCSC(::AbstractQ)` constructor and multiplication

### LINKED ISSUE #194: `SPQR.QRSparseQ` does not pretty print
This makes the result of `qr` look inconsistent and implies that `Q` is a dense matrix:
```julia
julia> x = sprand(50, 50, .01)
50×50 SparseMatrixCSC{Float64, Int64} with 30 stored entries:
⎡⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠁⠀⠀⠀⠀⠀⠀⡀⠀⠀⠀⎤ # nice!
⎢⠀⠀⠃⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠄⠀⠀⠀⠀⠀⎥
⎢⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⢀⠀⎥
⎢⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⡀⠀⠀⠂⠀⠀⠀⠀⠠⠀⠀⠠⠀⠀⠀⎥
⎢⠀⠀⠀⠀⠀⠀⠀⠀⠄⠀⠀⠐⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⎥
⎢⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠁⠄⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⎥
⎢⠄⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠁⠀⠀⠀⠀⠀⠀⎥
⎢⠀⠀⠀⠀⠠⠀⠀⠀⠀⠀⠀⠀⠁⠠⠀⠀⡀⠀⠀⠀⠀⠀⠀⠀⠀⎥
⎢⠀⠀⠄⠀⠀⠀⠀⠂⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⡀⠀⠀⠀⎥
⎢⠀⠀⠀⠀⠀⠀⠀⠐⠀⠀⠠⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠁⡀⠀⠀⎥
⎢⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⡁⠀⎥
⎢⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠐⠀⠀⠀⠀⠀⠀⠀⠀⠀⎥
⎣⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⎦

julia> qr(x)
SparseArrays.SPQR.QRSparse{Float64, Int64}
Q factor:
50×50 SparseArrays.SPQR.QRSparseQ{Float64, Int64}:
 1.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  …  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
 0.0  1.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0     0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  1.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0     0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  1.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0     0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  1.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0     0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  1.0  0.0  0.0  0.0  0.0  0.0  0.0  …  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0  1.0  0.0  0.0  0.0  0.0  0.0     0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
 ⋮                        ⋮                        ⋮         ⋱            ⋮                        ⋮                   
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0     0.0  0.0  0.0  0.0  0.0  1.0  0.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0     0.0  0.0  0.0  0.0  0.0  0.0  1.0  0.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  …  0.0  0.0  0.0  0.0  0.0  0.0  0.0  1.0  0.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0     0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  1.0  0.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0     0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  1.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0     0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  1.0  0.0
 0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0     0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  1.0
R factor:
50×50 SparseMatrixCSC{Float64, Int64} with 23 stored entries:
⎡⠑⢄⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠐⠀⠀⠀⠀⠀⠀⎤  # Very nice!
⎢⠀⠀⠑⢄⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⎥
⎢⠀⠀⠀⠀⠑⢄⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⎥
⎢⠀⠀⠀⠀⠀⠀⠑⢄⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠁⠀⎥
⎢⠀⠀⠀⠀⠀⠀⠀⠀⠑⢄⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⎥
⎢⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠁⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⎥
⎢⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⎥
⎢⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⎥
⎢⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⎥
⎢⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⎥
⎢⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⎥
⎢⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⎥
⎣⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⎦
Row permutation:
50-element Vector{Int64}:
 27
 31
 34
 38
 19
 16
 39
  ⋮
 13
 11
 10
  9
  8
  3
  2
Column permutation:
50-element Vector{Int64}:
  1
 10
 15
 16
 17
 21
 22
  ⋮
 40
 41
 42
 44
 46
 49
 50

julia> sparse(ans.Q)
50×50 SparseMatrixCSC{Float64, Int64} with 68 stored entries:
⎡⠑⢄⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⎤ # I think it would be better if Q showed like this
⎢⠀⠀⠑⢄⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⎥
⎢⠀⠀⠀⠀⠑⢄⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⎥
⎢⠀⠀⠀⠀⠀⠀⠑⢄⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⎥
⎢⠀⠀⠀⠀⠀⠀⠀⠀⠑⢄⠀⣀⠔⠉⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⎥
⎢⠀⠀⠀⠀⠀⠀⠀⠀⠀⢠⠛⣤⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⎥
⎢⠀⠀⠀⠀⠀⠀⠀⠀⡔⠁⠀⠀⠑⣤⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⎥
⎢⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠑⢄⠀⠀⠀⠀⠀⠀⠀⠀⠀⎥
⎢⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠑⢄⠀⠀⠀⠀⠀⠀⠀⎥
⎢⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠑⢄⠀⠀⠀⠀⠀⎥
⎢⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠑⢄⠀⠀⠀⎥
⎢⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠑⢄⠀⎥
⎣⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠑⎦
```

SobhanMP:
closes #194

### DIFF
--- a/src/solvers/spqr.jl
+++ b/src/solvers/spqr.jl
@@ -2,7 +2,7 @@
 
 module SPQR
 
-import Base: \
+import Base: \, *
 using Base: require_one_based_indexing
 using LinearAlgebra
 using ..LibSuiteSparse: SuiteSparseQR_C
@@ -167,11 +167,7 @@ julia> A = sparse([1,2,3,4], [1,1,2,2], [1.0,1.0,1.0,1.0])
 julia> qr(A)
 SparseArrays.SPQR.QRSparse{Float64, Int64}
 Q factor:
-4×4 SparseArrays.SPQR.QRSparseQ{Float64, Int64}:
- -0.707107    ⋅          ⋅        -0.707107
-   ⋅        -0.707107  -0.707107    ⋅
-   ⋅        -0.707107   0.707107    ⋅
- -0.707107    ⋅          ⋅         0.707107
+4×4 SparseArrays.SPQR.QRSparseQ{Float64, Int64}
 R factor:
 2×2 SparseMatrixCSC{Float64, Int64} with 2 stored entries:
  -1.41421    ⋅
@@ -284,8 +280,8 @@ function LinearAlgebra.rmul!(A::StridedMatrix, adjQ::Adjoint{<:Any,<:QRSparseQ})
     return A
 end
 
-*(Q::QRSparseQ, B::SparseMatrixCSC) = sparse(Q) * B
-*(A::SparseMatrixCSC, Q::QRSparseQ) = A * sparse(Q)
+(*)(Q::QRSparseQ, B::SparseMatrixCSC) = sparse(Q) * B
+(*)(A::SparseMatrixCSC, Q::QRSparseQ) = A * sparse(Q)
 
 @inline function Base.getproperty(F::QRSparse, d::Symbol)
     if d === :Q
--- a/src/solvers/spqr.jl
+++ b/src/solvers/spqr.jl
@@ -132,7 +132,7 @@ struct QRSparseQ{Tv<:CHOLMOD.VTypes,Ti<:Integer} <: LinearAlgebra.AbstractQ{Tv}
     τ::Vector{Tv}
     n::Int # Number of columns in original matrix
 end
-
+Base.print_array(io::IO, Q::QRSparseQ) = Base.print_array(io, sparse(Q))
 Base.size(Q::QRSparseQ) = (size(Q.factors, 1), size(Q.factors, 1))
 Base.axes(Q::QRSparseQ) = map(Base.OneTo, size(Q))
 
@@ -168,10 +168,10 @@ julia> qr(A)
 SparseArrays.SPQR.QRSparse{Float64, Int64}
 Q factor:
 4×4 SparseArrays.SPQR.QRSparseQ{Float64, Int64}:
- -0.707107   0.0        0.0       -0.707107
-  0.0       -0.707107  -0.707107   0.0
-  0.0       -0.707107   0.707107   0.0
- -0.707107   0.0        0.0        0.707107
+ -0.707107    ⋅          ⋅        -0.707107
+   ⋅        -0.707107  -0.707107    ⋅ 
+   ⋅        -0.707107   0.707107    ⋅ 
+ -0.707107    ⋅          ⋅         0.707107
 R factor:
 2×2 SparseMatrixCSC{Float64, Int64} with 2 stored entries:
  -1.41421    ⋅
@@ -252,6 +252,7 @@ function LinearAlgebra.rmul!(A::StridedMatrix, Q::QRSparseQ)
     end
     return A
 end
+
 
 function LinearAlgebra.lmul!(adjQ::Adjoint{<:Any,<:QRSparseQ}, A::StridedVecOrMat)
     Q = adjQ.parent
--- a/src/solvers/spqr.jl
+++ b/src/solvers/spqr.jl
@@ -132,7 +132,7 @@ struct QRSparseQ{Tv<:CHOLMOD.VTypes,Ti<:Integer} <: LinearAlgebra.AbstractQ{Tv}
     τ::Vector{Tv}
     n::Int # Number of columns in original matrix
 end
-Base.print_array(io::IO, Q::QRSparseQ) = Base.print_array(io, sparse(Q))
+
 Base.size(Q::QRSparseQ) = (size(Q.factors, 1), size(Q.factors, 1))
 Base.axes(Q::QRSparseQ) = map(Base.OneTo, size(Q))
 
--- a/src/solvers/spqr.jl
+++ b/src/solvers/spqr.jl
@@ -169,8 +169,8 @@ SparseArrays.SPQR.QRSparse{Float64, Int64}
 Q factor:
 4×4 SparseArrays.SPQR.QRSparseQ{Float64, Int64}:
  -0.707107    ⋅          ⋅        -0.707107
-   ⋅        -0.707107  -0.707107    ⋅ 
-   ⋅        -0.707107   0.707107    ⋅ 
+   ⋅        -0.707107  -0.707107    ⋅
+   ⋅        -0.707107   0.707107    ⋅
  -0.707107    ⋅          ⋅         0.707107
 R factor:
 2×2 SparseMatrixCSC{Float64, Int64} with 2 stored entries:
@@ -253,7 +253,6 @@ function LinearAlgebra.rmul!(A::StridedMatrix, Q::QRSparseQ)
     return A
 end
 
-
 function LinearAlgebra.lmul!(adjQ::Adjoint{<:Any,<:QRSparseQ}, A::StridedVecOrMat)
     Q = adjQ.parent
     if size(A, 1) != size(Q, 1)
@@ -285,6 +284,9 @@ function LinearAlgebra.rmul!(A::StridedMatrix, adjQ::Adjoint{<:Any,<:QRSparseQ})
     return A
 end
 
+*(Q::QRSparseQ, B::SparseMatrixCSC) = sparse(Q) * B
+*(A::SparseMatrixCSC, Q::QRSparseQ) = A * sparse(Q)
+
 @inline function Base.getproperty(F::QRSparse, d::Symbol)
     if d === :Q
         return QRSparseQ(F.factors, F.τ, size(F, 2))
@@ -312,6 +314,9 @@ function Base.show(io::IO, mime::MIME{Symbol("text/plain")}, F::QRSparse)
     show(io, mime, F.prow)
     println(io, "\nColumn permutation:")
     show(io, mime, F.pcol)
+end
+function Base.show(io::IO, ::MIME{Symbol("text/plain")}, Q::QRSparseQ)
+    summary(io, Q)
 end
 
 # With a real lhs and complex rhs with the same precision, we can reinterpret
--- a/src/sparsematrix.jl
+++ b/src/sparsematrix.jl
@@ -749,69 +749,20 @@ SparseMatrixCSC{Tv,Ti}(M::Adjoint{<:Any,<:AbstractSparseMatrixCSC}) where {Tv,Ti
 SparseMatrixCSC{Tv,Ti}(M::Transpose{<:Any,<:AbstractSparseMatrixCSC}) where {Tv,Ti} = SparseMatrixCSC{Tv,Ti}(copy(M))
 
 # we can only view AbstractQs as columns
-SparseMatrixCSC{Tv, Ti}(Q::LinearAlgebra.AbstractQ) where {Tv, Ti} = _from_lmul(Tv, Ti, Q; sizehint=false)
-function _from_eachcol(Tv, Ti, M; sizehint=false)
-    colptr = zeros(Ti, size(M, 2) + 1)
-    nzval = Vector{Tv}(undef, 0)
-    rowval = Vector{Ti}(undef, 0)
-    if sizehint
-        nz = 0
-        for col in eachcol(M)
-            for i in col
-                if iszero(i) == false # should be _isnotzero(v)
-                    nz += 1
-                end
-            end
-        end
-        sizehint!(nzval, nz)
-        sizehint!(rowval, nz)
-    end
-    colptr[1] = 1
-    ind = 1
-    for (j, col) in enumerate(eachcol(M))
-        for (i, v) in enumerate(col)
-            if iszero(v) == false
-                push!(nzval, v)
-                push!(rowval, i)
-                ind += 1
-            end
-        end
-        colptr[j + 1] = ind
-    end
-    return SparseMatrixCSC{Tv, Ti}(size(M)..., colptr, rowval, nzval)
-end
-
-function _from_lmul(Tv, Ti, M; sizehint=false)
-    
-    colptr = zeros(Ti, size(M, 2) + 1)
-    nzval = Vector{Tv}(undef, 0)
-    rowval = Vector{Ti}(undef, 0)
-    col = zeros(eltype(M), size(M, 1))
-    
-    if sizehint
-        nz = 0
-        for j in axes(M, 2)
-            fill!(col, false)
-            col[j] = one(Tv)
-            lmul!(M, col)
-            for i in col
-                if iszero(i) == false # should be _isnotzero(v)
-                    nz += 1
-                end
-            end
-        end
-        sizehint!(nzval, nz)
-        sizehint!(rowval, nz)
-    end
+function SparseMatrixCSC{Tv,Ti}(Q::LinearAlgebra.AbstractQ) where {Tv,Ti}
+    colptr = zeros(Ti, size(Q, 2) + 1)
+    nzval = Tv[]
+    rowval = Ti[]
+    col = zeros(eltype(Q), size(Q, 1))
 
     colptr[1] = 1
     ind = 1
-    for j in axes(M, 2)
+    for j in axes(Q, 2)
         fill!(col, false)
         col[j] = one(Tv)
-        lmul!(M, col)
+        lmul!(Q, col)
         for (i, v) in enumerate(col)
-            if iszero(v) == false
+            if iszero(v) == false # should be _isnotzero(v)
                 push!(nzval, v)
                 push!(rowval, i)
                 ind += 1
@@ -819,7 +770,7 @@ function _from_lmul(Tv, Ti, M; sizehint=false)
         end
         colptr[j + 1] = ind
     end
-    return SparseMatrixCSC{Tv, Ti}(size(M)..., colptr, rowval, nzval)
+    return SparseMatrixCSC{Tv,Ti}(size(Q)..., colptr, rowval, nzval)
 end
 
 # converting from SparseMatrixCSC to other matrix types
@@ -1211,9 +1162,9 @@ end
 """
     ftranspose!(X::AbstractSparseMatrixCSC{Tv,Ti}, A::AbstractSparseMatrixCSC{Tv,Ti}, f::Function) where {Tv,Ti}
 
-Transpose `A` and store it in `X` while applying the function `f` to the non-zero elements. 
+Transpose `A` and store it in `X` while applying the function `f` to the non-zero elements.
 Does not remove the zeros created by `f`. `size(X)` must be equal to `size(transpose(A))`.
-No additonal memory is allocated other than resizing the rowval and nzval of `X`, if needed. 
+No additonal memory is allocated other than resizing the rowval and nzval of `X`, if needed.
 
 See `halfperm!`
 """
@@ -1232,9 +1183,9 @@ end
 """
     transpose!(X::AbstractSparseMatrixCSC{Tv,Ti}, A::AbstractSparseMatrixCSC{Tv,Ti}) where {Tv,Ti}
 
-Transpose the matrix `A` and stores it in the matrix `X`. 
+Transpose the matrix `A` and stores it in the matrix `X`.
 `size(X)` must be equal to `size(transpose(A))`.
-No additonal memory is allocated other than resizing the rowval and nzval of `X`, if needed. 
+No additonal memory is allocated other than resizing the rowval and nzval of `X`, if needed.
 
 See `halfperm!`
 """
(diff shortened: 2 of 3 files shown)


SobhanMP:
is there any reason to avoid force push in branches like this?

dkarrasch:
This doesn't look right. `Q` is a 4x4 matrix, but now you display some 4x2 matrix.

src/solvers/spqr.jl
```diff
@@ -132,7 +132,7 @@ struct QRSparseQ{Tv<:CHOLMOD.VTypes,Ti<:Integer} <: LinearAlgebra.AbstractQ{Tv}
     τ::Vector{Tv}
     n::Int # Number of columns in original matrix
 end
-
+Base.print_array(io::IO, Q::QRSparseQ) = Base.print_array(io, Q.factors)
```
dkarrasch, line 135: This should perhaps be
```suggestion
Base.print_array(io::IO, Q::QRSparseQ) = Base.print_array(io, sparse(Q))
```
or something like this (see the corresponding issue) which demonstrates the required effort.

SobhanMP:
```
using LinearAlgebra, SparseArrays, Random, BenchmarkTools
Random.seed!(131);
A = sparse(I + sprandn(100, 100, 1/100));
q = qr(A);
@benchmark display($q.Q)
BenchmarkTools.Trial: 52 samples with 1 evaluation.
 Range (min … max):  14.399 ms …    1.835 s  ┊ GC (min … max): 0.00% …  0.00%
 Time  (median):     19.071 ms               ┊ GC (median):    6.26%
 Time  (mean ± σ):   98.233 ms ± 336.622 ms  ┊ GC (mean ± σ):  2.24% ± 10.47%

  █                                                             
  █▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▄▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▄ ▁
  14.4 ms       Histogram: log(frequency) by time        1.5 s <

 Memory estimate: 22.70 MiB, allocs estimate: 352056.

Base.print_array(io::IO, Q::SparseArrays.SPQR.QRSparseQ) = Base.print_array(io, sparse(Q))
@benchmark display($q.Q)

BenchmarkTools.Trial: 14 samples with 1 evaluation.
 Range (min … max):  312.287 ms … 445.235 ms  ┊ GC (min … max): 4.86% … 8.65%
 Time  (median):     355.080 ms               ┊ GC (median):    7.42%
 Time  (mean ± σ):   371.133 ms ±  47.507 ms  ┊ GC (mean ± σ):  7.66% ± 1.32%

  ▁ ▁  ▁   ▁  ▁▁   ▁   ▁            ▁     ▁ ▁             ▁   █  
  █▁█▁▁█▁▁▁█▁▁██▁▁▁█▁▁▁█▁▁▁▁▁▁▁▁▁▁▁▁█▁▁▁▁▁█▁█▁▁▁▁▁▁▁▁▁▁▁▁▁█▁▁▁█ ▁
  312 ms           Histogram: frequency by time          445 ms <

 Memory estimate: 763.71 MiB, allocs estimate: 32155.
```

is the performance loss acceptable?

SobhanMP: after looking at the code it seem that the get index is really not very efficient and the call to sparse gets redirected to the abstract implementation which means that there is a vector being created for every value look up. 

i'll try making the conversion to sparse faster by re-using the workspace.

dkarrasch: > i'll try making the conversion to sparse faster.

Good luck! We have a

```julia
Matrix{T}(Q::QRSparseQ) where {T} = lmul!(Q, Matrix{T}(I, size(Q, 1), min(size(Q, 1), Q.n)))
```

method (and corresponding `[l/r]mul!` methods), which suggests that you need some multiplication code to obtain any matrix representation, be it sparse or dense. For printing some `QRSparseQ` object, I guess it's actually better to use that dense matrix representation, because it comes out naturally and I'm not sure how sparse the Q matrix generically is. For printing some `QRSparse` object, I wonder if we want to skip printing the Q factor as a matrix, because, well, it kind of actually isn't.

SobhanMP:
after some looking up i think i found the problem. `itertate`  of `QRsparseQ` is implemented by AbstractQ with `lmul!`. one lmul! per index to get the column and then take the right elemenent


by handling it on a column per column basis, the run time for a random 100x100 matrix goes from `263.841 ms (20014 allocations: 763.41 MiB)` to `5.943 ms (390189 allocations: 25.96 MiB)`

src/sparsematrix.jl
```diff
@@ -748,6 +748,40 @@ SparseMatrixCSC{Tv}(M::Transpose{<:Any,<:AbstractSparseMatrixCSC}) where {Tv} =
...
+    for (j, col) in enumerate(eachcol(M))
+        for (i, v) in enumerate(col)
+            if iszero(v) == false
+                push!(nzval, v)
+                push!(rowval, i)
+                ind += 1
+            end
+        end
+        colptr[j + 1] = ind
+    end
```
dkarrasch, line 779: Doesn't that still index into each column of `Q` element per element? I guess what you want is something like
```suggestion
    for (j, colind) in enumerate(axes(M, 2))
        col = M[:, colind] # this could be further optimized by allocating col once and lmul! with Q
        for i in axes(v, 1)
            if !iszero(v) # should be _isnotzero(v)
                push!(nzval, v)
                push!(rowval, i)
                ind += 1
            end
        end
        colptr[j + 1] = ind
    end
```

SobhanMP: it does what i want already but i guess i can do  `size(Q, 2) - 1` fewer allocations. by accessing it first with `eachcol(M)` then on each element, i'm not doing a getindex (which loads the whole column) for every element.

dkarrasch: The elements in the iterator of eachcol are views into the columns, but remember that `Q` is not stored as a matrix, the columns are computed by lmul!-ing the canonical vectors. So I don't see how it should know to materialize that view via `getindex(Q, :, j)`.

SobhanMP: @dkarrasch i've implemented it like you said, it's faster but i don't understand why it's faster. As far i as understand, going over columns should only call lmul! once per column, right? or 

```
julia> @btime SparseArrays._from_lmul(Float64, Int, $q.Q);
  74.427 μs (3991 allocations: 197.88 KiB)

julia> @btime SparseArrays._from_col(Float64, Int, $q.Q);
  5.616 ms (390190 allocations: 25.96 MiB)
```

dkarrasch: `eachcol` does not call `getindex(Q, :, j)` for `j in ...`, which would make use of that `lmul!` "trick". It just creates a view into the columns, which are thin wrappers, no "colwise" magic behind the scenes. So, when you start to iterate over a view into one column, it ends up with scalar indexing into the parent array, which is exactly what you want to avoid.

dkarrasch: BTW, I would restrict that "`SparseMatrixCSC`" method to `AbstractQ`s directly. For matrices that are materialized in memory, you don't even want to consider this. Same for structured matrices, which most likely already have their own `SparseMatrixCSC` constructor.

SobhanMP: i've already done that no? 
```julia
SparseMatrixCSC{Tv, Ti}(Q::LinearAlgebra.AbstractQ) where {Tv, Ti} = _from_lmul(Tv, Ti, Q; sizehint=false)
```
i think putting a type on _from_lmul/_from_eachcol from makes it too narrow. i think the name is enough?

BTW. should i make a patch for `eachcol(AbstractQ)` or is it over specialization?

dkarrasch: Sorry, I might have missed the (then) latest changes. For experimenting its fine to have it like you have now, but eventually I would move the long function body right under the `SparseMatrixCSC` header, that's what I meant.

> BTW. should i make a patch for eachcol(AbstractQ) or is it over specialization?

The underlying problem is that `AbstratQ`s are not really matrices, and the expectation is that `eachcol` is just a thin iterator wrapper, not something that comes with a potentially significant computation price. I'm experimenting locally with removing the subtyping `AbstractQ <: AbstractMatrix`, because that has caused other issues in the past.

SobhanMP: > SparseMatrixCSC header, that's what I meant.
done i think.

i feel like removing the AbstractMatrix is going to break a bunch of stuff (that have poor performance) in random packages. I think having an `eachcol` is the right balance between writing complicated stuff and performance. i.e.
```julia
Base.eachcol(Q::AbstractQ) = (Q[:, i] for i in axes(Q, 2))
```

ViralBShah:
> is there any reason to avoid force push in branches like this?

Force pushing in branches is fine.

src/sparsematrix.jl
dkarrasch: But we're not using `eachcol(::AbstractQ)`, are we? The proposal is a good, but it seems to me that it doesn't follow the `eachcol` logic, it behaves differently. The docstring says that it returns views into the columns, so in particular you can write into them and thereby modify the underlying array, whic is not the case here. I'd suggest to avoid adding anything that is not really necessary for the current issue.

dkarrasch:
As for the original issue, I think we'll need to change

https://github.com/JuliaSparse/SparseArrays.jl/blob/d367f11157f1f36a1b566900ab1f5e5591dfdcca/src/solvers/spqr.jl#L306-L307

to something like

```julia
println(io, "Q factor: " * summary(F.Q))
```

or, alternatively, modify the display of `QRSparseQ` to

```julia
function Base.show(io::IO, mime::MIME{Symbol("text/plain")}, Q::QRSparseQ)
    summary(io, Q)
end
```

to only show the summary. Really, I think we shouldn't make it look like `Q`s are like arrays and shouldn't display any "elements".

test/spqr.jl
```diff
@@ -137,6 +139,15 @@ end
     @test all(iszero, (rank(spzeros(10, i)) for i in 1:10))
 end
 
+
+@testset "sparse" begin
+    A = I + sprandn(100, 100, 0.01)
+    Q = qr(A).Q
+    sQ = sparse(Q)
+    @test sQ == sparse(Matrix(Q))
+    @test sQ ≈ sparse(qr(Matrix(A)).Q)
```
SobhanMP, line 148: this is going to always fail, qr(::Sparse) is going to returned a permutated qr

SobhanMP: another problem is that the R and Q can be multiplied by -1 w/o changing the correctness.

SobhanMP:
IMHO leaving `_from_lmul!/_from_col` as a function was ok.

test/spqr.jl
dkarrasch: Yea, that was too quick. Thanks for fixing.

dkarrasch:
> IMHO leaving `_from_lmul!`/`_from_col` as a function was ok.

What for? I don't think we should "deposit" unused internal functions just for fun. I also removed the sizehint option because that was performing the expensive work (multiplication) twice, so I don't think that would be worth it in any case.

src/solvers/spqr.jl
```diff
@@ -132,7 +132,7 @@ struct QRSparseQ{Tv<:CHOLMOD.VTypes,Ti<:Integer} <: LinearAlgebra.AbstractQ{Tv}
     τ::Vector{Tv}
     n::Int # Number of columns in original matrix
 end
-
+Base.print_array(io::IO, Q::QRSparseQ) = Base.print_array(io, sparse(Q))
```
dkarrasch, line 135: I'm not sure we need this anymore. Neither `show` of `QRSParse` nor of `QRSparseQ` displays any kind of matrix representation for `Q`. Are there other paths that lead to this where this could be useful?

dkarrasch: The more I think about it, the more I believe this should be removed. Who wants to see a matrix representation can simply call `sparse(Q)` and that's it.

dkarrasch:
The current constructor does not apply to adjoints of `AbstractQ`s, but would apply automatically in case https://github.com/JuliaLang/julia/pull/46196 gets accepted. Also, the constructor turns out to be useful in multiplication. Previously, `F.Q * F.R` would fall back to `_generic_matmul!`! 😱  But now it makes use of the sparse constructor. 🎉

SobhanMP:
> > IMHO leaving `_from_lmul!`/`_from_col` as a function was ok.
> 
> What for? I don't think we should "deposit" unused internal functions just for fun. I also removed the sizehint option because that was performing the expensive work (multiplication) twice, so I don't think that would be worth it in any case.

i agree with remove the sizehint but the function `SparseMatrixCSC(::AbstractQ)` isn't really a property of AbstractQs but anything that implement lmul!

dkarrasch:
> isn't really a property of AbstractQs but anything that implement lmul!

That's right, up to the restriction that additionally there is no direct/fast `getindex` and no iteration protocol. If we find more of such types, we can add them to the signature. Do you have something in mind already?

SobhanMP:
apart from matrix free jacobians, no.

ViralBShah:
Off topic - We really do need to reduce the inertia on this package (and I'm doing my best to help speed things along). Some inertia is required since it is still part of Julia stdlibs and widely depended upon. What else can we do?

dkarrasch:
Sounds good. Let's have a documented internal function then. 😉 And bikeshed the name! I'll make a proposal and then we see.

SobhanMP:
looks good to me!

PR #197, opened 2022-07-25
don't skip lock in deserialize umfpacklu

SobhanMP:
de-serialize didn't de-serialize the lock. maybe there should be some test to make sure de-serialize consumes everything?

### DIFF
--- a/test/umfpack.jl
+++ b/test/umfpack.jl
@@ -365,8 +365,25 @@ end
         for nm in (:colptr, :m, :n, :nzval, :rowval, :status)
             @test getfield(F1, nm) == getfield(F2, nm)
         end
+        for nm in (:W, :Wi)
+            @test getfield(F1.workspace, nm) == getfield(F2.workspace, nm)
+        end
+        b1 = IOBuffer()
+        serialize(b1, (a=F1, b=F2))
+        seekstart(b1)
+        x = deserialize(b1)
+        for nm in (:colptr, :m, :n, :nzval, :rowval, :status)
+            @test getfield(F1, nm) == getfield(x.a, nm) == getfield(x.b, nm)
+        end
+        for nm in (:W, :Wi)
+            @test getfield(x.a.workspace, nm) == getfield(F1.workspace, nm)
+            @test getfield(x.b.workspace, nm) == getfield(F2.workspace, nm)
+        end
+
         umfpack_report(F1)
         umfpack_report(F2)
+        umfpack_report(x.a)
+        umfpack_report(x.b)
     end
 
     @testset "Reuse symbolic LU factorization" begin
--- a/src/solvers/umfpack.jl
+++ b/src/solvers/umfpack.jl
@@ -466,7 +466,7 @@ function show(io::IO, mime::MIME{Symbol("text/plain")}, F::UmfpackLU)
     end
 end
 
-function deserialize(s::AbstractSerializer, t::Type{UmfpackLU{Tv,Ti}}) where {Tv,Ti}
+function deserialize(s::AbstractSerializer, ::Type{UmfpackLU{Tv,Ti}}) where {Tv,Ti}
     symbolic = deserialize(s)
     numeric  = deserialize(s)
     m        = deserialize(s)
@@ -478,10 +478,10 @@ function deserialize(s::AbstractSerializer, t::Type{UmfpackLU{Tv,Ti}}) where {Tv
     workspace= deserialize(s)
     control  = deserialize(s)
     info     = deserialize(s)
-    deserialize(s) # lock
+    lock     = deserialize(s)
     obj      = UmfpackLU{Tv,Ti}(symbolic, numeric, m, n,
         colptr, rowval, nzval, status,
-        workspace, control, info, ReentrantLock())
+        workspace, control, info, lock)
 
     finalizer(umfpack_free_symbolic_nl, obj)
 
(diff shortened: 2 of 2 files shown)


SobhanMP:
@ViralBShah i'd recommend upgrading julia's sparse array to include this patch. a nasty surprise awaits anyone who serializes UmfpackLUs

```
julia> a = lu(I + sprandn(100, 100, 0.1));

julia> serialize("a", (a=a, b=a))

julia> deserialize("a")
ERROR: TypeError: in new, expected SparseArrays.UMFPACK.UmfpackLU{Float64, Int64}, got a value of type ReentrantLock
Stacktrace:
```

ViralBShah:
Submit a PR?

SobhanMP:
what do you mean? this is a pr, right?

ViralBShah:
Oh I misunderstood. Ok, we can bump everything together. Let's fix as many of the other issues that we know that are plaguing our CI in preparation.

ViralBShah:
Let's see how many of the current batch of PRs we can get to in the next few days, and then merge and update.

ViralBShah:
@Wimmerer Merge? Let's tag and release with this one?

ViralBShah:
@Wimmerer I believe you should have the priveleges to bump SparseArrays.jl

Wimmerer:
I do, give me a few minutes to do so.

test/umfpack.jl
```diff
@@ -365,8 +365,25 @@ end
         for nm in (:colptr, :m, :n, :nzval, :rowval, :status)
             @test getfield(F1, nm) == getfield(F2, nm)
         end
+        for nm in (:W, :Wi)
+            @test getfield(F1.workspace, nm) == getfield(F2.workspace, nm)
```
Keno, line 369: msan complains about this because the workspace is not initialized memory. Does the workspace need to survive serialization or is it just scratch memory? If it's just scratch, maybe we shouldn't serialize it at all?

Wimmerer: It's just scratched, there's no reason to serialize it.

Wimmerer: I'll remove it from the serialization function here in a bit.

SobhanMP: but that wouldn't that break the semantic of serialization? suppose there are two UMFPACKLUs with shaped workspace.

Wimmerer: Can you elaborate Sobhan?

SobhanMP: well you can define two umfpacklus that share the workspace not sure what to elaborate. if you just reallocate it, it will not be the same object when de-serialized. 

something like 
L1 = lu(A)
L2 = UmfpackLU(L1.n, L1.m, another zero index sparse matrix, L1.workspace, l1.lock) 
this would work fine.

i don't think you can just reallocate. Also the size of the workspace is not determined by us. it's resized by lu (given the config) but it could have the wrong size. so by reallocating it, you might also accidentally resize it. My point is that there is no guarantee that not serializing wont break "working" code. a better idea might be to just check for length of workspace in that test.

Keno: Serializing copies the workspace/lock anyway, it doesn't restore a reference to the original objects, so as long as the content doesn't matter to the semantics of UmfpackLU, reallocating it is fine.

Wimmerer: We should likely just do `serialize((size(lu.workspace.Wi), size(lu.workspace.W)))`, and provide the missing constructor to reallocate `UmfpackWS` from that tuple.

SobhanMP: > Serializing copies the workspace/lock anyway,

even if you serialize `(L1, L2)` ?

SobhanMP: works fine on master tho
```julia
using SparseArrays, Serialization, LinearAlgebra
a = lu(I + sprandn(10, 10, 0.1))
b = SparseArrays.UMFPACK.UmfpackLU(C_NULL, C_NULL, a.m, a.n,
    copy(a.colptr), copy(a.rowval), copy(a.nzval), 0, a.workspace, a.control, a.info, a.lock)
serialize("a.jld", (a, b))
x, y = deserialize("a.jld")
x.workspace == y.workspace
```
gives `true`

Wimmerer: I'm not really convinced that's an important feature for a workspace.

Regardless, changing the test to satisfy msan doesn't fix the underlying issue of serializing and deserializing uninitialized memory as far as I understand.

SobhanMP: > I'm not really convinced that's an important feature for a workspace.

i think not sanctifying this (regardless of usefulness), is a "wrong" in the sense that it does not respect the semantic of serialization. it's like saying that uninitialized vectors can be recreated on the fly instead (assuming there was a way for checking for that).

> underlying issue of serializing 

i'm not sure i understand the issue, checking for uninitialized memory reads is way of finding potential bugs and not a bug itself. any 64 bits is a valid int64 or float64. the problem is that it's possible that there is nan somewhere which i think i fixed a while back.

see https://github.com/JuliaSparse/SparseArrays.jl/blob/3b1ce54fe4575913140cc4f3b001d456da6091e9/test/umfpack.jl#L366

SobhanMP: also i messed up what i meant to use in the snippet above was
```julia
x.workspace === y.workspace
```
(still returns true)

Issue #178, opened 2022-07-11
CI: non-deterministic failure in SparseArrays/higherorderfns


DilumAluthge:
E.g. https://buildkite.com/julialang/julia-master/builds/13661#0181e2d2-87e0-4225-88a1-d953c29fce1f

```
Error in testset SparseArrays/higherorderfns:
Test Failed at /cache/build/default-amdci4-1/julialang/julia-master/julia-e750ba7d9a/share/julia/stdlib/v1.9/SparseArrays/test/higherorderfns.jl:25
Expression: #= /cache/build/default-amdci4-1/julialang/julia-master/julia-e750ba7d9a/share/julia/stdlib/v1.9/SparseArrays/test/higherorderfns.jl:25 =# @allocated(map!(sin, X, A)) == 0
Evaluated: 320 == 0
```

ViralBShah:
Is this happening repeatedly?

DilumAluthge:
I've seen it a couple different times, but it is pretty uncommon.

SobhanMP:
this is my guess for the cause, non-structural zeros causes sparse(Array(A)) to not have the exact structure as A. tho i'm not having a lot of success making rand() generate zeros.
https://github.com/JuliaSparse/SparseArrays.jl/pull/199

SobhanMP:
@DilumAluthge please open it again if it happens again.

PR #199, opened 2022-07-27
fix non-determinstic failure in higherorderfns (?)

SobhanMP:
Possible fix for #178. My guess is that `sprandn` is returning non-structural zeros. This patch removes those non-structural zeros.


Wimmerer:
Can we just use some small known sparse matrix for this instead of using sprand until we hit one with the desired properties?

dkarrasch:
Can't you simply `setindex!` elements to either zero or something that yields zero under function application? I mean, under the assumption that zeros and/or zero values cause the issue.

SobhanMP:
sure, that would also work

SobhanMP:
i think this is the best solution. the change is such that if this is not the solution, the tests will keep failing.

test/higherorderfns.jl
```diff
@@ -8,24 +8,35 @@ module HigherOrderFnsTests
...
+    # --> test map entry point
+    fA = Array(A)
+    shapeA = size(A)
+    @test map(sin, A) == sparse(map(sin, fA))
+    @test map(cos, A) == sparse(map(cos, fA))
+    # --> test map! entry point
+    fX = copy(fA); X = similar(A)
+    map!(sin, X, A); X = similar(A) # warmup for @allocated
+    map!(sin, X, A); X = similar(A) # warmup for @allocated
+    @test alloc_tests || @allocated(map!(sin, X, A)) == 0
```
dkarrasch, line 25: Is this meant to be
```suggestion
    alloc_tests && @test @allocated(map!(sin, X, A)) == 0
```
?

SobhanMP: does it really matter?

SobhanMP: it was mean to be the other way around

ViralBShah:
Can we update this PR to say which issue it is closing and put in a little bit description in the OP (the first comment in the PR which is empty)?

ViralBShah (approved):

test/higherorderfns.jl
```diff
@@ -8,24 +8,39 @@ module HigherOrderFnsTests
 
 using Test
 using SparseArrays
+using SparseArrays: nonzeros
```
dkarrasch, line 11: Seems like `nonzeros` is already exported by `SparseArrays`.
```suggestion
```

test/higherorderfns.jl
```diff
@@ -8,24 +8,39 @@ module HigherOrderFnsTests
...
+    shapeA = size(A)
+    @test map(sin, A) == sparse(map(sin, fA))
+    @test map(cos, A) == sparse(map(cos, fA))
+    # --> test map! entry point
+    fX = copy(fA); X = similar(A)
+    map!(sin, X, A); X = similar(A) # warmup for @allocated
+    map!(sin, X, A); X = similar(A) # warmup for @allocated
+    if @allocated(map!(sin, X, A)) == 0 || alloc_tests
+        println(stderr, "A in test_map_and_map! is ", A)
+    end
```
dkarrasch, line 27: I'm trying to follow the logic here, but fail. IIUC, you want this to print when `alloc_tests == false`(the third call below), which will happen when `@allocated(map!(sin, X, A)) == 0`? Do you mean
```suggestion
    if !(@allocated(map!(sin, X, A)) == 0 || alloc_tests)
        println(stderr, "A in test_map_and_map! is ", A)
    end
```
`alloc_tests = true` sounds to me like "do the allocation tests", whereas you seem to use it as "skip the allocation tests`.

dkarrasch (commented):
Just some minor comments, otherwise LGTM.

Issue #189, opened 2022-07-17
Ability to turn on and off scalar indexing for sparse matrices?


ChrisRackauckas:
`A[i,j] = ...` is very slow if `A isa SparseMatrixCSC`. In the wild, there's another case like this with CUDA.jl. In that package, it allows for doing `CUDA.allowscalar(false)` so that any `A[i,j]` operation (`getindex` and `setindex!`) causes an error to be thrown. This is helpful for finding unoptimized fallbacks which would otherwise "work" given the AbstractArray interface, but would be detrimental to performance. Without this, finding out which CUDA codes are hitting unwanted fallbacks can only be done by profiling.

It would be nice to have similar functionality available on SparseMatrixCSC. I am not sure I like the idea of having it global though, since then it's global, but if it's not global you need rules for how to carry it forward with broadcast etc. which might be hard to do in general and so the global is the easy option.

Wimmerer:
Is the idea to use this only for testing? I can't imagine it would be a good idea to do this in a package like SciML, disabling it for everything else.

ChrisRackauckas:
Yes, I'd only use it for testing. The main thing would be to ensure we don't hit any fallback that are looping scalar-wise.

ViralBShah:
If you are using with CUDA, you probably want to disable it, right? Packages can opt in to do that, I suppose.

ChrisRackauckas:
Yes, with CUDA its generally the case that you say "if using CUDA, then put `CUDA.allowscalar(false)` in your script because that will catch common performance pitfalls and turn them into errors". The similar fact is true for sparse matrices that if you're doing `A[i,j]` operations all over the place then you'll have bad performance, but there isn't a similar way with SparseMatrixCSC to turn performance pitfalls into errors. 

Of course it can be used incorrectly, the CUDA.jl one could be used incorrectly as well if a package sets it "for the user" and suddenly non-performance sensitive parts of the code are overly cautious, but I think over the years people have found `CUDA.allowscalar(false)` useful enough for testing and scripting safety that it would be nice to have a similar feature for sparse matrices.

Wimmerer:
I definitely prefer the StaticSparseMatrixCSC <: AbstractSparseMatrixCSC route. But it is a little more onerous to implement.

ChrisRackauckas:
What would you do with `A::StaticSparseMatrixCSC .+ B::SparseMatrixCSC`?

Wimmerer:
[...]

PR #200, opened 2022-07-28
add allowscalar

SobhanMP:
closes https://github.com/JuliaSparse/SparseArrays.jl/issues/189

the only downside is that revise might act a bit weird. 

also maybe RCI should call `@eval` on `exp` direclty instead of returning `exp`?

another detail is that should this also block stuff like
```
A = sprandn(100, 100, 0.1)
A[i, :]
A[1:10, :]
```
or just the ones i marked?

This does NOT cause any degradation in any way, unlike adding a state variable.


Wimmerer:
You mean it causes a 2% slow down?

SobhanMP:
no it does not, Chris said that adding a flag for scalar (with a global `const scalar = Ref{Bool}(true)`) would cause a 2% slowdown. this does not do that.

Wimmerer:
Are we doing both the static type and the macro?

SobhanMP:
what static type?

SobhanMP:
bump

ViralBShah (approved):

ViralBShah:
I think the `allowscalar` API should be documented and marked experimental (and not exported).

Issue #190, opened 2022-07-17
Ability to fix the sparsity pattern of a sparse matrix


ChrisRackauckas:
We had a long discussion in Slack over the following example:

```julia
julia> using LinearAlgebra, SparseArrays

julia> J = sparse(Diagonal(ones(4)))
4×4 SparseMatrixCSC{Float64, Int64} with 4 stored entries:
 1.0   ⋅    ⋅    ⋅
  ⋅   1.0   ⋅    ⋅
  ⋅    ⋅   1.0   ⋅
  ⋅    ⋅    ⋅   1.0

julia> W = sparse(Diagonal(ones(4)));

julia> J[4,4] = 0
0

julia> gamma = 1.0
1.0

julia> W .= gamma .* J
4×4 SparseMatrixCSC{Float64, Int64} with 3 stored entries:
 1.0   ⋅    ⋅    ⋅
  ⋅   1.0   ⋅    ⋅
  ⋅    ⋅   1.0   ⋅
  ⋅    ⋅    ⋅    ⋅
```

Personally, I thought the answer would be 

```julia
julia> W .= gamma .* J
4×4 SparseMatrixCSC{Float64, Int64} with 3 stored entries:
 1.0   ⋅    ⋅    ⋅
  ⋅   1.0   ⋅    ⋅
  ⋅    ⋅   1.0   ⋅
  ⋅    ⋅    ⋅    0.0
```

but I was wrong: `.=` can change the sparsity pattern of the matrix `W`. There is a clear argument for this behavior: it stems from the fact that `setindex!` can change the sparsity pattern, and `.=` is just doing `setindex!` stuff, so therefore it can change the sparsity pattern (not just add, but also decrease structural non-zeros). And this is kind of documented:

https://docs.julialang.org/en/v1/stdlib/SparseArrays/

> Julia has support for sparse vectors and [sparse matrices](https://en.wikipedia.org/wiki/Sparse_matrix) in the SparseArrays stdlib module. Sparse arrays are arrays that contain enough zeros that storing them in a special data structure leads to savings in space and execution time, compared to dense arrays.

You can read that and say "that contain enough zeros", it didn't need them all, therefore it dropped them. QED, end of story.

But.

[...]

PR #201, opened 2022-07-30
Sparse matrix/vectors with fixed sparsity pattern.

SobhanMP:
This pull request adds
- an abstract type like AbstractMatrixCSC for sparse vectors (AbstractSparseVectorC)
- an abstract type for AbstractMatrixCSC whose colptr and rowval cannot be changed
- an abstract type for AbstractSparseVectorC whose colptr and rowval cannot be changed
- broadcast is the union of pattern in the presence of fixed elements
- through the magic of constprop `return _is_fixed(A, Bs...) ? fixed(r) : r` is typesafe.
- the tests don't check for every thing but check that the operations are correct and do not allocate.
- the `fixed` function can be used as a shortcut for creating sparse things with fixed sparsity.

once all green this should be ready to merge. this commit is _NOT_ breaking. 

apart from the bike-shed possible, i have a few concerns
- should the nonzeros field be non-resizable? right now, since rowvals is non-resizable, it is technically non-resizable (in correct arrays).
- i broadened the definitions in the higher order fns file. Mainly i changed the definition from `SparseMatrixCSC` to `AbstractSparseMatrixCSC`. is this correct?
- when this get merged, i think it would be nice to have an operator on non zero values, something nicer than
```julia
x = sprandn(100, 100, 0.1)
y = typeof(x)(sizeof(x)..., copy(getcolptr(x)), copy(rowvals(x)), nonzeros(x) .+ 2)
```
- should we share the colptr between the instances of FixedSparseCSC

i too feel that it's unfortunately long, :upside_down_face:. i thought it would be easier than this.

Ref: https://github.com/JuliaSparse/SparseArrays.jl/issues/190


SobhanMP:
after some hacking, i've come with some cases that needs discussing.
let $x$ be some CSC matrix with fixed pattern, what would  `x .+ 1` be? should it have no zeros or should it just add 1 to the nonzero entries? if `y` is another matrix like `x`, what should `x .+ y` look like? should it error if the structure of one does not fit in the other?

Wimmerer:
X .+ Y should be a union of the patterns I believe. X .+ 1 is tricky, and I'll defer to @ChrisRackauckas here since he asked for this initially.

ChrisRackauckas:
> X .+ Y should be a union of the patterns I believe.

Yes. 

> X .+ 1 is tricky

That's just dense: there's no other representation. So make it a dense sparse matrix, but of course, almost certainly the user would never do that.

SobhanMP:
for reasons i don't fully understand this seems to work. 
broadcast now takes the union of the structure of the `FixedSparseCSC`s

I propose adding
```
fill!(x::FixedSparseCSC, v) = fill!(nonzeros(x), v)
```
as `fill!(::FixedSparseCSC, v::Any)` only works if `iszero(v)` or `x` is already a dense sparse matrix(sparse matrix with all entries). In either case, this is equivalent and only changes the behavior from error to something weird in some cases.

SobhanMP:
@ChrisRackauckas could you test it a bit? the instruction to use a dev version of this package is in the README, and you can use `fixed` to convert to fixed patterns.

ViralBShah:
There's a lot going on in this PR - I'll need time to understand.

Wimmerer:
Is AbstractSparseVectorC any different from AbstractSparseVector? I'm not sure it makes sense to differentiate between storage orders for a 1-d array. 

I think storage order should be a trait anyway but that's a bigger issue.

SobhanMP:
i assumed one can implement a AbstractSparseVector with a dictionary or start from 0 instead of 1. Hence AbstractSparseVectorC is the subset that work like SparseVector, with one indexed nonzeros/nonzeroinds

SobhanMP:
case in point, look at #219

ViralBShah:
I am wondering if this code can live in a separate package?

SobhanMP:
i don't it can be a separate package w/o much redundancy (and piracy) in higherorderfns. having fixed arrays changes the meaning of broadcast/map/map!

src/abstractsparse.jl
```diff
@@ -16,6 +16,24 @@ Supertype for one-dimensional sparse arrays (or array-like types) with elements
 of type `Tv` and index type `Ti`. Alias for `AbstractSparseArray{Tv,Ti,1}`.
 """
 const AbstractSparseVector{Tv,Ti} = AbstractSparseArray{Tv,Ti,1}
+
+"""
+    AbstractSparseVectorC{Tv,Ti}
+
+Supertype for vectors stored using a compressed map.
+"""
+abstract type AbstractSparseVectorC{Tv,Ti} <: AbstractSparseVector{Tv,Ti} end
```
Wimmerer, line 25: Can this be `AbstractCompressedVector`? `C` is confusing to me every time I see it.

rayegun (commented):
A few comments on the actual code inline.

You mentioned that this can't be split out because that would be piracy. I don't think there would be any need for piracy (these changes involve a new type anyway).

I would argue this should look like:
1. Make `higherorderfns` use `AbstractSparseMatrixCSC` rather than `SparseMatrixCSC` in the `SparseOrStructuredMatrix` union. Most other instances of `SparseMatrixCSC` in signatures can be turned into `AbstractSparseMatrixCSC`.
2. Add trait to determine whether new indices can be inserted. `_is_fixed` is fine as is, `_can_insert` might be more clear.
3. `_noshapecheck_map` can be overloaded in a new package for `AbstractFixed...`, perhaps as simply as `_noshapecheck_map(f, A::AbstractFixed, Bs...) = fixed(invoke(_noshapecheck_map, AbstractSparseMatrixCSC, ..., A, Bs...)`. 

That avoids any piracy (as far as I can tell), and lets us avoid significant additions to the SparseArrays API before Julia 1.10 when we can really look at major changes (SparseArrays will still be an stdlib for 1.9 unfortunately).

src/sparsematrix.jl
```diff
@@ -44,6 +44,50 @@ function SparseMatrixCSC(m::Integer, n::Integer, colptr::Vector, rowval::Vector,
...
+fixed(x::AbstractSparseMatrixCSC) = FixedSparseCSC(x)
+_unsafe_unfix(s::FixedSparseCSC) = SparseMatrixCSC(size(s)..., inner(getcolptr(s)), inner(rowvals(s)), nonzeros(s))
+_unsafe_unfix(s::SparseMatrixCSC) = s
+const SorF = Union{<:SparseMatrixCSC, <:FixedSparseCSC}
+"""
+    SparseMatrixCSC(x::FixedSparseCSC)
+
+Get a non read-only view of x. get a copy for the sake of safety.
+"""
+SparseMatrixCSC(x::FixedSparseCSC) = SparseMatrixCSC(size(x, 1), size(x, 2), getcolptr(x).x, rowval(x).x, nonzeros(x))
```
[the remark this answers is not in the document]
SobhanMP: yes, it's an unsafe function.

src/abstractsparse.jl
SobhanMP: sure

src/sparsematrix.jl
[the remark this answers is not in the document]
SobhanMP: added the copy, `_unsafe_unfix` is the name of this now.

SobhanMP:
> Add trait to determine whether new indices can be inserted. _is_fixed is fine as is, _can_insert might be more clear.

sounds good, but we can't test it here. also i think that @ChrisRackauckas 's argument for putting this here is valid.

> _noshapecheck_map can be overloaded in a new package for AbstractFixed..., perhaps as simply as _noshapecheck_map(f, A::AbstractFixed, Bs...) = fixed(invoke(_noshapecheck_map, AbstractSparseMatrixCSC, ..., A, Bs...).

if this package respect the trait, there would be no need for this. but the problem is that how do we test that we respect this w/o having an implementation? Also this is not a major overhaul of the api, it's backward compatible, it's a fairly small addition (two concrete types, two abstract types, one trait) for a much wanted feature.

ViralBShah:
Yeah I am not worried about type piracy in an external package in this case. We need to make it easier to do these things in external packages and this does introduce a lot of new code. Things can move faster if they are not in here.

SobhanMP:
> this does introduce a lot of new code.
arguably, most of the code is just broadening types. there is a few small changes for `dropzero` and `fkeep` but this is mostly broadening types because the types are either too narrow or in the case SparseVector, AbstractSparseVector are assumed to have some features that are not always true. Regardless of whether we add a FixedSparseCSC, around half of the code here has to be added anyways.

> Things can move faster if they are not in here.
but can't be tested properly if they are not in here. Since this is an stdlib we can't use non stdlib pacakge to test it right? so the only solution would be to have a copy of the package in the test folder to test it which seems a bit ugly to say the least.

SobhanMP:
yes but every commit here has to run the test there.

Wimmerer:
We can have pseudo test-deps using by following https://github.com/JuliaIO/Tar.jl according to Stefan.

SobhanMP:
> I don't see why. 
if the implementation is in another package, the tests can't be here. sure we can what Tar.jl does but IMHO it's meh. you can't run those tests with the daily tests that people run and find bugs. At the end of the day since most of the implementation will stay here (including operations for fixed matrices), that package wont be easy to develop and the inertia of this package will be inherited by that package so it's kinda pointless. i doubt it's much worse than everything being here.

Only ReadOnly (31 lines) + FixedCSC (40 lines) + FixedVector (40 lines) can be moved to another package. ~300 lines of the patch will still have to be added (including AbstractCompressedVector (which is long overdue IMHO), AbstractFixedSparseVector, 
and AbstractFixedCSC), removing the last argument fkeep! which isn't used anywhere and broadening functions like
```julia
function _fillnonzero!(arr::SparseVector{Tv,Ti}, val) where {Tv,Ti}
    n = length(arr::SparseVector)
```

Wimmerer:
https://github.com/bkamins/ReadOnlyArrays.jl

SobhanMP:
the readonly type is a small implementation similar to it, except 
```julia
a[i]=a[i]
```
is not an error.

Wimmerer:
My thoughts are:

1. SparseCompressedVector should be removed from this PR and made into a new PR by itself. That will definitely be merged and can hopefully help fix some other issues.
2. I am uncomfortable adding two new array _types_ (one of which is dense!) only one release cycle (1.9) before we will release v1.0 of this stdlib (1.10) as a separate package. It isn't really about lines of code, rather, these types must now be supported essentially as-is in perpetuity. As our issues with 0-handling recently have shown, we're sort of stuck with any decisions made in this library.
4. This entire PR can be made into its own package in <600 loc at most (we don't really _use_ the broadcast machinery, otherwise that would be much more difficult). Duplicating a small number of functions is simple enough, and IMHO is the correct way to add new types. 

I think both @ViralBShah and I agree that (other than SparseCompressedVector) this PR should be extracted into its own package (including the higherorderfns.jl changes). 

I anticipate something like the `can_insert` or `can_modify_pattern` may end up in a future `AbstractSparse.jl` interface package. But that will come around much later with a lot of experimentation in the process.

SobhanMP:
> array types (one of which is dense!)

what?

SobhanMP:
also if the higherorderfn machinery goes outside, how am i supposed to overload that function, i have to overwrite it? that sounds really bad.

Wimmerer:
> what?

ReadOnly. I guess it's more accurate to call it a wrapper, but it's definitely useful outside of SparseArrays.jl.

> but this does not apply to experimental non-exported types, right?

Well, but in order to transition from experimental to solidly supported we need people to use it! Technically speaking you're right, but either no one will use it if it's not advertised, or people will, and it still becomes breaking-ish.

SobhanMP:
> but it's definitely useful outside of SparseArrays.jl.

i doubt, the fact that write does not error on `a[i] = copy(a[i])`, is very specific to this package.

> Well, but in order to transition from experimental to solidly supported we need people to use it! 

that can happen in 1.10.

SobhanMP:
@ViralBShah @Wimmerer i removed the abstract types. now there is a fairly small api

FixedSparseCSC minimal fixed implementations of CSC matrix
FixedSparseVector minimal AbstractCompressedVector implementation
_is_fixed(x::AbstractArray) which returns true if the indices are read only. I wrote the operations as shallow as i could meaning that colptr/rowvals do not get copied

there's also a few helper
fixed(x...) equivalent to sparse(x...) but returns a sparse array whose _is_fixed is true
move_fixed(x) like fixed but does not allocate
@if_move_fixed a1 a2 ... an x, returns fixed(x) if any of a1 to an is fixed o.w. returns x 

that's it.

src/higherorderfns.jl
```diff
@@ -476,6 +486,7 @@ end
 # (7) _broadcast_zeropres!/_broadcast_notzeropres! specialized for a single (input) sparse vector/matrix
 function _broadcast_zeropres!(f::Tf, C::SparseVecOrMat, A::SparseVecOrMat) where Tf
     isempty(C) && return _finishempty!(C)
+    isfixed = _is_fixed(C, A)
```
SobhanMP, line 489:
```suggestion
    isfixed = _is_fixed(A)
```

src/higherorderfns.jl
```diff
@@ -561,6 +572,7 @@ end
 # (8) _broadcast_zeropres!/_broadcast_notzeropres! specialized for a pair of (input) sparse vectors/matrices
 function _broadcast_zeropres!(f::Tf, C::SparseVecOrMat, A::SparseVecOrMat, B::SparseVecOrMat) where Tf
     isempty(C) && return _finishempty!(C)
+    isfixed = _is_fixed(C, A, B)
```
SobhanMP, line 575:
```suggestion
    isfixed = _is_fixed(A, B)
```

src/higherorderfns.jl
```diff
@@ -883,6 +895,7 @@ end
 # (9) _broadcast_zeropres!/_broadcast_notzeropres! for more than two (input) sparse vectors/matrices
 function _broadcast_zeropres!(f::Tf, C::SparseVecOrMat, As::Vararg{SparseVecOrMat,N}) where {Tf,N}
     isempty(C) && return _finishempty!(C)
+    isfixed = _is_fixed(C, As...)
```
SobhanMP, line 898:
```suggestion
    isfixed = _is_fixed(As...)
```

src/readonly.jl
```diff
@@ -0,0 +1,31 @@
...
+Wrapper around an abstract vector, blocks changing elements (but not `setindex!`).
+This is done by only throwing errors when the modifying operations changes the structure.
+For instance resizing a vector to its original length will not throw an error as it is basically a no op.
+"""
+struct ReadOnly{T,V<:AbstractVector{T}} <: AbstractVector{T}
+    parent::V
+end
+Base.getproperty(::ReadOnly, ::Symbol) = error("Use parent instead.")
+@inline Base.parent(x::ReadOnly) = getfield(x, :parent)
+Base.@propagate_inbounds getindex(x::ReadOnly, v...) = getindex(parent(x), v...)
```
SobhanMP, line 15:
```suggestion
Base.@propagate_inbounds @inline getindex(x::ReadOnly, v...) = getindex(parent(x), v...)
```

src/readonly.jl
```diff
@@ -0,0 +1,31 @@
...
+for i in [:length, :lastindex]
+    Base.@propagate_inbounds @eval Base.$i(x::ReadOnly) = Base.$i(parent(x))
+end
+Base.@propagate_inbounds Base.setindex!(x::ReadOnly, v, ind...) = if v == getindex(parent(x), ind...)
```
SobhanMP, line 19:
```suggestion
Base.@propagate_inbounds @inline Base.setindex!(x::ReadOnly, v, ind...) = if v == getindex(parent(x), ind...)
```

src/readonly.jl
```diff
@@ -0,0 +1,31 @@
...
+For instance resizing a vector to its original length will not throw an error as it is basically a no op.
+"""
+struct ReadOnly{T,V<:AbstractVector{T}} <: AbstractVector{T}
+    parent::V
+end
+Base.getproperty(::ReadOnly, ::Symbol) = error("Use parent instead.")
+@inline Base.parent(x::ReadOnly) = getfield(x, :parent)
+Base.@propagate_inbounds getindex(x::ReadOnly, v...) = getindex(parent(x), v...)
+for i in [:length, :lastindex]
+    Base.@propagate_inbounds @eval Base.$i(x::ReadOnly) = Base.$i(parent(x))
```
SobhanMP, line 17:
```suggestion
    @eval Base.@propagate_inbounds @inline Base.$i(x::ReadOnly) = Base.$i(parent(x))
```

src/sparsematrix.jl
```diff
@@ -44,6 +44,58 @@ function SparseMatrixCSC(m::Integer, n::Integer, colptr::Vector, rowval::Vector,
...
+end
+@inline _is_fixed(::FixedSparseCSC) = true
+FixedSparseCSC(m::Integer, n::Integer,
+    colptr::ReadOnly{Ti, Vector{Ti}},
+    rowval::ReadOnly{Ti, Vector{Ti}},
+    nzval::Vector{Tv}) where {Tv,Ti<:Integer} =
+    FixedSparseCSC{Tv,Ti}(m, n, colptr, rowval, nzval)
+FixedSparseCSC{Tv,Ti}(m::Integer, n::Integer, colptr::Vector{Ti}, rowval::Vector{Ti}, nzval::Vector{Tv}) where {Tv,Ti} =
+    FixedSparseCSC{Tv,Ti}(m, n, ReadOnly(colptr), ReadOnly(rowval), nzval)
+FixedSparseCSC(m::Integer, n::Integer, colptr::Vector{Ti}, rowval::Vector{Ti}, nzval::Vector{Tv}) where {Tv,Ti} =
```
SobhanMP, line 75: is this method dangerous? it's pretty handy and since SparseMatrixCSC edits the vectors, i guess this is safe.

src/sparsematrix.jl
```diff
@@ -1632,6 +1697,22 @@ function fkeep!(A::AbstractSparseMatrixCSC, f, trim::Bool = true)
...
+            # If this element should be kept, rewrite in new position
+            if f(Ai, Aj, Ax)
+                nonzeros(A)[k] = zero(eltype(A))
+            end
+        end
+    end
+    return A
+end
+
+fkeep!(A::AbstractSparseMatrixCSC, f) = _is_fixed(A) ? _fkeep!_fixed(A, f) : _fkeep!(A, f)
```
SobhanMP, line 1714: after this patch is done, i'll fix fkeep! once and for all. the function is always first.

src/sparsematrix.jl
```diff
@@ -1632,6 +1697,22 @@ function fkeep!(A::AbstractSparseMatrixCSC, f, trim::Bool = true)
     return A
 end
 
+function _fkeep!_fixed(A::AbstractSparseMatrixCSC, f) where S
```
SobhanMP, line 1700:
```suggestion
function _fkeep!_fixed(A::AbstractSparseMatrixCSC, f::F) where F
```

src/sparsematrix.jl
```diff
@@ -1597,7 +1662,7 @@ julia> SparseArrays.fkeep!(A, (i, j, v) -> isodd(v))
  ⋅  ⋅  ⋅  ⋅
 ```
 """
-function fkeep!(A::AbstractSparseMatrixCSC, f, trim::Bool = true)
+function _fkeep!(A::AbstractSparseMatrixCSC, f)
```
SobhanMP, line 1665:
```suggestion
function _fkeep!(A::AbstractSparseMatrixCSC, f::F) where F
```

src/sparsematrix.jl
SobhanMP, line 1714:
```suggestion
fkeep!(A::AbstractSparseMatrixCSC, f::F) where F= _is_fixed(A) ? _fkeep!_fixed(A, f) : _fkeep!(A, f)
```

src/sparsevector.jl
```diff
@@ -339,7 +370,8 @@ end
 
 ### Element access
 
-@RCI function setindex!(x::SparseVector{Tv,Ti}, v::Tv, i::Ti) where {Tv,Ti<:Integer}
+
```
SobhanMP, line 373:
```suggestion
```

src/sparsevector.jl
```diff
@@ -357,7 +389,7 @@ end
     x
```
SobhanMP, line 389:
```suggestion
    return x
```

Wimmerer:
Is this ready @SobhanMP ? Everything including concrete types can be merged once green. Please be sure to document that this is not public right now.

PR #202, opened 2022-07-31
broaden umfpack to use AbstractSparseMatrixCSC instead of SparseMatrixCSC

SobhanMP:


Wimmerer (approved):
LGTM, I didn't see any direct field accesses which might be an issue.

PR #203, opened 2022-08-03
Adjustment to introduction of `AdjointQ` in LinAlg

dkarrasch:
This is the companion PR to https://github.com/JuliaLang/julia/pull/46196. It will pass tests only once that PR is merged, if it gets accepted.


dkarrasch:
This passes tests on current master (obviously) and on my PR branch (locally). So I'll merge this.

PR #204, opened 2022-08-03
Fix deprecated Threads.nthreads()

IanButterworth:
`Threads.nthreads()` is going to be deprecated in https://github.com/JuliaLang/julia/pull/45447

This needs to be fixed before https://github.com/JuliaLang/julia/pull/45447 merges because Base tests run with `--depwarn=error` https://github.com/JuliaLang/julia/blob/df846436bb91b291be988f5e18e256d514b97347/test/testenv.jl#L19

### DIFF
--- a/src/solvers/cholmod.jl
+++ b/src/solvers/cholmod.jl
@@ -221,7 +221,11 @@ function __init__()
         ### Initiate CHOLMOD
         ### common controls the type of factorization and keeps pointers
         ### to temporary memory. We need to manage a copy for each thread.
-        nt = Threads.nthreads()
+        nt = @static if isdefined(Threads, :maxthreadid)
+            Threads.maxthreadid()
+        else
+            Threads.nthreads()
+        end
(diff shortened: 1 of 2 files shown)


DilumAluthge (dismissed):

IanButterworth:
I think this needs signoff from @JeffBezanson / @vtjnash that `Threads.maxthreadid()` is the name that's going to be used, given https://github.com/JuliaLang/julia/pull/45447 can't be merged first

vtjnash (changes requested):
FWIW, this would be very wrong, since there is no replacement for `nthreads` now

Wimmerer:
That will require a pretty significant rewrite of the CHOLMOD wrapper. We'll have to carry workspaces and settings around with each Factor object now I believe.

DilumAluthge:
Could you use `Base.task_local_storage`?

IanButterworth:
I'll leave the necessary changes up to @Wimmerer

PR #206, opened 2022-08-04
Remove Thread-local storage from CHOLMOD and SPQR

Wimmerer:
See https://github.com/JuliaSparse/SparseArrays.jl/pull/204

This PR takes the path of least resistance and just uses `task_local_storage`. The part that most needs review is the addition of a call to `cholmod_l_finish`. This is attached to the `cholmod_common` object which is lazily created in each task (which uses CHOLMOD). 

I *think* this is correct, but I'm not 100% certain on the lifetime of `task_local_storage`.

### DIFF
--- a/src/solvers/cholmod.jl
+++ b/src/solvers/cholmod.jl
@@ -138,7 +138,7 @@ macro cholmod_param(kwarg, code)
     value = kwarg.args[2]
 
     common_param = # Read `common.param`
-        Expr(:., :(COMMONS[Threads.threadid()][]), QuoteNode(param))
+        Expr(:., :(task_local_storage(:cholmod_common)[]), QuoteNode(param))
 
     return quote
         default_value = $common_param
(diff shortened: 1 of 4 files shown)


Wimmerer:
@IanButterworth does this fix the issue?

src/solvers/cholmod.jl
```diff
@@ -151,7 +151,18 @@ macro cholmod_param(kwarg, code)
...
+function getcommon()
+    if !haskey(task_local_storage(), :cholmod_common) # havent yet started cholmod on this task
+        common = finalizer(cholmod_l_finish, Ref(cholmod_common()))
+        result = cholmod_l_start(common)
+        @assert result == TRUE "failed to run `cholmod_l_start`!"
+        common[].print = 0  # no printing from CHOLMOD by default
+        common[].error_handler = @cfunction(error_handler, Cvoid, (Cint, Cstring, Cint, Cstring))
+        task_local_storage(:cholmod_common, common)
+    else
+        return task_local_storage(:cholmod_common)
```
vchuravy, line 163: TLS is untyped so you might want to add a `::T` here. Also I think it supports the `get!` interface

src/solvers/cholmod.jl
```diff
@@ -151,7 +151,18 @@ macro cholmod_param(kwarg, code)
...
+    common = finalizer(cholmod_l_finish, Ref(cholmod_common()))
+    result = cholmod_l_start(common)
+    @assert result == TRUE "failed to run `cholmod_l_start`!"
+    common[].print = 0  # no printing from CHOLMOD by default
+    common[].error_handler = @cfunction(error_handler, Cvoid, (Cint, Cstring, Cint, Cstring))
+    return common
+end
+
+function getcommon()
+    return get!(task_local_storage(), :cholmod_common, newcommon())::Ref{cholmod_common}
```
fredrikekre, line 164: Should this not be
```
get!(newcommon, task_local_storage(), :cholmod_common)::Ref{cholmod_common}
```
? As it is now a new object is created for every call.

Wimmerer: The docs say it's `collection, key, default` right?

IanButterworth: There's another method where you can give a function/type as a first arg. Note that Fredrick's suggestion gives the type but doesn't evaluate it

PR #209, opened 2022-08-07
use _isnotzero(v)

### LINKED ISSUE #15: Problem with sparse arrays: missing values
I am not sure where to report it, but I think there is a problem with handling `missing` values for sparse arrays:
```
julia> ms = sparse([1, 2], [1, 2], [10, missing])
2×2 SparseMatrixCSC{Union{Missing, Int64}, Int64} with 2 stored entries:
 10  ⋅
  ⋅   missing

julia> md = Matrix(ms)
2×2 Matrix{Union{Missing, Int64}}:
 10  0
  0   missing

julia> sparse(md)
ERROR: TypeError: non-boolean (Missing) used in boolean context

julia> ms[1, 2] = missing
ERROR: TypeError: non-boolean (Missing) used in boolean context

julia> ms[1, 1] = missing
missing

julia> ms
2×2 SparseMatrixCSC{Union{Missing, Int64}, Int64} with 2 stored entries:
  missing  ⋅
 ⋅          missing

julia> x = spzeros(3)
3-element SparseVector{Float64, Int64} with 0 stored entries

julia> y = similar(x, Union{Int, Missing})
3-element SparseVector{Union{Missing, Int64}, Int64} with 0 stored entries

julia> y[1] = missing
ERROR: TypeError: non-boolean (Missing) used in boolean context

julia> y[1] = 10
10

julia> y[1] = missing
missing
```

CC @nalimilan

SobhanMP:
closes #15

replaces a bunch of `!iszero` with `_isnotzero` in order to handle `Missing`.

### DIFF
--- a/test/issues.jl
+++ b/test/issues.jl
@@ -9,6 +9,28 @@ using InteractiveUtils: @which
 include("forbidproperties.jl")
 include("simplesmatrix.jl")
 
+
+@testset "Issue #15" begin
+    s = sparse([1, 2], [1, 2], [10, missing])
+    d = Matrix(s)
+    
+    s2 = sparse(d)
+
+    @test s2[1, 1] == 10
+    @test s2[2, 1] == 0
+    @test s2[1, 2] == 0
+    @test s2[2, 2] === missing
+    @test typeof(s2) == typeof(s)
+
+    x = spzeros(3)
+    y = similar(x, Union{Int, Missing})
+    y[1] = missing
+    y[2] = 10
+    @test y[1] === missing
+    @test y[2] == 10
+    @test y[3] == 0
+end
+
 @testset "Issue #33169" begin
     m21 = sparse([1, 2], [2, 2], SimpleSMatrix{2,1}.([rand(2, 1), rand(2, 1)]), 2, 2)
     m12 = sparse([1, 2], [2, 2], SimpleSMatrix{1,2}.([rand(1, 2), rand(1, 2)]), 2, 2)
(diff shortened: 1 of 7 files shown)


SobhanMP:
the only question is whether findprev/findnext should specialize based on f. i would say that it makes sense that it does because it calls f.

ViralBShah:
Where is `_isnotzero` defined?

SobhanMP:
`_isnotzero(x)` is defined in in SparseArrays.jl:39, it's the function @dkarrasch added to fix the handle missings in #177

ViralBShah:
Ah I see it now - turns out I had not done a `git pull`.

SobhanMP:
should i add tests to it?

ViralBShah:
Yes, let's get a few tests and merge.

cc @bkamins

ViralBShah:
Is there any noticeable performance impact since some of these can be deep in an inner loop?

SobhanMP:
it has an inline overload for ::number so i should be zero

test/issues.jl
```diff
@@ -9,6 +9,18 @@ using InteractiveUtils: @which
 include("forbidproperties.jl")
 include("simplesmatrix.jl")
 
+
+@testset "issue #15" begin
+    s = sparse([1, 2], [1, 2], [10, missing])
+    d = Matrix(s)
+    @test (sparse(d); true)
+    s2 = sparse(d)
```
bkamins, line 17: I would rather test type of `s2`.

test/issues.jl
```diff
@@ -9,6 +9,18 @@ using InteractiveUtils: @which
...
+    @test s2[1, 1] == 10
+    @test s2[2, 1] == 0
+    @test s2[1, 2] == 0
+    @test s2[2, 2] === missing
```
bkamins, line 21: maybe also add `setindex!` tests that were failing in https://github.com/JuliaSparse/SparseArrays.jl/issues/15?

SobhanMP:
the remaining `iszero`s seem to be used on the indexes (which are always integer) so should not cause any problem.

ViralBShah (approved):

PR #215, opened 2022-08-10
Enable julia stable in CI

ViralBShah:
Let's try to make sure we can keep this package going on Julia stable, and maybe even Julia 1.6.

@dkarrasch I believe the `Base.extremamap` is Julia 1.8 only (I believe it should have been added through your recent PR).  If possible, it would be nice to make the package in its current shape work on Julia 1.7, but if not, we can make 1.8 the minimum required going forward, but be more conservative going forward about using features in new Julia versions.

My purpose here is to just be a bit more deliberate.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,7 +22,7 @@ jobs:
       fail-fast: false
       matrix:
         version:
-          - '1'
+          - '1.8-'
           - 'nightly'
         os:
           - ubuntu-latest
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,7 +22,7 @@ jobs:
       fail-fast: false
       matrix:
         version:
-          # - '1.6'
+          - '1'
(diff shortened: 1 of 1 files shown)


SobhanMP:
is there any reason? it's not like 1.6 is going to get a bumped version of this package.

ViralBShah:
The main reason is not to backport or ship new SparseArrays with old releases, but to make sure that we don't break backwards compatibility accidentally. It's definitely a nice to have, among many of the other things that need doing here, rather than must have.

Wimmerer:
It's important for people who depend on SparseArrays. If a new feature is added that doesn't support 1.6 it can become annoying for packages that *so* support 1.6

SobhanMP:
i didn't know 1.6 still got SparseArrays updates.

Wimmerer:
It doesn't. I was unclear.

ViralBShah:
Right, only crucial bugfixes would get backported to lts and stable.

PR #216, opened 2022-08-10
Add documentation for similar of sparse matrices

### LINKED ISSUE #14: Strange behavior of similar for sparse arrays
We have the following:

```
julia> x = spzeros(1)
1-element SparseVector{Float64, Int64} with 0 stored entries

julia> similar(x)
1-element SparseVector{Float64, Int64} with 0 stored entries

julia> x[1] = 1
1

julia> x
1-element SparseVector{Float64, Int64} with 1 stored entry:
  [1]  =  1.0

julia> similar(x)
1-element SparseVector{Float64, Int64} with 1 stored entry:
  [1]  =  6.36599e-314
```

Which seems strange to me as `similar` should produce an uninitialized array, while in the second example it is initialized as I would consider sparse array uninitialized if it had 0 stored entries.

CC @nalimilan

ViralBShah:
`similar` does slightly different things for sparse matrices. Document that. Fix #14 

I'm not sure if this is the right way to add these docs. Would appreciate a review from someone more knowledgable.

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -212,6 +212,7 @@ SparseArrays.SparseVector
 SparseArrays.SparseMatrixCSC
 SparseArrays.sparse
 SparseArrays.sparsevec
+SparseArrays.similar
 SparseArrays.issparse
 SparseArrays.nnz
 SparseArrays.findnz
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -212,7 +212,7 @@ SparseArrays.SparseVector
 SparseArrays.SparseMatrixCSC
 SparseArrays.sparse
 SparseArrays.sparsevec
-SparseArrays.similar
+Base.similar
 SparseArrays.issparse
 SparseArrays.nnz
 SparseArrays.findnz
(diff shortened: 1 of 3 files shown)


Wimmerer (approved):
Lgtm. I didn't actually realize similar behaved like this for CSC. It feels like it should be a new function, or a keyword `keep_pattern`, but the docs are good.

ViralBShah:
We do want it in the html docs.

ViralBShah:
What am I doing wrong?

```
┌ Warning: no docs found for 'SparseArrays.similar' in `@docs` block in src/index.md:207-238
```

ViralBShah:
@Wimmerer Any ideas what is wrong here. It doesn't show up in the API reference.

docs/src/index.md
```diff
@@ -212,6 +212,7 @@ SparseArrays.SparseVector
 SparseArrays.SparseMatrixCSC
 SparseArrays.sparse
 SparseArrays.sparsevec
+SparseArrays.similar
```
SobhanMP, line 215:
```suggestion
Base.similar
```

SobhanMP: image

PR #229, opened 2022-08-21
Fix doc build in stdlibbump

Wimmerer:
I'm just going to bump after this is green to see if it fixes the `similar` issue on julialang/julia.


Wimmerer:
Oops, I didn't want to drag all that git history with me.

PR #243, opened 2022-09-01
add faster sum/any/all/iszero

SobhanMP:
closes https://github.com/JuliaSparse/SparseArrays.jl/issues/237
i'll write tests tomorrow

would it make sense that the functions for transpose/adjoint be added to LinearAlgebra.jl?

can nonzeros be larger than nnz? there are code where it would be valid and code that it would not be.

### DIFF
--- a/src/abstractsparse.jl
+++ b/src/abstractsparse.jl
@@ -193,3 +193,5 @@ macro if_move_fixed(a...)
     h, v = esc.(a[1:end - 1]), esc(a[end])
     :(_is_fixed($(h...)) ? move_fixed($v) : $v)
 end
+
+const SparseVecOrMat = Union{AbstractCompressedVector,AbstractSparseMatrixCSC}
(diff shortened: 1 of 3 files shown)


dkarrasch:
But LinearAlgebra doesn't know about `SparseXYZ`. I think these functions should live here. I remember we have made an effort to implement efficient reduction functions. Don't they apply here? Maybe there is some dispatch going in the wrong direction? Looking at sparsevector.jl lines 1425ff, there is sparsity-aware code.

dkarrasch:
Wait. This doesn't apply to `SparseVector`s, but to sparse matrices? But then again, we have code in sparsematrix.jl lines 2132ff, and seemingly a specialization for `+`.

dkarrasch (changes requested):
I believe this is all actually handled well, see my comment in the original issue. What is missing is a `count` specialization for `Bool` sparse arrays.

SobhanMP:
> But LinearAlgebra doesn't know about SparseXYZ. 

sure but what ever the parent is, shouldn't Transpose just forward it to the parent? like always

dkarrasch:
I see what you mean. Like

```julia
iszero(x::AdjOrTrans) = iszero(parent(x))
```

right?

SobhanMP:
yeah

SobhanMP:
closed in favor #244

PR #245, opened 2022-09-02
fix findmin/findmax (#93)

dkarrasch:
Co-authored-by: Steven G. Johnson <stevenj@mit.edu>


dkarrasch:
@KristofferC @DilumAluthge 

I hope this is the right thing to do. If this one is correct, then what's next to make Julia Base aware of this new commit and make it included in the v1.8.1 release?

PR #248, opened 2022-09-06
add == for vectors

### LINKED ISSUE #246: Missing method for == on sparse vectors
The following tries to iterate all elements of the sparse vector and takes forever:
```
julia> using SparseArrays

julia> a = SparseVector(div(typemax(Int), 2), [1], [1])
4611686018427387903-element SparseVector{Int64, Int64} with 1 stored entry:
  [1]  =  1

julia> a == a
```

SobhanMP:
merge on green.
Closes #246

### DIFF
--- a/test/sparsematrix_ops.jl
+++ b/test/sparsematrix_ops.jl
@@ -4,7 +4,7 @@ module SparseTests
 
 using Test
 using SparseArrays
-using SparseArrays: getcolptr, nonzeroinds, _show_with_braille_patterns, _isnotzero
+using SparseArrays: getcolptr, nonzeros, nonzeroinds, _show_with_braille_patterns, _isnotzero
 using LinearAlgebra
 using Printf: @printf # for debug
 using Random
@@ -535,13 +535,27 @@ end
     for t in [Int, UInt8, Float64]
         a = Counting.(sprand(t, 100, 0.5))
         b = Counting.(sprand(t, 100, 0.5))
+
+        c = if nnz(a) != 0
+            c = copy(a)
+            nonzeros(c)[1] = 0
+            c
+        else
+            c = copy(a)
+            push!(nonzeros(c), zero(t))
+            push!(nonzerosinds(c), 1)
+            c
+        end
+        d = dropzeros(c)
+
         for m in [identity, transpose, adjoint]
-            ma = m(a)
-            mb = m(b)
+            ma, mb, mc, md = m.([a, b, c, d])
 
             resetcounter()
             ma == mb
             @test getcounter() <= nnz(a) + nnz(b)
+
+            @test (mc == md) == (Array(mc) == Array(md))
         end
     end
 end
--- a/test/sparsematrix_ops.jl
+++ b/test/sparsematrix_ops.jl
@@ -530,4 +530,20 @@ Base.transpose(x::Counting) = Counting(transpose(x.elt))
     end
 end
 
+
+@testset "Issue #246" begin
+    for t in [Int, UInt8, Float64]
+        a = Counting.(sprand(t, 100, 0.5))
+        b = Counting.(sprand(t, 100, 0.5))
+        for m in [identity, transpose, adjoint]
+            ma = m(a)
+            mb = m(b)
+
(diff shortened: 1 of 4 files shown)


dkarrasch:
This reminds of our broadcast code. Couldn't we replace the code by `all(A .== B)` after the size check?

SobhanMP:
```julia
julia> using SparseArrays

julia> a = sprandn(100000, 0.9);

julia> b = sprandn(100000, 0.9);

julia> @time a .== b;
  0.002611 seconds (7 allocations: 879.344 KiB)
```
i agree that they are similar but that would make it allocating. i don't think we want that.

dkarrasch:
What I had in mind is something lazy like `all(Broadcast.broadcasted(==, Y, Y))`, which is allocation-free, but doesn't seem to use any sparse technology. It seems to just iterate via slow `getindex`.

dkarrasch:
Do we have tests for this? Including things like equal vectors where one has a stored zero?

SobhanMP:
i added a test with explicit zero

SobhanMP:
if you have something specific in mind, do share

test/sparsematrix_ops.jl
```diff
@@ -4,7 +4,7 @@ module SparseTests
 
 using Test
 using SparseArrays
-using SparseArrays: getcolptr, nonzeroinds, _show_with_braille_patterns, _isnotzero
+using SparseArrays: getcolptr, nonzeros, nonzeroinds, _show_with_braille_patterns, _isnotzero
```
dkarrasch, line 7: Seems like `nonzeros` is already exported.

```suggestion
using SparseArrays: getcolptr, nonzeroinds, _show_with_braille_patterns, _isnotzero
```

ViralBShah:
Do we really need all the `@inbounds`? Does it have a noticeable impact here?

SobhanMP:
@ViralBShah on really large examples, yeah

```julia
julia> @benchmark eq1($A, $B)
BenchmarkTools.Trial: 29 samples with 1 evaluation.
 Range (min … max):  168.045 ms … 186.107 ms  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     173.698 ms               ┊ GC (median):    0.00%
 Time  (mean ± σ):   174.602 ms ±   4.069 ms  ┊ GC (mean ± σ):  0.00% ± 0.00%

  ▁     ▁▁▁▁█  ██ ▁ █▁▁ ▁▁▁█  ▁▁   ▁ ▁   ▁  ▁      ▁          ▁  
  █▁▁▁▁▁█████▁▁██▁█▁███▁████▁▁██▁▁▁█▁█▁▁▁█▁▁█▁▁▁▁▁▁█▁▁▁▁▁▁▁▁▁▁█ ▁
  168 ms           Histogram: frequency by time          186 ms <

 Memory estimate: 0 bytes, allocs estimate: 0.

julia> @benchmark eq2($A, $B)
BenchmarkTools.Trial: 23 samples with 1 evaluation.
 Range (min … max):  189.635 ms … 370.411 ms  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     200.284 ms               ┊ GC (median):    0.00%
 Time  (mean ± σ):   221.056 ms ±  49.187 ms  ┊ GC (mean ± σ):  0.00% ± 0.00%

   █▅                                                            
  ▅███▁██▅▁▁▁▁▅▁▁▁▁▅▅▅▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▅▁▅ ▁
  190 ms           Histogram: frequency by time          370 ms <
```

SobhanMP:
i could replace the last two loops with 
```julia
return all(iszero, view(nonzeros(A), i:nnz(A))) && all(iszero, view(nonzeros(B), j:nnz(B)))
```

ViralBShah (approved):

SobhanMP:
@ViralBShah so do i replace or no?

ViralBShah:
I think either way is fine.

Issue #251, opened 2022-09-07
Change of behavior/regression between julia 1.7 and 1.8


lucasondel:
Hello, 

I've observed some inconsistencies between julia 1.7 and 1.8 with the SparseArrays module:
1. Concatenation of sparse vectors:
```julia
julia> VERSION
v"1.7.2"
julia> using SparseArrays
julia> x = sparsevec([1], [3])
1-element SparseVector{Int64, Int64} with 1 stored entry:
  [1]  =  3
julia> vcat(x, zero(eltype(x)))
2-element SparseVector{Int64, Int64} with 1 stored entry:  # <- a sparse vector is returned as expected
  [1]  =  3
```

```julia
julia> VERSION
v"1.8.0"
julia> using SparseArrays
julia> x = sparsevec([1], [3])
1-element SparseVector{Int64, Int64} with 1 stored entry:
  [1]  =  3
julia> vcat(x, zero(eltype(x)))
2×1 SparseMatrixCSC{Int64, Int64} with 1 stored entry: # <- a sparse matrix CSC is returned
 3
 ⋅
```
2. Failure to use `sum` with types not supporting integer multiplication:
```julia
julia> VERSION
v"1.7.2"
julia> using SparseArrays
julia> using Semirings
julia> K = LogSemiring{Float64} # This type cannot be multiplied with an integer
LogSemiring{Float64}
julia> sum(sparse([2], [3], K[one(K)], 10, 10))
0.0
```
```julia
julia> VERSION
v"1.8.0"
julia> using SparseArrays
julia> using Semirings
julia> K = LogSemiring{Float64} # This type cannot be multiplied with an integer
LogSemiring{Float64}
julia> sum(sparse([2], [3], K[one(K)], 10, 10))
ERROR: promotion of types LogSemiring{Float64} and Int64 failed to change any arguments
Stacktrace:
  [1] error(::String, ::String, ::String)
    @ Base ./error.jl:44
  [2] sametype_error(input::Tuple{LogSemiring{Float64}, Int64})
    @ Base ./promotion.jl:383
  [3] not_sametype(x::Tuple{LogSemiring{Float64}, Int64}, y::Tuple{LogSemiring{Float64}, Int64})
    @ Base ./promotion.jl:377
  [4] promote
    @ ./promotion.jl:360 [inlined]
  [5] *(x::LogSemiring{Float64}, y::Int64)
    @ Base ./promotion.jl:389
  [6] _mapreducezeros
[...]

Issue #249, opened 2022-09-06
Create a `release-1.8` branch and backport fixes to it


KristofferC:
Stdlibs that live in their own repo generally needs to do a similar backport process as in the Julia repo. Roughly this is done by:

- Have a `release-1.x` branch.
- Before a release, create a `backports-release-1.x` branch.
- Backport fixes on master that are suitable for backporting to `backports-release-1.x`
- When CI looks good on that branch, merge it into `release-1.x`.
- Bump the `stdlib/SparseArrays.version` on the backport branch in the Julia repo.

Wimmerer:
To run CI on that branch I have to PR from it correct? PR from `backports-release-1.8` to `release-1.8`. And then we can also collect the solution to https://github.com/JuliaSparse/SparseArrays.jl/issues/251 as well hopefully.

DilumAluthge:
>     * [ ]  Bump the `stdlib/SparseArrays.version` on the backport branch in the Julia repo.

Note: currently, this step needs to be done manually. However, once https://github.com/JuliaLang/BumpStdlibs.jl/issues/76 has been done, we'll be able to use the BumpStdlibs bot to do this step automatically.

PR #253, opened 2022-09-08
Fix `vcat` of sparse vectors with numbers

dkarrasch:
Fixes the `vcat` aspect of #251.


ViralBShah:
I'm not sure how backports will work. @DilumAluthge @KristofferC ?

DilumAluthge:
The high-level process is described in https://github.com/JuliaSparse/SparseArrays.jl/issues/249

dkarrasch:
Ok, shall we merge this and then cherry-pick the commit and PR into `backports-release-1.8`? From there, we should PR into `release-1.8` and declare that in Julia, right?

ViralBShah:
Sounds about right. Please merge when you are ready.

dkarrasch:
Let's see what the backport github action does once I merge...

PR #255, opened 2022-09-13
Run Julia v1.8 on the `release-1.8` branch

dkarrasch:
I think this is only consistent with what this branch is good for: make sure Julia v1.8 is working.


Wimmerer (approved):

dkarrasch:
This is no longer needed. It is going to be pulled from `backports-release-1.8`.

PR #256, opened 2022-09-13
On the backports-v1.8 branch, run CI on v1.8

dkarrasch:
Maybe replaces #255?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,8 +22,8 @@ jobs:
       fail-fast: false
       matrix:
         version:
-          # - '1.6'
-          - 'nightly'
+          - '1.8'
+          # - 'nightly'
         os:
           - ubuntu-latest
           - macOS-latest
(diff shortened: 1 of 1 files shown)


Wimmerer:
Should we open a PR to merge backports-release-1.8 => release-1.8 with this in it? That should trigger CI right?

dkarrasch:
> backports-release-1.8 => release-1.8

Yes. We could do that right away, or include the vcat fix before that. But I guess the order doesn't matter much as long as we don't change the commit SHA in Julia `release-v1.8`, right?

dkarrasch:
I'm merging this. It can't be wrong to do that IIUC.

Wimmerer:
> change the commit SHA in Julia release-v1.8, right?

Yeah I say we merge into backports and just open a PR for CI purposes. Once we've done that and it passes we can follow up with Kristoffer or Dilum to be absolutely sure we've got it right.

PR #259, opened 2022-09-14
Remove deepcopy from UMFPACK, add kwarg to copy

Wimmerer:
@JeffBezanson is this more acceptable? The semantics of the previous version didn't really match `deepcopy` anyway (only certain fields were copied). 

Can `copy` signature be modified without affecting Base specialization changes?

### DIFF
--- a/src/solvers/umfpack.jl
+++ b/src/solvers/umfpack.jl
@@ -265,7 +265,7 @@ This must be done if multiple threads may call factorization or refactorization 
 on the copy and original simultaneously.
 """
 # Not using simlar helps if the actual needed size has changed as it would need to be resized again
-Base.copy(F::UmfpackLU{Tv, Ti}, ws=UmfpackWS(F); copynumeric = false, copysymbolic) where {Tv, Ti} =
+Base.copy(F::UmfpackLU{Tv, Ti}, ws=UmfpackWS(F); copynumeric = false, copysymbolic = false) where {Tv, Ti} =
     UmfpackLU(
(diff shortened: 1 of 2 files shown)


Wimmerer:
@SobhanMP is this an acceptable interface for copying? I believe you were the primary person looking for it (for multicore stuff?). I've made a PR to SuiteSparse upstream to add copying, it should hopefully be merged before feature freeze.

SobhanMP:
looks fine to me.

SobhanMP:
might want to also deal with 

https://github.com/JuliaSparse/SparseArrays.jl/blob/e15a37e5ec722aabb66229d82890ccd0e32182a7/src/solvers/umfpack.jl#L282

Wimmerer:
I don't think this needs a backport @KristofferC SuiteSparse isn't in this repo in 1.8

JeffBezanson:
Thanks, this is better. Is `safecopy` so expensive that it can't be the default? Shame to need that option.

SobhanMP:
That's almost all of the computation when calculating the LU factorization. Since we rename duplicate to copy, copy should be as cheap as possible for multi-threaded applications.

Wimmerer:
There may be an improved default (implemented as a memcpy) once SuiteSparse upstream can be updated (I have a PR in, it just needs some tests and rebasing). That avoids the recomputation of the LU, which is what has to be done now for safecopy. I'm hoping to enable this for all the SuiteSparse solvers, but I don't know Tim's SuiteSparse release cycle timeline.

TLDR: Default can be made safe and fast ~Julia 1.10

PR #260, opened 2022-09-15
Fix direction of circshift

fredrikekre:
See https://github.com/JuliaLang/julia/pull/46759.

### DIFF
--- a/src/SparseArrays.jl
+++ b/src/SparseArrays.jl
@@ -11,6 +11,9 @@ using Base.Sort: Forward
 using LinearAlgebra
 using LinearAlgebra: AdjOrTrans, matprod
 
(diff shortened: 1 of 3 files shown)


dkarrasch:
If we backport this and the fix in Base, do we need a workaround, or are both fixes released simultaneously and we can switch signs constantly?

dkarrasch:
I see, the Base fix is not merged yet, so I guess you are planning to rerun CI when that is merged and afterward remove the temporary workarounds before merging this?

fredrikekre:
Base fix is failing due to SparseArrays, so have to merge here first, upgrade in base, then merge the fix...

dkarrasch (approved):

PR #263, opened 2022-09-19
add undef initializers

matbesancon:
This is useful to have a unified way to create sparse and dense vectors.

Will add tests if the feature is considered useful / needed


Wimmerer:
Is this semantically correct compared to other undef initializers? If so this seems like a fine addition.

matbesancon:
> Is this semantically correct compared to other undef initializers

it seems so, the type has to be given as type parameter, it creates a new object of the same type

Wimmerer:
I'll approve with a few tests and maybe a short docstring to clarify that this is an empty spmatrix.

matbesancon:
done

PR #271, opened 2022-09-28
Fix `all`/`any` on empty sparse arrays

### LINKED ISSUE #270: New error in reductions of empty sparse array
`master`:
```
julia> using SparseArrays; all(iszero, spzeros(0, 0))
ERROR: MethodError: reducing over an empty collection is not allowed; consider supplying `init` to the reducer
```
`1.8.1`:
```
julia> using SparseArrays; all(iszero, spzeros(0, 0))
true
```

`git bisect` points to 0d63db09beaaa13f8104f4e700aa7b7f1a3a69f1 (cc @dkarrasch).

Noticed in https://github.com/Ferrite-FEM/Ferrite.jl/issues/492.

dkarrasch:
Fixes #270.


ViralBShah:
Yes, we should merge this.

Wimmerer:
After this I'd like to bumpstdlibs and run a pkgeval.

PR #252, opened 2022-09-08
Backports for Julia v1.8.x

Wimmerer:
Co-authored-by: Steven G. Johnson <stevenj@mit.edu>

Note: I want to be sure the attribution is correct here, I *think* it should say "authored by Jerry and Steven" and "committed by Wimmerer".

Backported PRs:

- [x] https://github.com/JuliaSparse/SparseArrays.jl/pull/93
- [x] https://github.com/JuliaSparse/SparseArrays.jl/pull/253
- [x] https://github.com/JuliaSparse/SparseArrays.jl/pull/260
- [x] https://github.com/JuliaSparse/SparseArrays.jl/pull/176

PR #272, opened 2022-09-29
Remove temporary circshift direction workaround

dkarrasch:
Will the Base PR be backported to v1.8.2, or not? I.e., can this be included in our backport PR, which hopefully still makes it for v1.8.2?


fredrikekre:
Probably too late for 1.8.2, but will be in 1.8.3 I suppose. At this point I believe #252 is too late for 1.8.2 too, unfortunately.

dkarrasch:
Just for reference: the Base PR is going to be included in v1.8.3.

PR #276, opened 2022-10-19
Add a non-compute constructor to UmfpackLU

Wimmerer:


Wimmerer:
Merging and bumping to fix downstream issues for pkgeval

PR #280, opened 2022-10-21
update umfpack doc, change copy defualts

SobhanMP:
@Wimmerer please don't set enable copynumer/symbolic in copy, your just making multithreading harder.

### DIFF
--- a/src/solvers/umfpack.jl
+++ b/src/solvers/umfpack.jl
@@ -194,9 +194,9 @@ _isnotnull(x::Union{Symbolic, Numeric}) = x.p != C_NULL
 """
 Working space for Umfpack so `ldiv!` doesn't allocate.
 
-To use multiple threads, each thread should have their own workspace that can be allocated using `Base.similar(::UmfpackWS)`
-and passed as a kwarg to `ldiv!`. Alternativly see `copy(::UmfpackLU)`. The constructor is overloaded so to create appropriate
-sized working space given the lu factorization or the sparse matrix and if refinement is on.
+To use multiple threads, each thread should have their own workspace this can be done using  `copy(::UmfpackLU)`
+The constructor is overloaded so to create appropriate sized working space given the lu
+factorization or the sparse matrix and if refinement is on.
 """
 struct UmfpackWS{T<:UMFITypes}
     Wi::Vector{T}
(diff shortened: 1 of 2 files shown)


Wimmerer:
Given that we weren't even thread safe until 1.9 and only a few people noticed I would argue that the number of people doing multithreaded UMFPACK solves is very very few. Two keyword arguments doesn't seem to be onerous and prevents anyone from accidentally segfaulting.

SobhanMP:
given the docstring of copy

> Create a shallow copy of x: the outer structure is copied, but not all internal values. For example, copying an array produces a new array with identically-same elements as the original.

I think doing less work is best IMHO

SobhanMP:
The problem isn't that adding two copy kwargs (shouldn't it be just one for both?), it's that you have to ``know'' that you have to add those kwargs otherwise it will redo the factorization so many times.

SobhanMP:
@ViralBShah do you also agree with Will?

Wimmerer:
To be honest it doesn't matter to me much, we didn't even have the method defined previously so the chance of race conditions happening is slim to none. I do think the kwargs aren't too bad with the "safe" defaults given you, me and maybe 2 other users will ever use that method *directly*. But for that same reason the slightly more dangerous defaults likely won't impact anyone.

SobhanMP:
yeah, honestly if it's well documented, it'll be fine.

Wimmerer:
https://github.com/JuliaSparse/SparseArrays.jl/pull/259.

SobhanMP:
@Wimmerer How about not having those keys? if the operations are implemented like that, they will be still thread safe.

SobhanMP:
also isn't the current implementation leaking numerics with lu!?

Wimmerer (approved):
Indeed :). Re-solves are rare but this would have appeared in SciML so thanks. Not sure where I got it in my head that umfpack would modify the numeric object. A little surprising it refuses to reuse that memory ha.

SobhanMP:
but honestly, the idea of making the symbolic/numeric different types with their own finalizer is pretty smart

Wimmerer:
It's much nicer to have finalizers be as granular as possible, and *not* at the level of user objects. That way you can have multiple references to the same Numeric object and not have any issues with tracking aliasing.

Figured that out the long/hard way with SuiteSparseGraphBLAS

PR #283, opened 2022-10-27
Fix order of triangular and adjortrans wrappers

dkarrasch:
I'm not sure why this hasn't come up anywhere, but it may be that it simply lead to performance penalties. In any case, for a while adjoints of triangular matrices are passed to the wrapped matrix while the triangular wrapper is flipped. This should be backported to v1.8.

### DIFF
--- a/src/linalg.jl
+++ b/src/linalg.jl
@@ -543,9 +543,9 @@ end
 
 # forward multiplication for adjoint and transpose of LowerTriangular CSC matrices
 function _lmul!(U::UpperTriangularWrapped, B::StridedVecOrMat)
-    A = U.parent.data
+    A = parent(parent(U))
     unit = U.parent isa UnitDiagonalTriangular
-    adj = U isa Adjoint
+    adj = parent(U) isa Adjoint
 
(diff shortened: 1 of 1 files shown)


dkarrasch:
Hm, the swap already happened in the v1.6 cycle: https://github.com/JuliaLang/julia/pull/38168, so we need to backport this.

Wimmerer:
So we need to queue this up for backports to 1.6, 1.7 and 1.8?

dkarrasch:
How embarrising, and the fault is all mine:

```julia
julia> @which lmul!(UpperTriangular(sprand(10,10,0.3))', zeros(10,4))
lmul!(xA::LowerTriangular{var"#s804", var"#s803"} where {var"#s804", var"#s803"<:Adjoint}, B::StridedVecOrMat{T} where T) in LinearAlgebra at /Applications/Julia-1.6.app/Contents/Resources/julia/share/julia/stdlib/v1.6/LinearAlgebra/src/triangular.jl:1013
```

> So we need to queue this up for backports to 1.6, 1.7 and 1.8?

Yes, for 1.6 and 1.8.

PR #284, opened 2022-10-30
Improved memory usage when creating the sparsity pattern

lijas:
This PR makes it more memory efficient to create the sparsity pattern,  i.e. when you dont care about the values of `V`.

For a cube mesh with 70x70x70 elements (and linear interpolation with 1 dof per node), i get these benchmarks:
```
@show 2*Base.summarysize(I)/(1024^2)        #Memory of I and J
@btime sparse($I, $J, 1.0)                  # This PR
@btime sparse($I, $J, fill(1.0, length(J))) # Previously
```

```
(2 * Base.summarysize(I)) / 1024 ^ 2 = 340.42230224609375 #MB
200.528 ms (17 allocations: 321.74 MiB)
318.534 ms (21 allocations: 662.17 MiB)

357911×357911 SparseMatrixCSC{Float64, Int64} with 9393931 stored entries:
⎡⠻⣦⡀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⎤
⎢⠀⠈⠻⣦⡀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⎥
⎢⠀⠀⠀⠈⠻⣦⡀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⎥
⎢⠀⠀⠀⠀⠀⠈⠻⣦⡀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⎥
⎢⠀⠀⠀⠀⠀⠀⠀⠈⠻⣦⡀⠀⠀⠀⠀⠀⠀⠀⠀⠀⎥
⎢⠀⠀⠀⠀⠀⠀⠀⠀⠀⠈⠻⣦⡀⠀⠀⠀⠀⠀⠀⠀⎥
⎢⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠈⠻⣦⡀⠀⠀⠀⠀⠀⎥
⎢⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠈⠻⣦⡀⠀⠀⠀⎥
⎢⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠈⠻⣦⡀⠀⎥
⎣⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠈⠻⣦⎦
```

I can't get the tests to pass (I get a bunch of ambiguity errors that I dont understand how to fix), but if you like this PR I will fix them.

### DIFF
--- a/src/sparsematrix.jl
+++ b/src/sparsematrix.jl
@@ -1000,7 +1000,7 @@ julia> sparse(Is, Js, Vs)
 sparse(I::AbstractVector{Ti}, J::AbstractVector{Ti}, V::AbstractVector{Tv}, m::Integer, n::Integer, combine) where {Tv,Ti<:Integer} =
     _sparse(I, J, V, m, n, combine)
 sparse(I::AbstractVector{Ti}, J::AbstractVector{Ti}, v::Tv, m::Integer, n::Integer, combine) where {Tv<:Number,Ti<:Integer} = 
-    _sparse(I, J, iszero(v) ? 0.0 : fill(v,length(I)), Int(m), Int(n), combine)
+    _sparse(I, J, iszero(v) ? v : fill(v,length(I)), Int(m), Int(n), combine)
 
 function _sparse(I::AbstractVector{Ti}, J::AbstractVector{Ti}, V::Union{Tv,AbstractVector{Tv}}, m::Integer, n::Integer, combine) where {Tv,Ti<:Integer}
     require_one_based_indexing(I, J, V)
--- a/src/sparsematrix.jl
+++ b/src/sparsematrix.jl
@@ -1093,7 +1093,7 @@ F. Gustavson, "Two fast algorithms for sparse matrices: multiplication and permu
 transposition," ACM TOMS 4(3), 250-269 (1978) inspired this method's use of a pair of
 counting sorts.
 """
-function sparse!(I::AbstractVector{Ti}, J::AbstractVector{Ti}, V::Union{Type{Tv},AbstractVector{Tv}}, 
+function sparse!(I::AbstractVector{Ti}, J::AbstractVector{Ti}, V::Union{Tv,AbstractVector{Tv}}, 
         m::Integer, n::Integer, combine, klasttouch::Vector{Tj},
         csrrowptr::Vector{Tj}, csrcolval::Vector{Ti}, csrnzval::Vector{Tv},
         csccolptr::Vector{Ti}, cscrowval::Vector{Ti}, cscnzval::Vector{Tv}) where {Tv,Ti<:Integer,Tj<:Integer}
@@ -1227,7 +1227,7 @@ end
 
 dimlub(I) = isempty(I) ? 0 : Int(maximum(I)) #least upper bound on required sparse matrix dimension
 
-sparse(I,J,v::Number) = sparse(I, J, fill(v,length(I)))
+sparse(I,J,v::Number) = sparse(I, J, v, dimlub(I), dimlub(J))
 
 sparse(I,J,V::AbstractVector) = sparse(I, J, V, dimlub(I), dimlub(J))
 
--- a/src/sparsematrix.jl
+++ b/src/sparsematrix.jl
@@ -1098,7 +1098,7 @@ function sparse!(I::AbstractVector{Ti}, J::AbstractVector{Ti}, V::Union{Tv,Abstr
         csrrowptr::Vector{Tj}, csrcolval::Vector{Ti}, csrnzval::Vector{Tv},
         csccolptr::Vector{Ti}, cscrowval::Vector{Ti}, cscnzval::Vector{Tv}) where {Tv,Ti<:Integer,Tj<:Integer}
 
-    require_one_based_indexing(I, J, (V isa Type) ? 0 : V)
+    require_one_based_indexing(I, J, V)
     sparse_check_Ti(m, n, Ti)
     sparse_check_length("I", I, 0, Tj)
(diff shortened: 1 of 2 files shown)


Wimmerer:
Nice optimization! If you post the ambiguities I can give you pointers in the right direction.

lijas:
Thank you, but I have direct access to @fredrikekre, so I might bother him first :)

lijas:
Here is the ambiguity error:
```
sparse([1,2,3], [1,2,3], [1,2,3], 3, 3)
ERROR: MethodError: sparse(::Vector{Int64}, ::Vector{Int64}, ::Vector{Int64}, ::Int64, ::Int64, ::typeof(+)) is ambiguous. Candidates:
  sparse(I::AbstractVector{Ti}, J::AbstractVector{Ti}, V::Union{AbstractVector{Tv}, Tv}, m::Integer, n::Integer, combine) where {Tv, Ti<:Integer} in SparseArrays at /home/elias/Documents/julia_packages/SparseArrays.jl/src/sparsematrix.jl:998
  sparse(I::AbstractVector, J::AbstractVector, V::Union{Number, AbstractVector}, m::Integer, n::Integer, combine) in SparseArrays at /home/elias/Documents/julia_packages/SparseArrays.jl/src/sparsematrix.jl:1037
Possible fix, define
  sparse(::AbstractVector{Ti}, ::AbstractVector{Ti}, ::Union{Number, AbstractVector}, ::Integer, ::Integer, ::Any) where Ti<:Integer
```

It feels like [998](https://github.com/JuliaSparse/SparseArrays.jl/blob/c41b71cac2df93f0d55e2dd95315e7d918ec2cea/src/sparsematrix.jl#L998) is more specific than [1037](https://github.com/JuliaSparse/SparseArrays.jl/blob/c41b71cac2df93f0d55e2dd95315e7d918ec2cea/src/sparsematrix.jl#L1037) so I don't really understand the ambiguity.

dkarrasch:
In the third argument, `Tv` is not a subtype of `Number`, hence the fix proposal.

lijas:
So I think I solved it by changing [1037](https://github.com/JuliaSparse/SparseArrays.jl/blob/c41b71cac2df93f0d55e2dd95315e7d918ec2cea/src/sparsematrix.jl#L1037) by removing the type-annotation for the third argument.

But now I ran into an interesting test fail:
```
@test sparse([1, 1, 2, 2, 2], [1, 2, 1, 2, 2], 1.0, 2, 2, +) == sparse([1, 1, 2, 2], [1, 2, 1, 2], [1.0, 1.0, 1.0, 2.0], 2, 2)

julia> sparse([1, 1, 2, 2], [1, 2, 1, 2], [1.0, 1.0, 1.0, 2.0], 2, 2)
2×2 SparseMatrixCSC{Float64, Int64} with 4 stored entries:
 1.0  1.0
 1.0  2.0

julia> sparse([1, 1, 2, 2, 2], [1, 2, 1, 2, 2], 1.0, 2, 2, +)
2×2 SparseMatrixCSC{Float64, Int64} with 4 stored entries:
 1.0  1.0
 1.0  1.0

julia> sparse([1, 1, 2, 2, 2], [1, 2, 1, 2, 2], [1.0,1.0,1.0,1.0,1.0], 2, 2, +)
2×2 SparseMatrixCSC{Float64, Int64} with 4 stored entries:
 1.0  1.0
 1.0  2.0
```

I doubt anyone is setting V=1.0 (Number) and expecting to sum the values at duplicate positions :thinking: ?
But maybe it is better to create a new method `sparse(I,J, m,n, combine)` ?

KristofferC:
Maybe just `sparse(Float64, I,J, m,n)` and that sets `Float64(0.0)` at the positions given by `I` and `J`? I doubt specifying a concrete number is that common, or?

But I also don't think the current behavior is so bad. You can always fix your combine function if you don't want them to add.

lijas:
I like `sparse(Float64, I, J, m, n)`... it is a common pattern in julia to add the type first (zeros, ones etc).

Yeah, I doubt anyone is doing `sparse(I, J, v::Number)` with `v>0` , but if there is, this PR would be silently be breaking.

src/sparsematrix.jl
```diff
@@ -995,14 +995,18 @@ julia> sparse(Is, Js, Vs)
  ⋅  ⋅  3
 ```
 """
-function sparse(I::AbstractVector{Ti}, J::AbstractVector{Ti}, V::AbstractVector{Tv}, m::Integer, n::Integer, combine) where {Tv,Ti<:Integer}
+sparse(I::AbstractVector{Ti}, J::AbstractVector{Ti}, V::AbstractVector{Tv}, m::Integer, n::Integer, combine) where {Tv,Ti<:Integer} =
+    _sparse(I, J, V, m, n, combine)
+sparse(I::AbstractVector{Ti}, J::AbstractVector{Ti}, v::Tv, m::Integer, n::Integer, combine) where {Tv<:Number,Ti<:Integer} = 
+    _sparse(I, J, iszero(v) ? 0.0 : fill(v,length(I)), Int(m), Int(n), combine)
```
dkarrasch, line 1001: `iszero(v) ? 0.0 : fill(v,length(I))` doesn't look type stable? Perhaps `zero(Tv)`, or just `v`?

fredrikekre:
Perhaps this could be a method of `spzeros`? `spzeros(I, J, m, n)` feels like a pretty nice way to create the pattern only.

PR #288, opened 2022-11-17
Fix "no method matching" test

apaz-cli:
Right now we're thinking about changing the `MethodError` message. Unfortunately, this breaks a lot of tests, including this one. Discussion at https://github.com/JuliaLang/julia/pull/47369.

~~If the change below does happen, merge this PR afterwards.
https://github.com/JuliaLang/julia/pull/47369/commits/44367ce8fed017c1aee6a01fb273b12653e7f5c9~~


ViralBShah:
@vtjnash is this good to merge?

PR #297, opened 2022-11-24
Backports for release 1.8.4

dkarrasch:


fredrikekre:
I'll merge this so we don't miss 1.8.4. If there are more backports and there is time we can make a second backport PR to release-1.8.

PR #300, opened 2022-11-24
Backports for release 1.9

dkarrasch:


dkarrasch:
We currently don't have a pre-release of v1.9, so tests obviously fail. Once we have, then we should restrict testing to v1.9 and exclude nightly, to avoid visual clutter and test noise.

.github/workflows/ci.yml
```diff
@@ -22,7 +22,7 @@ jobs:
       fail-fast: false
       matrix:
         version:
-          - '1'
+          - '1.9'
```
fredrikekre, line 25:
```suggestion
          - '~1.9.0-0'
```

fredrikekre:
Probably good to merge here and update in JuliaLang/julia sooner rather than later so we don't miss 1.9.

.github/workflows/ci.yml
dkarrasch: Awesome!

dkarrasch:
Perhaps we should finalize #335, include it here and then go.

KristofferC:
Can this be updated, merged and bumped on `release-1.9`?

dkarrasch:
We don't squash this, right? Just create a merge commit. And how do we bump it over at `backports-release-1.9`? With the bot?

wsshin:
Just wondering if this is going to be merged in 1.9-rcX.  1.9-rc2 was released yesterday.  Previous Julia major releases usually had 3–4 release candidates, which means that we will probably have just one or two more release candidates.  I think it will be nice to test these backports in release candidates before the official release.

PR #309, opened 2022-12-02
document rfn argument of sprand

stevengj:
I noticed that the `rfn` argument of `sprand` was poorly documented:

1. We didn't say what its arguments should be.
2. We didn't say clearly that it defaults to `rand`
3. The type `T` argument goes in a different place if `rfn` is passed.   Moreover, although in this case you can pass `T` as the *last* argument, it seems that it doesn't really work — if it doesn't match the return type of `rfn`, it is ignored!  So, it seems better to leave that undocumented for now until we decide what to do with it.

### DIFF
--- a/src/sparsematrix.jl
+++ b/src/sparsematrix.jl
@@ -1888,14 +1888,24 @@ function sparse_sortedlinearindices!(I::Vector{Ti}, V::Vector, m::Int, n::Int) w
 end
 
 """
-    sprand([rng],[type],m,[n],p::AbstractFloat,[rfn])
+    sprand([rng],[T::Type],m,[n],p::AbstractFloat)
+    sprand([rng],m,[n],p::AbstractFloat,[rfn=rand])
 
 Create a random length `m` sparse vector or `m` by `n` sparse matrix, in
 which the probability of any element being nonzero is independently given by
-`p` (and hence the mean density of nonzeros is also exactly `p`). Nonzero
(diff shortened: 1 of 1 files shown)


stevengj:
See also https://github.com/JuliaLang/julia/pull/30637

src/sparsematrix.jl
```diff
@@ -1888,14 +1888,24 @@ function sparse_sortedlinearindices!(I::Vector{Ti}, V::Vector, m::Int, n::Int) w
...
+
+By default, nonzero values are sampled from a uniform distribution using
+the [`rand`](@ref) function, i.e. by `rand(T)`, or `rand(rng, T)` if `rng`
+is supplied; for the default `T=Float64`, this corresponds to nonzero values
+sampled uniformly in `[0,1)`.
+
+You can sample nonzero values from a different distribution by passing a
+custom `rfn` function instead of `rand`.   This should be a function `rfn(n)`
+that returns an array of `n` random numbers sampled from the desired distribution;
+alternatively, if `rng` is supplied, it should be a function `rfn(rng, n)`.
```
stevengj, line 1908:
```suggestion
You can sample nonzero values from a different distribution by passing a
custom `rfn` function instead of `rand`.   This should be a function `rfn(k)`
that returns an array of `k` random numbers sampled from the desired distribution;
alternatively, if `rng` is supplied, it should instead be a function `rfn(rng, k)`.
```
Better not to use `n` here since that's the size of the matrix.

PR #320, opened 2023-01-04
Make `cholmod` vector solve return vector

dkarrasch:
Fixes https://github.com/JuliaLang/julia/issues/48095.

### DIFF
--- a/src/solvers/cholmod.jl
+++ b/src/solvers/cholmod.jl
@@ -907,10 +907,10 @@ function Hermitian{Tv,SparseMatrixCSC{Tv,SuiteSparse_long}}(A::Sparse{Tv}) where
 end
 convert(T::Type{Hermitian{Tv,SparseMatrixCSC{Tv,SuiteSparse_long}}}, A::Sparse{Tv}) where {Tv<:VTypes} = T(A)
 
-function sparsevec(A::Sparse{Float64})
+function sparsevec(A::Sparse{Tv}) where {Tv}
     s = unsafe_load(pointer(A))
     @assert s.stype == 0
-    return SparseVector{Float64,SuiteSparse_long}(A)
+    return SparseVector{Tv,SuiteSparse_long}(A)
 end
 
 function sparse(A::Sparse{Float64}) # Notice! Cannot be type stable because of stype
--- a/src/solvers/cholmod.jl
+++ b/src/solvers/cholmod.jl
@@ -28,7 +28,8 @@ export
     Factor,
     Sparse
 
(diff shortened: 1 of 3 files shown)


src/solvers/cholmod.jl
```diff
@@ -1526,7 +1544,10 @@ end
 function (\)(L::FactorComponent, B::Matrix)
     Matrix(L\Dense(B))
 end
-function (\)(L::FactorComponent, B::SparseVecOrMat)
+function (\)(L::FactorComponent, B::SparseVector)
+    sparsevec(L\Sparse(B))
```
stevengj, line 1548: I'm a little confuse about why we are returning a sparse vector — isn't the result generally dense?

src/solvers/cholmod.jl
```diff
@@ -891,10 +903,16 @@ function Hermitian{Tv,SparseMatrixCSC{Tv,SuiteSparse_long}}(A::Sparse{Tv}) where
     ishermitian(A) || throw(ArgumentError("matrix is not Hermitian"))
     args = _extract_args(s, Tv)
     s.sorted == 0 && _sort_buffers!(args...)
-    Hermitian(_trim_nz_builder!(args...), s.stype > 0 ? :U : :L)
+    Hermitian(SparseMatrixCSC(_trim_nz_builder!(args...)...), s.stype > 0 ? :U : :L)
 end
 convert(T::Type{Hermitian{Tv,SparseMatrixCSC{Tv,SuiteSparse_long}}}, A::Sparse{Tv}) where {Tv<:VTypes} = T(A)
 
+function sparsevec(A::Sparse{Float64})
```
stevengj, line 910: Why not `Sparse{Tv}`?

src/solvers/cholmod.jl
dkarrasch: I was confused about that myself, but note that previously we returned a `SparseMatrixCSC`. This change only changes the dimension of the returned array. But, of course, we may as well change to dense return arrays. The actual type of the return value doesn't seem to be tested, so I don't know what "promises" we made.

src/solvers/cholmod.jl
dkarrasch: True. I was following the pattern below, but in any case the complex version is missing, and also this function by itself need not be restricted.

src/solvers/cholmod.jl
stevengj: I guess we might as well stay consistent here.

See also https://discourse.julialang.org/t/sparse-solve-with-sparse-rhs/81595/3?u=stevengj

stevengj:
LGTM.

ViralBShah:
Merge?

PR #322, opened 2023-01-05
RFC: Handle `Hermitian(::Sparse) \ SparseArray`

dkarrasch:
While playing with https://github.com/JuliaLang/julia/issues/48095 a little more I noticed one peculiarity.

On v1.8

```julia
Hermitian(hess_mat) \ hcat(grad_vec, grad_vec)
ERROR: MethodError: no method matching lu!(::SparseMatrixCSC{Float64, Int64}, ::RowMaximum; check=true)
Closest candidates are:
  lu!(::StridedMatrix{T}, ::RowMaximum; check) where T<:Union{Float32, Float64, ComplexF32, ComplexF64} at /Applications/Julia-1.8.app/Contents/Resources/julia/share/julia/stdlib/v1.8/LinearAlgebra/src/lu.jl:80
  lu!(::Union{Hermitian{T, S}, Symmetric{T, S}} where {T, S}, ::Union{NoPivot, RowMaximum}; check) at /Applications/Julia-1.8.app/Contents/Resources/julia/share/julia/stdlib/v1.8/LinearAlgebra/src/lu.jl:89
  lu!(::StridedMatrix{T} where T, ::Union{NoPivot, RowMaximum}; check) at /Applications/Julia-1.8.app/Contents/Resources/julia/share/julia/stdlib/v1.8/LinearAlgebra/src/lu.jl:135
  ...
Stacktrace:
 [1] lu!(A::Hermitian{Float64, SparseMatrixCSC{Float64, Int64}}, pivot::RowMaximum; check::Bool)
   @ LinearAlgebra /Applications/Julia-1.8.app/Contents/Resources/julia/share/julia/stdlib/v1.8/LinearAlgebra/src/lu.jl:91
 [2] lu(A::Hermitian{Float64, SparseMatrixCSC{Float64, Int64}}, pivot::RowMaximum; check::Bool)
   @ LinearAlgebra /Applications/Julia-1.8.app/Contents/Resources/julia/share/julia/stdlib/v1.8/LinearAlgebra/src/lu.jl:279
 [3] lu(A::Hermitian{Float64, SparseMatrixCSC{Float64, Int64}}, pivot::RowMaximum) (repeats 2 times)
   @ LinearAlgebra /Applications/Julia-1.8.app/Contents/Resources/julia/share/julia/stdlib/v1.8/LinearAlgebra/src/lu.jl:278
 [4] \(A::Hermitian{Float64, SparseMatrixCSC{Float64, Int64}}, B::SparseMatrixCSC{Float64, Int64})
   @ LinearAlgebra /Applications/Julia-1.8.app/Contents/Resources/julia/share/julia/stdlib/v1.8/LinearAlgebra/src/generic.jl:1110
```

On v1.9:

```julia
julia> Hermitian(hess_mat) \ hcat(grad_vec, grad_vec)
6×2 Matrix{Float64}:
 0.508367  0.508367
 0.508367  0.508367
 2.0       2.0
 0.5       0.5
 0.307692  0.307692
 0.5       0.5
```

While this works, the problem is that this runs by `_generic_lufact!` due to the relaxations in https://github.com/JuliaLang/julia/pull/47107. We have two options: (i) throw and suggest to densify the rhs, or (ii) densify silently. I think (ii) is not too bad, because if a user expects the solve to be successful, then a dense version of the result will be typically more memory efficient than a sparse version thereof due to intrinsic density.


dkarrasch:
Whatever we decide on, this should be backported to v1.9 I guess.

dkarrasch:
Since this was not working on earlier versions, I think it's more natural to make this throw for now, as a quickfix. The general case about what to do with sparse rhs should be discussed first, and then probably requires modifications in more places than this.

ViralBShah:
Sparse RHS can be optimized, but usually I believe you need to be able to leverage that in your solver as discussed in the Sparse Matrices in Matlab paper.

@DrTimothyAldenDavis @Wimmerer IIRC SuiteSparse doesn't support sparse RHS in solvers. I just wanted to confirm or be corrected.

SobhanMP (commented):
lgtm

DrTimothyAldenDavis:
I do exploit the sparse right-hand-side internally in several places (it forms the key kernel for cs_lu and KLU for example, and is called n times to do the entire LU factorization, solving x = L\b where x, L, and b are all saprse).  I also exploit it inside CHOLMOD in the update/downdate.  However, it's hard to exploit in general for solving x=A\b because if b is sparse and A forms a single strongly-connected component (the typical case) then x is dense.  So it's hard to exploit.  x can be sparse for x=A\b if A is permutable into block triangular form (via cs_dmperm for example).    I use the BTF form in KLU+BTF but I can't recall if I also allow b to be sparse when doing x=A\b.

Wimmerer:
@DrTimothyAldenDavis I don't see any support for it in the KLU interface, but I may experiment with doing it with the Julia port.

Wimmerer (approved):
lgtm

DrTimothyAldenDavis:
OK .. then I guess I don't have one.  I think I may have one for UMFPACK but I convert it to full first ...  I can't recall exactly ... too much code over too many years.

dkarrasch:
Making this throw (now) is only a continuation of the same earlier behavior. We can still always make it work if we want by either calling specific code paths, or densifying the rhs.

Issue #323, opened 2023-01-09
Wrong answer with `\` correct with `lu`


KristofferC:
I have a symmetric (non-posdef) sparse matrix `K` where I get the wrong solution to a linear system if I use `K \ f` but correct if I use `lu(K) \ f`.

The matrix is inside a zipped serialized Julia file [repro.jls.zip](https://github.com/JuliaSparse/SparseArrays.jl/files/10372404/repro.jls.zip) which should be deserializable on both 1.8 and 1.9.

Running:

```julia
using SparseArrays, Serialization, LinearAlgebra, IterativeSolvers
d = deserialize("repro.jls");
K = sparse(d["K"])
f = ones(size(K, 1));

u = K \ f;
u_lu = lu(K) \ f;
u_dense = Matrix(K) \ f;
u_gmres = gmres(K, f, restart=200)

compute_residual(K, u, f) = norm(f - K * u) / norm(f);

@show compute_residual(K, u, f)
# 314.5598350582391
@show compute_residual(K, u_lu, f)
# 1.9400817552325415e-12
@show compute_residual(K, u_dense, f)
# 3.31677221636043e-13
@show compute_residual(K, u_gmres, f)
# 9.888236216146402e-9
```

shows the residuals in the comments. As can be seen,  `u` when computed as `K \ f` gives a very bad residual while both `lu` and the dense case gives good answers. Even the iterative solvers gets a decent result (although it requires boosting the default values a bit).

Looking a bit what is happening, since this matrix is hermitian we enter:

https://github.com/JuliaSparse/SparseArrays.jl/blob/82ae2718c2d9e7e1edd38b9e0fcacd06a913dc20/src/linalg.jl#L1607

and then

https://github.com/JuliaSparse/SparseArrays.jl/blob/a3116b95add064055c1f737abd98b5a5839c4147/src/solvers/cholmod.jl#L1575-L1587

where the `cholesky` call fails but the `ldlt!` call succeeds. It seems to me that this call to `ldlt!` should not succeed since it gives a bad answer.

This can very well be some upstream thing.

DrTimothyAldenDavis:
It's not easy for me to replicate this in julia, but if you can post the matrix in Matrix Market format, I could take a look.  It looks to me that julia is using ldlt incorrectly; it should check the recip. condition estimate I provide.

PR #325, opened 2023-01-10
do not call `ldlt!` in the poly algorithm for hermitian matrices

KristofferC:
The current usage of `ldlt!` is not correct according to the test failure from the matrix in https://github.com/JuliaSparse/SparseArrays.jl/issues/323 and the comment in https://github.com/DrTimothyAldenDavis/SuiteSparse/issues/271#issuecomment-1376380786 by @DrTimothyAldenDavis.

This PR removes the usage of `ldlt!` in the poly algorithm as well as adds a test with the matrix that had the issue.


andreasnoack (approved):

KristofferC:
@dkarrasch, any opinions about this?

dkarrasch:
I think that's fine, and it reminds me of https://github.com/JuliaLang/julia/blob/25b27468ea515e54de1b7aa6a521eb333415c4b4/stdlib/LinearAlgebra/src/dense.jl#L1392-L1395. The underlying reason seems to be the same: for robust LDLt/"Bunchkaufman" one may need 2x2 pivots, which we currently don't have in the non-dense case.

PR #327, opened 2023-01-17
Allow real adjoint cholmod solve with complex rhs

dkarrasch:
This extends the same behavior of `Factor` to adjoints of `Factor`s, and helps to prevent a potential method ambiguity in https://github.com/JuliaLang/julia/pull/46874.


dkarrasch:
Is there anything controversial here or can I just merge?

ViralBShah:
Thanks

PR #312, opened 2022-12-05
Faster collect

KnutAM:
For vector `sv=sprand(n,0.2)`:
```julia
Vector: n = 16
master:   119.365 ns (1 allocation: 192 bytes)
pr:       45.455 ns (1 allocation: 192 bytes)
Vector: n = 65536
master:   2.682 ms (2 allocations: 512.05 KiB)
pr:       37.200 μs (2 allocations: 512.05 KiB)
```
<details>
<summary>Benchmark code with more cases</summary>

```julia
julia> using SparseArrays, BenchmarkTools
julia> function collect_pr(x::Union{AbstractSparseVector,AbstractSparseMatrix})
           if Base.has_offset_axes(x)
               return Base._collect_indices(axes(x), x)
           else
               return Array(x)
           end
       end;
julia> for n in [16^i for i in 1:4]
        sv = sprand(n, 0.2)
        println("Vector: n = $n")
        print("master: "); @btime collect($sv)
        print("pr:     "); @btime collect_pr($sv)
        if n <= 16^3
         println("Matrix: n = $n")
         sm = sprand(n, n, 0.2)
         print("master: "); @btime collect($sm)
         print("pr:     "); @btime collect_pr($sm)
        end
       end
Vector: n = 16
master:   119.365 ns (1 allocation: 192 bytes)
pr:       45.455 ns (1 allocation: 192 bytes)
Matrix: n = 16
master:   660.606 ns (1 allocation: 2.12 KiB)
pr:       660.355 ns (1 allocation: 2.12 KiB)
Vector: n = 256
master:   2.456 μs (1 allocation: 2.12 KiB)
pr:       365.502 ns (1 allocation: 2.12 KiB)
Matrix: n = 256
master:   120.700 μs (2 allocations: 512.05 KiB)
pr:       121.700 μs (2 allocations: 512.05 KiB)
Vector: n = 4096
master:   108.600 μs (2 allocations: 32.05 KiB)
pr:       2.556 μs (2 allocations: 32.05 KiB)
Matrix: n = 4096
master:   64.762 ms (2 allocations: 128.00 MiB)
pr:       63.805 ms (2 allocations: 128.00 MiB)
Vector: n = 65536
master:   2.682 ms (2 allocations: 512.05 KiB)
pr:       37.200 μs (2 allocations: 512.05 KiB)
```
</details>

Not sure how to avoid using `Base._collect_indices` without just copying that code.
[...]

PR #328, opened 2023-01-17
Faster conversion to dense matrix

KnutAM:
Ref #312 @ViralBShah 
(I also added a test to cover the `collect` - calls)

```julia
using SparseArrays, BenchmarkTools
for n in [16^i for i in 1:3]
    print("n = $n: ")
    sm = sprand(n, n, 0.2)
    @btime collect($sm)
end
# main
n = 16:   367.136 ns (1 allocation: 2.12 KiB)
n = 256:   64.300 μs (2 allocations: 512.05 KiB)
n = 4096:   45.581 ms (2 allocations: 128.00 MiB)
# PR
n = 16:   261.872 ns (1 allocation: 2.12 KiB)
n = 256:   23.200 μs (2 allocations: 512.05 KiB)
n = 4096:   27.871 ms (2 allocations: 128.00 MiB)
```
Edit: Benchmark updated for changing `copyto!` instead

<details>
<summary>versioninfo</summary>

```julia
Julia Version 1.8.5
Commit 17cfb8e65e (2023-01-08 06:45 UTC)
Platform Info:
  OS: Windows (x86_64-w64-mingw32)
  CPU: 8 × 11th Gen Intel(R) Core(TM) i7-1185G7 @ 3.00GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-13.0.1 (ORCJIT, tigerlake)
  Threads: 1 on 8 virtual cores
Environment:
  JULIA_PKG_OFFLINE = false
  JULIA_PKG_USE_CLI_GIT = true
```
</details>

### DIFF
--- a/src/sparsematrix.jl
+++ b/src/sparsematrix.jl
@@ -641,6 +641,32 @@ function copyto!(dest::AbstractMatrix, Rdest::CartesianIndices{2},
     return dest
 end
 
+# Faster version for non-abstract Array and SparseMatrixCSC 
+function Base.copyto!(A::Array{T}, S::SparseMatrixCSC{<:Number}) where {T<:Number}
+    isempty(S) && return A
+    length(A) < length(S) && throw(BoundsError())
+    
+    # Zero elements that are also in S, don't change rest of A 
+    @inbounds for i in 1:length(S)
+        A[i] = zero(T)
+    end
+    # Copy the structural nonzeros from S to A using 
+    # the linear indices (to work when size(A)!=size(S))
+    num_rows = size(S,1)
+    rowval = getrowval(S)
+    nzval = getnzval(S)
+    linear_index_col0 = 0   # Linear index before column (linear index = linear_index_col0 + row)
+    for col in 1:size(S, 2)
+        for i in nzrange(S, col)
+            row = rowval[i]
+            val = nzval[i]
+            A[linear_index_col0+row] = val
+        end
+        linear_index_col0 += num_rows
+    end
+    return A
+end
+
 ## similar
 #
 # parent method for similar that preserves stored-entry structure (for when new and old dims match)
@@ -930,31 +956,6 @@ function Matrix(S::AbstractSparseMatrixCSC{Tv}) where Tv
     return A
 end
 Array(S::AbstractSparseMatrixCSC) = Matrix(S)
-
-function Base.copyto!(A::Array{T}, S::SparseMatrixCSC{<:Number}) where {T<:Number}
-    isempty(S) && return A
-    length(A) < length(S) && throw(BoundsError())
-    
-    # Zero elements that are also in S, don't change rest of A 
-    @inbounds for i in 1:length(S)
-        A[i] = zero(T)
-    end
-    # Copy the structural nonzeros from S to M using 
-    # the linear indices (to work when size(M)!=size(S))
-    num_rows = size(S,1)
-    colptr = getcolptr(S)
-    rowval = getrowval(S)
-    nzval = getnzval(S)
-    for col in 1:size(S, 2)
-        for i in nzrange(S, col)
-            row = rowval[i]
-            val = nzval[i]
-            linear_index = num_rows*(col-1)+row
-            A[linear_index] = val
-        end
-    end
-    return A
-end
 
 convert(T::Type{<:AbstractSparseMatrixCSC}, m::AbstractMatrix) = m isa T ? m : T(m)
 
--- a/src/sparsematrix.jl
+++ b/src/sparsematrix.jl
@@ -954,7 +954,7 @@ function Base.copyto!(M::Matrix{<:Number}, S::SparseMatrixCSC{T}) where T<:Numbe
     elseif length(M) >= length(S)
         m_view = view(M, 1:length(S))
         copyto!(m_view, S)
-        return m_view
+        return M
     else
         throw(BoundsError())
     end
--- a/src/sparsematrix.jl
+++ b/src/sparsematrix.jl
@@ -931,9 +931,16 @@ function Matrix(S::AbstractSparseMatrixCSC{Tv}) where Tv
 end
 Array(S::AbstractSparseMatrixCSC) = Matrix(S)
 
-# Copy the structural nonzeros from S to M
-# using the linear indices (to work when size(M)!=size(S))
-function _copy_nonzeros_to!(A::Array, S::SparseMatrixCSC)
+function Base.copyto!(A::Array{T}, S::SparseMatrixCSC{<:Number}) where {T<:Number}
+    isempty(S) && return A
+    length(A) < length(S) && throw(BoundsError())
+    
+    # Zero elements that are also in S, don't change rest of A 
+    @inbounds for i in 1:length(S)
+        A[i] = zero(T)
+    end
+    # Copy the structural nonzeros from S to M using 
+    # the linear indices (to work when size(M)!=size(S))
     num_rows = size(S,1)
     colptr = getcolptr(S)
     rowval = getrowval(S)
@@ -946,18 +953,6 @@ function _copy_nonzeros_to!(A::Array, S::SparseMatrixCSC)
             A[linear_index] = val
         end
     end
-    return A
-end
-
-function Base.copyto!(A::Array{T}, S::SparseMatrixCSC{<:Number}) where {T<:Number}
-    isempty(S) && return A
-    length(A) < length(S) && throw(BoundsError())
-    
-    # Zero elements that are also in S, don't change rest of A 
-    @inbounds for i in 1:length(S)
-        A[i] = zero(T)
-    end
-    _copy_nonzeros_to!(A, S)
     return A
 end
 
--- a/src/sparsematrix.jl
+++ b/src/sparsematrix.jl
@@ -931,10 +931,8 @@ function Matrix(S::AbstractSparseMatrixCSC{Tv}) where Tv
 end
 Array(S::AbstractSparseMatrixCSC) = Matrix(S)
 
-# Faster version for SparseMatrixCSC
-function Matrix(S::SparseMatrixCSC{Tv}) where Tv
-    n, m = size(S)
-    M = zeros(Tv, n, m);
+# Copy the structural nonzeros from S to M
+function _copy_nonzeros_to!(M::Matrix, S::SparseMatrixCSC)
     isempty(M) && return M
     colptr = getcolptr(S)
     rowval = getrowval(S)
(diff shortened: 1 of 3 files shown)


src/sparsematrix.jl
```diff
@@ -931,6 +931,24 @@ function Matrix(S::AbstractSparseMatrixCSC{Tv}) where Tv
 end
 Array(S::AbstractSparseMatrixCSC) = Matrix(S)
 
+# Faster version for SparseMatrixCSC
+function Matrix(S::SparseMatrixCSC{Tv}) where Tv
```
dkarrasch, line 935: What if you overload

```julia
Base.copyto!(A::Matrix, S::SparseMatrixCSC)
```

Then you get a fast constructor and a fast `copyto!` method at the same time.

KnutAM: That's a nice idea - I do have to handle the cases that A is larger than S in that case, but seems to work. 
If you think the current update is good, then I think it makes sense to use the same structure for the vector case (for which `copyto!` currently is slow as well)

src/sparsematrix.jl
```diff
@@ -931,6 +931,35 @@ function Matrix(S::AbstractSparseMatrixCSC{Tv}) where Tv
...
+
+function Base.copyto!(M::Matrix{<:Number}, S::SparseMatrixCSC{T}) where T<:Number
+    isempty(S) && return M
+    if size(M) == size(S)
+        fill!(M, zero(T))
+        _copy_nonzeros_to!(M, S)
+    elseif length(M) >= length(S)
+        m_view = view(M, 1:length(S))
+        copyto!(m_view, S)
+        return m_view
```
dkarrasch, line 957: This should return `M`, shouldn't it? And do we have a fast `copyto!` for the call one line above?

KnutAM: Yes it should (fixed, but perhaps I should be a test)
No, not yet (I see now that it is actually very slow). One option could be something like 
```julia
m_view = view(M, 1:length(S))
copyto!(m_view, sparsevec(S))
```
and then move the fast implementation of `Vector(SparseVector)` to `copyto!`, which we could define for 
`Union{Vector, SubArray{<:Any,1,<:Vector}, SparseVector`

dkarrasch: How about computing the linear index from `col` and `rowval`, like `ind = col*m+rowval[i]`, and then write into `M[ind]`? Would be good to avoid allocating an entire sparse array object.

KnutAM: Do you mean the `sparsevec` call? 
Are there cases when that doesn't share data with `S`?

KnutAM: Thanks for the tips @dkarrasch - now `copyto!(::Array, ::SparseMatrixCSC)` works for `Array` with only one function def. - much nicer. 
I'll do some cleanup (and potential fixes) and add tests when I have some more time.

Do you know if there already is an iterator to loop over the stored linear indices instead of calculating them manually? Would be nice in order to make `copyto!` work for a sparse vector as well?

fredrikekre: You can use [`LinearIndices`](https://docs.julialang.org/en/v1/base/arrays/#Base.LinearIndices), I think. Something like
```julia
linear_indices = LinearIndices(S)
for ...
    idx = linear_indices[row, col]
    A[idx] = val
```

src/sparsematrix.jl
```diff
@@ -931,6 +931,31 @@ function Matrix(S::AbstractSparseMatrixCSC{Tv}) where Tv
...
+    @inbounds for i in 1:length(S)
+        A[i] = zero(T)
+    end
+    # Copy the structural nonzeros from S to M using 
+    # the linear indices (to work when size(M)!=size(S))
+    num_rows = size(S,1)
+    colptr = getcolptr(S)
+    rowval = getrowval(S)
+    nzval = getnzval(S)
+    for col in 1:length(colptr)-1
```
fredrikekre, line 948:
```suggestion
    for col in 1:size(S, 2)
```

src/sparsematrix.jl
```diff
@@ -931,6 +931,31 @@ function Matrix(S::AbstractSparseMatrixCSC{Tv}) where Tv
...
+        for i in colptr[col]:(colptr[col+1]-1)
```
fredrikekre, line 949: This is `nzrange`
```suggestion
        for i in nzrange(S, col)
```

KnutAM: Thanks!

src/sparsematrix.jl
```diff
@@ -931,6 +931,31 @@ function Matrix(S::AbstractSparseMatrixCSC{Tv}) where Tv
...
+    # the linear indices (to work when size(M)!=size(S))
+    num_rows = size(S,1)
+    colptr = getcolptr(S)
+    rowval = getrowval(S)
+    nzval = getnzval(S)
+    for col in 1:size(S, 2)
+        for i in nzrange(S, col)
+            row = rowval[i]
+            val = nzval[i]
+            linear_index = num_rows*(col-1)+row
```
dkarrasch, line 952: Not sure it makes a big difference, but do you want to hoist out `linear_index = num_rows*(col-1)` out of the inner loop and write into `A[linear_index+row]`? (Or actually start with `linear_index = 0` outside of all loops and then `linear_index += num_rows` in the `col` loop? And perhaps mark the `i`-loop as `@inbounds`?

KnutAM: Thanks, the accumulation part is for sure nice! I'll update with that.

For the inbounds, I'm not sure which checks will be required in the outer loop? 
E.g., is `nzrange` safe for indexing `rowval` and `nzval`? And further, I don't understand/know if one can rely on 
`rowval[nzrange(S,col)]` to be sorted (and only check first and last value).

KnutAM: Now I also moved it to the other specialized copyto! methods (I think it is better to put them all together or?)
We don't need any `unalias` right? I could not construct a case to make, e.g. `nzvals` and `A` aliased.

src/sparsematrix.jl
KnutAM: Thanks!
I don't see how to write one method for both `SparseVector` and `SparseMatrixCSC` with that though - but I'm probably missing something. 

I was thinking about an iterator of the type `(ind, val) in indices_nzvals(S)` where `S[ind]==val`, in this case to be used as
```julia
for (ind, val) in indices_nzvals(S)
 A[ind] = val
end
```

dkarrasch (approved):
LGTM. What about the timings with the latest changes?

KnutAM:
> LGTM. What about the timings with the latest changes?

Unfortunately, I couldn't see any significant difference (which surprised me). 
```julia
A = Matrix{Float64}(undef, 100, 100);
S = sprand(100,100,0.2);
@btime copyto_1!($A, $S); # latest change
  2.022 μs (0 allocations: 0 bytes)
@btime copyto_2!($A, $S); # before latest change
  2.056 μs (0 allocations: 0 bytes)
@btime copyto_3!($A, $S); # latest change + inbounds
  1.770 μs (0 allocations: 0 bytes)
```

dkarrasch:
> Unfortunately, I couldn't see any significant difference (which surprised me).

Well, but it's not that many nonzeros that you need to copy, about 2k. I think we should do whatever is safe and whatever does not make it worse, but simpler. Last thing: I believe we can add the inbounds annotation, the ingredients for the two loops and the `setindex!` look safe.

ViralBShah:
Yeah try the benchmarks on something like sprand(1000, 1000, 0.5).

KnutAM:
Thank's for the tip!
I'm not sure why `@inbounds` doesn't help as much on this one, perhaps branch prediction?
```julia
n=1000; d=0.5;
 A = zeros(n,n); S=sprand(n,n,d);
julia> @btime copyto_1!($A, $S); # Current state
  795.100 μs (0 allocations: 0 bytes)

julia> @btime copyto_2!($A, $S); # Earlier (num_rows*(col-1) + row)
  792.900 μs (0 allocations: 0 bytes)

julia> @btime copyto_3!($A, $S); # Current state + inbounds
  775.700 μs (0 allocations: 0 bytes)
  ```
  
  <details>
  <summary> Function definitions </summary>
  
  ```julia
julia> function copyto_1!(A::Array{T}, S::SparseMatrixCSC{<:Number}) where {T<:Number}
           isempty(S) && return A
           length(A) < length(S) && throw(BoundsError())

           # Zero elements that are also in S, don't change rest of A
           @inbounds for i in 1:length(S)
               A[i] = zero(T)
           end
           # Copy the structural nonzeros from S to A using
           # the linear indices (to work when size(A)!=size(S))
           num_rows = size(S,1)
           rowval = getrowval(S)
           nzval = getnzval(S)
           linear_index_col0 = 0   # Linear index before column (linear index = linear_index_col0 + row)
           for col in 1:size(S, 2)
               for i in nzrange(S, col)
                   row = rowval[i]
                   val = nzval[i]
                   A[linear_index_col0+row] = val
               end
               linear_index_col0 += num_rows
           end
           return A
       end
copyto_1! (generic function with 1 method)

julia> function copyto_2!(A::Array{T}, S::SparseMatrixCSC{<:Number}) where {T<:Number}
           isempty(S) && return A
           length(A) < length(S) && throw(BoundsError())

           # Zero elements that are also in S, don't change rest of A
           @inbounds for i in 1:length(S)
               A[i] = zero(T)
           end
           # Copy the structural nonzeros from S to A using
           # the linear indices (to work when size(A)!=size(S))
           num_rows = size(S,1)
           rowval = getrowval(S)
           nzval = getnzval(S)
           for col in 1:size(S, 2)
               for i in nzrange(S, col)
                   row = rowval[i]
                   val = nzval[i]
                   A[num_rows*(col-1)+row] = val
               end
           end
           return A
       end
copyto_2! (generic function with 1 method)

julia> function copyto_3!(A::Array{T}, S::SparseMatrixCSC{<:Number}) where {T<:Number}
           isempty(S) && return A
           length(A) < length(S) && throw(BoundsError())

           # Zero elements that are also in S, don't change rest of A
           @inbounds for i in 1:length(S)
               A[i] = zero(T)
           end
           # Copy the structural nonzeros from S to A using
           # the linear indices (to work when size(A)!=size(S))
           num_rows = size(S,1)
           rowval = getrowval(S)
           nzval = getnzval(S)
           linear_index_col0 = 0   # Linear index before column (linear index = linear_index_col0 + row)
           for col in 1:size(S, 2)
               @inbounds for i in nzrange(S, col)
                   row = rowval[i]
                   val = nzval[i]
                   A[linear_index_col0+row] = val
               end
               linear_index_col0 += num_rows
           end
           return A
       end
copyto_3! (generic function with 1 method)
  ```
</details>

> Last thing: I believe we can add the inbounds annotation, the ingredients for the two loops and the setindex! look safe.

I can break the `@inbounds` assumption:
```julia
A = zeros(n,n); S = sprand(n,n,d);
S[end,end] = 1.5
S.rowval[end] += 1 # Doing something stupid... (alternatively change S.colptr to assign some potentially random value)
copyto_1!(A,S); # BoundsError: attempt to access 1000×1000 Matrix{Float64} at index [1000001]
copyto_3!(A,S); # no error, can't this corrupt memory? E.g.
function test(A) # Deliberately looking at A[end+1]
    @inbounds begin
        a = A[length(A)+1]
    end
    return a
end
julia> A[end]
0.0
julia> test(A)
1.5
```

Maybe I understand [it](https://docs.julialang.org/en/v1/base/base/#Base.@inbounds) wrong, but I thought that without checking for such issues, I shouldn't use  `@inbounds`?

dkarrasch:
Ok, if it doesn't help much then leaving it is fine, I think. OTOH, if someone messes with the internal representation after construction, then I'm not sure we can guarantee anything. I wonder if that would also break other code.

PR #329, opened 2023-01-18
Replace AbstractTriangular by UpperOrLowerTriangular

dkarrasch:
This will allow to push https://github.com/JuliaLang/julia/pull/26307 across the finish line. I don't think the intention here was really `AbstractTriangular`, but that the intention was to refer to the LinAlg types, which is exactly what `UpperOrLowerTriangular` is referring to.


dkarrasch:
Any objections? Does anybody know about custom `AbstractTriangular` types that might be affected here?

PR #337, opened 2023-01-28
Set version to 1.10.0

IanButterworth:
As noted in https://github.com/JuliaLang/julia/pull/48438 the version field is unset and should match julia's


ViralBShah:
Thanks.

PR #340, opened 2023-02-03
float(S) and complex(S) should not copy data unnecessarily

stevengj:
I noticed that `float(S)` and `complex(S)` always copy the data, including the index arrays.   I don't think this should be the default, since these methods in `Base` share storage if possible:
```jl
julia> x = rand(ComplexF64, 3); float(x) === x
true

julia> x = rand(ComplexF64, 3); complex(x) === x
true
```
If the caller needs to make a copy of the array, they can do `float.(S)` or `copy(float(S))`.

(It was also inconsistent, because `float` of a sparse *vector* did *not* make a copy of the nonzeros if they were already of a floating-point type, e.g. `ComplexF64`, but *did* copy the indices.)

### DIFF
--- a/src/sparsematrix.jl
+++ b/src/sparsematrix.jl
@@ -970,8 +970,8 @@ convert(T::Type{<:LowerTriangular}, m::AbstractSparseMatrixCSC) = m isa T ? m :
 convert(T::Type{<:UpperTriangular}, m::AbstractSparseMatrixCSC) = m isa T ? m :
     istriu(m) ? T(m) : throw(ArgumentError("matrix cannot be represented as UpperTriangular"))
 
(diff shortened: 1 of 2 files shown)


stevengj:
CI failures (`UndefVarError: UpperOrLowerTriangular not defined`) seem unrelated — I guess the current master of SparseArrays only works on Julia nightly?

dkarrasch:
We should perhaps stop testing on v1.8 on `main`. The current `main` of SparseArrays is indeed only supposed to work on Julia nightly. We should test on v1.8 and v1.9 on the respective release and/or backport branches. Long story short: CI is good for this PR.

dkarrasch (approved):

PR #342, opened 2023-02-06
Resolve `kron` ambiguities

dkarrasch:
This is required to merge the recent `kron` changes into Julia Base. Includes now ambiguity testing of `LinearAlgebra`, after loading of `SparseArrays`.

### DIFF
--- a/src/linalg.jl
+++ b/src/linalg.jl
@@ -1380,11 +1380,25 @@ end
     end
     return z
 end
-# due to the sparse result type, there is no risk to override dense ⊗ dense here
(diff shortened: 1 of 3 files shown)


ViralBShah:
This change is unfortunately a bit fussy, but looks like it should be fine to merge.

dkarrasch:
Yes, indeed. The issue is we don't have ambiguity testing working at all! The current test is `@test_nowarn detect_ambiguities(...)`, so `detect_ambiguities` happily returns a list of ambiguities and nobody ever gets to know (because it doesn't throw a warning if it finds ambiguities, even within SparseArrays!). Turning on the old-style ambiguity testing is currently also no option.

Issue #350, opened 2023-02-23
map does not respect sources index type


dpinol:
This currently does not hold
```
    v1 = spzeros(Float32, Int16, 10)
    v2 = spzeros(Float32, Int32, 10)
    @test eltype(typeof(SparseArrays.nonzeroinds(map(max, v1, v2)))) == Int32
```
I'll submit a PR

PR #351, opened 2023-02-23
Fix "map does not respect sources index type" (#350)

dpinol:


SobhanMP:
that's a nice catch!

src/sparsevector.jl
```diff
@@ -1468,7 +1469,7 @@ function _binarymap_mode_1!(f::Function, mx::Int, my::Int,
     return ir
 end
 
-function _binarymap_mode_2!(f::Function, mx::Int, my::Int,
+function _binarymap_mode_2!(f::Function, mx::Integer, my::Integer,
```
SobhanMP, line 1472: but now `ix` will not have the same type as `mx`, i think the safe thing is just make every function take Int and then store the "right" type. `Int` seems to be fairly standard for indexing in julia IMHO.

SobhanMP: My suggestion would be to undo all changes after line 1395, unless we test for type stability.

dpinol: Ok, I undid changes after line 1395 and rebased my commit to keep history clean

Wimmerer (approved):

PR #355, opened 2023-03-04
Test code quality, resolve ambiguities

dkarrasch:
This resolves two ambiguities: comparison of `ReadOnly`s and adjoint vectors times AdjointQ's. This also adds code quality tests via `Aqua.jl`. The new ambiguity test is currently turned off, because it reports the `ReadOnly` ambiguity already existing in the sysimg. After this PR is merged and the stdlib bumped in Julia, the ambiguity test can be turned on. Some other, smaller changes were necessary to help the other code quality checks, like unbound type arguments.

### DIFF
--- a/src/readonly.jl
+++ b/src/readonly.jl
@@ -35,5 +35,8 @@ Base.copy(x::ReadOnly) = ReadOnly(copy(parent(x)))
 (==)(x::ReadOnly, y::AbstractVector) = parent(x) == y
 (==)(x::AbstractVector, y::ReadOnly) = x == parent(y)
 (==)(x::ReadOnly, y::ReadOnly) = parent(x) == parent(y)
+# disambiguation
+(==)(x::ReadOnly{T,1,<:AbstractVector{T}}, y::ReadOnly{S,1,<:AbstractVector{S}}) where {T,S} =
+    parent(x) == parent(y)
 
 Base.dataids(::ReadOnly) = tuple()
--- a/src/readonly.jl
+++ b/src/readonly.jl
@@ -35,8 +35,5 @@ Base.copy(x::ReadOnly) = ReadOnly(copy(parent(x)))
 (==)(x::ReadOnly, y::AbstractVector) = parent(x) == y
 (==)(x::AbstractVector, y::ReadOnly) = x == parent(y)
 (==)(x::ReadOnly, y::ReadOnly) = parent(x) == parent(y)
-# disambiguation
-(==)(x::ReadOnly{T,1,<:AbstractVector{T}}, y::ReadOnly{S,1,<:AbstractVector{S}}) where {T,S} =
-    parent(x) == parent(y)
 
 Base.dataids(::ReadOnly) = tuple()
--- a/src/sparsevector.jl
+++ b/src/sparsevector.jl
@@ -88,12 +88,20 @@ _unsafe_unfix(s::FixedSparseVector) = SparseVector(length(s), parent(nonzeroinds
 const SparseColumnView{Tv,Ti}  = SubArray{Tv,1,<:AbstractSparseMatrixCSC{Tv,Ti},Tuple{Base.Slice{Base.OneTo{Int}},Int},false}
 const SparseVectorView{Tv,Ti}  = SubArray{Tv,1,<:AbstractSparseVector{Tv,Ti},Tuple{Base.Slice{Base.OneTo{Int}}},false}
 const SparseVectorUnion{Tv,Ti} = Union{AbstractCompressedVector{Tv,Ti}, SparseColumnView{Tv,Ti}, SparseVectorView{Tv,Ti}}
-const AdjOrTransSparseVectorUnion{Tv,Ti} = LinearAlgebra.AdjOrTrans{Tv, <:SparseVectorUnion{Tv,Ti}}
-const SVorFSV{Tv,Ti} = Union{SparseVector{Tv,Ti},FixedSparseVector{Tv,Ti}}
+const AdjOrTransSparseVectorUnion{Tv,Ti} = AdjOrTrans{Tv, <:SparseVectorUnion{Tv,Ti}}
+const SVorFSV{Tv,Ti} = Union{SparseVector{Tv,Ti},FixedSparseVector{Tv,Ti}} # TODO: remove?
+# untyped aliases, necessary for effective multiple dispatch
+const _SparseColumnView = SubArray{<:Any,1,<:AbstractSparseMatrixCSC,Tuple{Base.Slice{Base.OneTo{Int}},Int},false}
+const _SparseVectorView = SubArray{<:Any,1,<:AbstractSparseVector,Tuple{Base.Slice{Base.OneTo{Int}}},false}
+const _SparseVectorUnion = Union{AbstractCompressedVector, _SparseColumnView, _SparseVectorView}
+const _AdjOrTransSparseVectorUnion = AdjOrTrans{<:Any,<:SparseVectorUnion}
+
 ### Basic properties
 
-length(x::SVorFSV)   = getfield(x, :n)
-size(x::SVorFSV)     = (getfield(x, :n),)
+length(x::SparseVector) = getfield(x, :n)
+length(x::FixedSparseVector) = getfield(x, :n)
+size(x::SparseVector) = (getfield(x, :n),)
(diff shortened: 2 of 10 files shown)


ViralBShah:
Let's merge and bump if you are happy with these.

src/readonly.jl
```diff
@@ -35,5 +35,8 @@ Base.copy(x::ReadOnly) = ReadOnly(copy(parent(x)))
 (==)(x::ReadOnly, y::AbstractVector) = parent(x) == y
 (==)(x::AbstractVector, y::ReadOnly) = x == parent(y)
 (==)(x::ReadOnly, y::ReadOnly) = parent(x) == parent(y)
+# disambiguation
+(==)(x::ReadOnly{T,1,<:AbstractVector{T}}, y::ReadOnly{S,1,<:AbstractVector{S}}) where {T,S} =
```
SobhanMP, line 39: Why does this help? i really don't understand.

SobhanMP:
```suggestion
(==)(x::ReadOnly{T,N,<:AbstractVector{T}}, y::ReadOnly{S,N,<:AbstractVector{S}}) where {T,S,N} =
```

SobhanMP: or does it make it ambiguous ?

dkarrasch: I'm not 100% sure TBH. But looking at

```julia
julia> typeintersect(SparseArrays.ReadOnly, AbstractVector)
SparseArrays.ReadOnly{T, 1, V} where {T, V<:AbstractVector{T}}
```

I believe the ambiguity is real for `ReadOnly` of vectors? I think we should include it for now and then turn on ambiguity testing and try to remove this method and see what happens. But with the confusion of sysimg and external stdlib, it's not easy to tell.

src/sparsevector.jl
```diff
@@ -88,12 +88,20 @@ _unsafe_unfix(s::FixedSparseVector) = SparseVector(length(s), parent(nonzeroinds
 const SparseColumnView{Tv,Ti}  = SubArray{Tv,1,<:AbstractSparseMatrixCSC{Tv,Ti},Tuple{Base.Slice{Base.OneTo{Int}},Int},false}
 const SparseVectorView{Tv,Ti}  = SubArray{Tv,1,<:AbstractSparseVector{Tv,Ti},Tuple{Base.Slice{Base.OneTo{Int}}},false}
 const SparseVectorUnion{Tv,Ti} = Union{AbstractCompressedVector{Tv,Ti}, SparseColumnView{Tv,Ti}, SparseVectorView{Tv,Ti}}
-const AdjOrTransSparseVectorUnion{Tv,Ti} = LinearAlgebra.AdjOrTrans{Tv, <:SparseVectorUnion{Tv,Ti}}
-const SVorFSV{Tv,Ti} = Union{SparseVector{Tv,Ti},FixedSparseVector{Tv,Ti}}
+const AdjOrTransSparseVectorUnion{Tv,Ti} = AdjOrTrans{Tv, <:SparseVectorUnion{Tv,Ti}}
+const SVorFSV{Tv,Ti} = Union{SparseVector{Tv,Ti},FixedSparseVector{Tv,Ti}} # TODO: remove?
```
dkarrasch, line 92: Can this be removed?

src/sparsevector.jl
```diff
@@ -88,12 +88,20 @@ _unsafe_unfix(s::FixedSparseVector) = SparseVector(length(s), parent(nonzeroinds
...
+# untyped aliases, necessary for effective multiple dispatch
+const _SparseColumnView = SubArray{<:Any,1,<:AbstractSparseMatrixCSC,Tuple{Base.Slice{Base.OneTo{Int}},Int},false}
+const _SparseVectorView = SubArray{<:Any,1,<:AbstractSparseVector,Tuple{Base.Slice{Base.OneTo{Int}}},false}
+const _SparseVectorUnion = Union{AbstractCompressedVector, _SparseColumnView, _SparseVectorView}
+const _AdjOrTransSparseVectorUnion = AdjOrTrans{<:Any,<:SparseVectorUnion}
```
dkarrasch, line 97: Can these replace the above ones? In the code, we rarely dispatch on `Ti` and `Tv` in the method signature. And in the cases where we do, we could simply add lines like

```julia
Ti = indtype(x)
Tv  = eltype(x)
```

The "typed" aliases cause issues in multiple dispatch over large unions (as our `const`s), when the other (desired) candidate is untyped.

src/readonly.jl
SobhanMP: I see

dkarrasch: Ha, I removed it but now I remember the issue. There is an ambiguity reported between `(==)(x::ReadOnly, y::AbstractVector)` and `(==)(x::AbstractVector, y::ReadOnly)`, one taken from the sysimg, and the other one from the package. Without resolving this stupid ambiguity, we can't turn on `Aqua.jl`'s ambiguity testing. IIRC, then this ambiguity is also returned by `detect_ambiguities`, the current ambiguity test.

src/sparsevector.jl
SobhanMP: if you mean `SVorFSV`, it's handy five method use it but the type variables are useless? you could just replace it with abstractsparsevector if it doesn't break anything.

src/sparsevector.jl
SobhanMP: we don't seem to be using the type variables so it should be fine :thinking:

src/sparsevector.jl
dkarrasch: Yea, I already made it "obsolete" within SparseArrays.jl, but I wonder if this is used elsewhere.

src/sparsevector.jl
dkarrasch: Same here: is it used in other packages? Is there a way to search for, e.g., `SparseColumnView{` in the ecosystem?

src/sparsevector.jl
SobhanMP: It should not be...still experimental

src/sparsevector.jl
```diff
@@ -85,15 +85,17 @@ _unsafe_unfix(s::FixedSparseVector) = SparseVector(length(s), parent(nonzeroinds
...
 # Define an alias for a view of a whole column of a SparseMatrixCSC. Many methods can be written for the
 # union of such a view and a SparseVector so we define an alias for such a union as well
-const SparseColumnView{Tv,Ti}  = SubArray{Tv,1,<:AbstractSparseMatrixCSC{Tv,Ti},Tuple{Base.Slice{Base.OneTo{Int}},Int},false}
-const SparseVectorView{Tv,Ti}  = SubArray{Tv,1,<:AbstractSparseVector{Tv,Ti},Tuple{Base.Slice{Base.OneTo{Int}}},false}
-const SparseVectorUnion{Tv,Ti} = Union{AbstractCompressedVector{Tv,Ti}, SparseColumnView{Tv,Ti}, SparseVectorView{Tv,Ti}}
-const AdjOrTransSparseVectorUnion{Tv,Ti} = LinearAlgebra.AdjOrTrans{Tv, <:SparseVectorUnion{Tv,Ti}}
-const SVorFSV{Tv,Ti} = Union{SparseVector{Tv,Ti},FixedSparseVector{Tv,Ti}}
+const SparseColumnView = SubArray{<:Any,1,<:AbstractSparseMatrixCSC,Tuple{Base.Slice{Base.OneTo{Int}},Int},false}
+const SparseVectorView = SubArray{<:Any,1,<:AbstractSparseVector,Tuple{Base.Slice{Base.OneTo{Int}}},false}
+const SparseVectorUnion = Union{AbstractCompressedVector, SparseColumnView, SparseVectorView}
```
fredrikekre, line 90: Why were these changes needed? I know this is/should be internals but a lot of packages seem to use these. If we need new definitions, maybe use new names and keep the old ones?

dkarrasch: If there's a lot of packages that use it, I can revert these changes. The fact that the types were parameterized caused lots of pain in multiple dispatch internally. I should've run pkgeval on the bump PR. 😢

fredrikekre: But you can define
```
SparseVectorUnion2 = ...
```
and use that internally, just so the old names still exist for the "outside world" I mean.

dkarrasch: Yes, I'm already done with that locally. ;-)

PR #357, opened 2023-03-06
Install Aqua in the right place

DilumAluthge:


ViralBShah:
I would have thought that `runtests.jl` was the right place.

DilumAluthge:
Depends on the stdlib.

In SparseArrays, because `test/testgroups` exists, Base.runtests never actually runs `test/runtests.jl`. Instead, Base iterates over each $group in testgroups, and then does `include($group.jl)`. Different groups will usually get scheduled on different test workers.

ViralBShah (approved):

PR #361, opened 2023-03-10
Bring back typed aliases

dkarrasch:
I kept the diff minimal. Internally, we could replace all of these aliases by their untyped version, because the type parameters were previously (before #355) used only in 1 or 2 cases "for convenience", i.e., without effect on multiple dispatch, otherwise untyped.


fredrikekre (approved):
Great, thanks!

PR #362, opened 2023-03-10
Update the wrappers for SuiteSparse 7

ViralBShah:
Try to update the SuiteSparse wrappers for SuiteSparse 7 using Julia master.

cc @fxcoudert

### DIFF
--- a/gen/Manifest.toml
+++ b/gen/Manifest.toml
@@ -1,11 +1,12 @@
 # This file is machine-generated - editing it directly is not advised
 
-julia_version = "1.7.3"
+julia_version = "1.8.5"
 manifest_format = "2.0"
 project_hash = "61c6ad9feab9227ef7d89383b05c518b03764044"
 
 [[deps.ArgTools]]
 uuid = "0dad84c5-d112-42e6-8d28-ef12dabb789f"
+version = "1.1.1"
 
 [[deps.Artifacts]]
 uuid = "56f22d72-fd6d-98f1-02f0-08ddc0907c33"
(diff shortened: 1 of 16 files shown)


ViralBShah:
@Gnimuc Would the generator.jl file for generating the SuiteSparse wrappers need to be updated for the new Clang.jl? The script runs through with Julia master, but in the generated files, there are various C constants like `int64_t`, `LONG_MAX`, `__STDC_VERSION__` etc.

Help appreciated.

Gnimuc:
Please do not bump Clang.jl's minor version and the C lib's major version simultaneously. It's hard to figure out what causes the changes in the generated result.

Gnimuc:
@Wimmerer does SuiteSparse 7 introduce breaking changes? I mean we may need to upgrade those high-level wrappers as well.

Wimmerer:
Very possibly. I need to examine what we do with SuiteSparse_config

Wimmerer:
I will try to do that tonight

ViralBShah:
I believe @fxcoudert implement that change relating to the API change with the config strict in one of the recent PRs to this package.

There may be other changes.

ViralBShah:
> Please do not bump Clang.jl's minor version and the C lib's major version simultaneously. It's hard to figure out what causes the changes in the generated result.

Makes sense. I wasn't sure how to do this. We should close this PR.

Gnimuc:
I tested the generator locally. The generated results have no changes after bumping Clang.jl from v0.15 to v0.17.

Gnimuc:
But I failed to upgrade `SuiteSparse_jll` to v7.0.1:

```
(gen) pkg> add SuiteSparse_jll@v7.0.1
   Resolving package versions...
  No Changes to `~/.julia/dev/SparseArrays.jl/gen/Project.toml`
  No Changes to `~/.julia/dev/SparseArrays.jl/gen/Manifest.toml`

(gen) pkg> st
Status `~/.julia/dev/SparseArrays.jl/gen/Project.toml`
  [40e3b903] Clang v0.17.3
  [bea87d4a] SuiteSparse_jll v5.10.1+6
```

ViralBShah:
Yes, because it is an stdlib, it gets pinned. I find it easiest to disable it in the system image and then bump the version on Julia master. I have a local copy that I can provide as a branch.

ViralBShah:
@Gnimuc I have prepared https://github.com/JuliaLang/julia/pull/48977, which should make it easy to update these wrappers.

PR #363, opened 2023-03-12
Update the wrappers for SuiteSparse 7 and remove `cholmod_blas.h` bindings

Gnimuc:
~~It looks like our old API in use is stable in SuiteSparse 7.~~

cc @ViralBShah

### DIFF
--- a/gen/Manifest.toml
+++ b/gen/Manifest.toml
@@ -2,7 +2,7 @@
 
 julia_version = "1.10.0-DEV"
 manifest_format = "2.0"
-project_hash = "df2af1a593c4d7f0be9943645e66eabc322dac8b"
+project_hash = "61c6ad9feab9227ef7d89383b05c518b03764044"
 
 [[deps.ArgTools]]
 uuid = "0dad84c5-d112-42e6-8d28-ef12dabb789f"
--- a/gen/Manifest.toml
+++ b/gen/Manifest.toml
@@ -1,11 +1,12 @@
 # This file is machine-generated - editing it directly is not advised
 
-julia_version = "1.7.3"
+julia_version = "1.10.0-DEV"
 manifest_format = "2.0"
-project_hash = "61c6ad9feab9227ef7d89383b05c518b03764044"
+project_hash = "df2af1a593c4d7f0be9943645e66eabc322dac8b"
 
 [[deps.ArgTools]]
 uuid = "0dad84c5-d112-42e6-8d28-ef12dabb789f"
+version = "1.1.1"
 
 [[deps.Artifacts]]
 uuid = "56f22d72-fd6d-98f1-02f0-08ddc0907c33"
@@ -20,19 +21,15 @@ version = "0.4.2"
 
 [[deps.Clang]]
 deps = ["CEnum", "Clang_jll", "Downloads", "Pkg", "TOML"]
-git-tree-sha1 = "b6998b85ea2abbc8a02ca1ccca9438338ed5f8e5"
+git-tree-sha1 = "ac81f3ea7c53b20d64ad1609a0298d9731fbdcf8"
 uuid = "40e3b903-d033-50b4-a0cc-940c62c95e31"
-version = "0.15.8"
+version = "0.17.3"
 
 [[deps.Clang_jll]]
-deps = ["Artifacts", "JLLWrappers", "Libdl", "Pkg", "Zlib_jll", "libLLVM_jll"]
-git-tree-sha1 = "8cf7e67e264dedc5d321ec87e78525e958aea057"
+deps = ["Artifacts", "JLLWrappers", "Libdl", "Pkg", "TOML", "Zlib_jll", "libLLVM_jll"]
+git-tree-sha1 = "b88c99c9093f9db49a40d0715ea0e3ae5bbd91f7"
 uuid = "0ee61d77-7f21-5576-8119-9fcc46b10100"
-version = "12.0.1+3"
-
-[[deps.CompilerSupportLibraries_jll]]
-deps = ["Artifacts", "Libdl"]
-uuid = "e66e0078-7015-5450-92f7-15fbd957f2ae"
+version = "14.0.6+2"
 
 [[deps.Dates]]
 deps = ["Printf"]
@@ -41,6 +38,7 @@ uuid = "ade2ca70-3891-5945-98fb-dc099432e06a"
 [[deps.Downloads]]
 deps = ["ArgTools", "FileWatching", "LibCURL", "NetworkOptions"]
 uuid = "f43a241f-c20a-4ad4-852c-f6b1247861c6"
+version = "1.6.0"
 
 [[deps.FileWatching]]
 uuid = "7b1f6079-737a-58dc-b8bc-7a2ca5c1b5ee"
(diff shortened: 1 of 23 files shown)


ViralBShah:
There was a small API change that was implemented in https://github.com/JuliaSparse/SparseArrays.jl/commit/c09f0ac575165ddf4d264c15284d8fc0f8ac54b0

ViralBShah:
Actually seems like it is not picking up the new SuiteSparse 7: https://github.com/JuliaSparse/SparseArrays.jl/blob/6a10bdabc8ba42df2aa501b5861be3baef572bab/src/solvers/lib/aarch64-apple-darwin20.jl#L2171

ViralBShah:
I have updated `generator.jl` to work with this setup, but it generates wrappers that have things like `int64_t` and various other undefined things in them.

Gnimuc:
I still get
```
ERROR: LoadError: Cannot locate artifact 'SuiteSparse' in '/Users/gnimuc/Code/juliasrc/usr/share/julia/stdlib/v1.10/SuiteSparse_jll/src/../Artifacts.toml'
Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:35
 [2] ensure_artifact_installed(name::String, artifacts_toml::String; platform::Base.BinaryPlatforms.Platform, pkg_uuid::Nothing, verbose::Bool, quiet_download::Bool, io::Base.TTY)
   @ Pkg.Artifacts ~/Code/juliasrc/usr/share/julia/stdlib/v1.10/Pkg/src/Artifacts.jl:354
 [3] ensure_artifact_installed(name::String, artifacts_toml::String)
   @ Pkg.Artifacts ~/Code/juliasrc/usr/share/julia/stdlib/v1.10/Pkg/src/Artifacts.jl:346
 [4] top-level scope
   @ ~/.julia/dev/SparseArrays.jl/gen/generator.jl:12
in expression starting at /Users/gnimuc/.julia/dev/SparseArrays.jl/gen/generator.jl:12
```

We should doc how to develop and update a stdlib.

Gnimuc:
> I am not checking in the generated files, since they do not work - but they are essentially what I had in #362.

It should be fine. I just checked the src of SuiteSparse:

https://github.com/DrTimothyAldenDavis/SuiteSparse/blob/1d379153cd5f1ea5bc67d27dfaf5f018eb3038ad/CHOLMOD/Include/cholmod.h#L1017

these changes are intended.

ViralBShah:
Do you get that error using my PR as well? I really hope we can make SparseArrays.jl a standalone package and not an stdlib.

cc @KristofferC

ViralBShah:
@Gnimuc The problem with those generated files in #362 is that `int64_t` and some of those other things are not defined anywhere in the generated julia code. So they come up as undefined variables.

Gnimuc:
> @Gnimuc The problem with those generated files in #362 is that `int64_t` and some of those other things are not defined anywhere in the generated julia code. So they come up as undefined variables.

Sorry. I finally understood the problem in https://github.com/JuliaSparse/SparseArrays.jl/pull/362#issuecomment-1464236676 . I thought there were something wrong when Clang.jl mistakenly translated those `Int64` types.

Those "constants" are used in macros, so we need to either ignore the macro identifier or manually add those definitions like what we have done here: https://github.com/JuliaSparse/SparseArrays.jl/blob/main/src/solvers/LibSuiteSparse.jl#L11

Gnimuc:
To ignore a macro identifier, we can add it to this list: https://github.com/JuliaSparse/SparseArrays.jl/blob/main/gen/generator.toml#L9

For example, `__STDC_VERSION__` and `restrict` are being used in the definition of `SUITESPARSE_STDC_VERSION` and `SUITESPARSE_RESTRICT` respectively. 
```
const SUITESPARSE_STDC_VERSION = __STDC_VERSION__

const SUITESPARSE_RESTRICT = restrict
```

`SUITESPARSE_STDC_VERSION` and `SUITESPARSE_RESTRICT` should be added to the ignore list.

Gnimuc:
> Do you get that error using my PR as well?

Yes. This is the full version info:

```
julia> versioninfo()
Julia Version 1.10.0-DEV.810
Commit 841e9df121 (2023-03-11 15:48 UTC)
Platform Info:
  OS: macOS (x86_64-apple-darwin22.2.0)
  CPU: 8 × Apple M1
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-14.0.6 (ORCJIT, westmere)
  Threads: 4 on 8 virtual cores
Environment:
  JULIA_EDITOR = code
  JULIA_NUM_THREADS = 4
```

I remembered that in the old days, to develop a stdlib, one simple hack is to manually edit the UUID of that package. I believe this hack can not work with the latest master?

ViralBShah:
> I remembered that in the old days, to develop a stdlib, one simple hack is to manually edit the UUID of that package. I believe this hack can not work with the latest master?

Yes, that's why I pinged @KristofferC here. I'm hoping that https://github.com/JuliaLang/julia/pull/48979 will merge soon, which will make this a lot easier, and the next step is to remove SparseArrays from the stdlibs altogether.

fxcoudert:
Is there any news?

fxcoudert:
Warning:

```
┌ Warning: CHOLMOD version incompatibility
│ 
│ Julia was compiled with CHOLMOD version 4.0.3. It is
│ currently linked with version 3.0.14.
│ This might cause Julia to terminate when working with
│ sparse matrix factorizations, e.g. solving systems of
│ equations with \.
│ 
│ It is recommended that you use Julia with the same major
│ version of CHOLMOD as the one used during the build, or
│ download the generic binaries from www.julialang.org,
│ which ship with the correct versions of all dependencies.
└ @ SparseArrays.CHOLMOD D:\a\SparseArrays.jl\SparseArrays.jl\src\solvers\cholmod.jl:199
```

and failure on 32-bit Windows.

ViralBShah:
I wouldn't worry about the failures here, which are using the nightly built on julia master. If it works locally, I think we should merge it, and then bump on Julia master, with some tender love and care.

Gnimuc:
I should point out that I didn't test it locally.

ViralBShah:
I did run the SparseArrays tests on a build of Julia that pulls in SuiteSparse 7 and SparseArrays.jl from this PR. Everything works fine, except for a couple of QR factorization failures for `ordering=6`.

cc @Wimmerer

ViralBShah:
The missing ordering above is added by enabling PARTITION (METIS) support in SuiteSparse. https://github.com/JuliaPackaging/Yggdrasil/pull/6445.

ViralBShah:
@Gnimuc I think this doesn't treat the 32-bit OSes right.

https://buildkite.com/julialang/julia-master/builds/22268#018714ca-ee8c-4b4e-a2d6-9a1fa3c6c7b2

ViralBShah:
The way to test is to use julia branch with [the SuiteSparse 7 PR](https://github.com/JuliaLang/julia/pull/48977) and update `SparseArrays.version` in `stdlib` to be the commit SHA of the commit in your branch/PR on this repo.

Since SparseArrays is no longer in the system image, you don't have to necessarily build Julia everytime you make a change.

ViralBShah:
Trying in #372

jvo203:
As per the comment below, this problem does not seem to have been fixed in the new Julia 1.9.0 yet. Is there any reasonable time estimate as to when this problem might be fixed?

https://github.com/JuliaLang/julia/pull/48977#issuecomment-1547111853

PR #372, opened 2023-03-25
Deal with 32-bit SuiteSparse 7

ViralBShah:


Gnimuc:
A more proper fix would be to add [these compiler flags](https://github.com/JuliaPackaging/Yggdrasil/blob/99b076321ab5bbc87b8dfbb39ba33aaa3bb04da6/S/SuiteSparse/SuiteSparse%407/build_tarballs.jl#L29-L38) to https://github.com/JuliaSparse/SparseArrays.jl/blob/994bca1f8ff713f5cc55b7a23783806b1528dd13/gen/generator.jl#L39-L40

Gnimuc:
But it seems those flags only affects `SuiteSparse_long`, so the fix in this PR is also valid and stable.

ViralBShah:
Yes, it seems like we don't need `SuiteSparse_long` at all, and even those compiler flags are only relevant for SuiteSparse internally calling BLAS. They don't seem to make an appearance in the head files we process.

Let's see how if https://github.com/JuliaLang/julia/pull/48977 with the new builds works out.

PR #373, opened 2023-03-26
Continuing SuiteSparse 7 32-bit support

ViralBShah:


ViralBShah:
@Gnimuc I'm trying to use `ccall` rather than `@ccall` in case it gives some more information on 32-bit. The main issue is that I don't have a 32-bit linux setup to test.

ViralBShah:
@Gnimuc I think the wrappers are fine. 

`SuiteSparse_long` was discontinued in SuiteSparse 6. So we now need to use the `cholmod_l_` routines only on 64-bit architectures, but use `cholmod_` (non-l) routines on 32-bit architectures. This needs a bit of careful work.

cc @fxcoudert

PR #378, opened 2023-04-02
Use SuiteSparse_jll for LibSuiteSparse.jl

### LINKED ISSUE #349: Move to actually using JLLs
I believe we should be able to actually use the JLLs that ship with Julia as hinted at in this comment:

https://github.com/JuliaSparse/SparseArrays.jl/blob/38604c3b9f80059f108b196853d2ab9f9956c986/src/solvers/LibSuiteSparse.jl#L3-L6

This should standardize the library SONAME used by the main Julia distribution.

amontoison:
close #349


ViralBShah:
We'll need to hold this until the SuiteSparse 7 stuff is merged in (over the next few days).

PR #380, opened 2023-04-04
Allow Int32 indices, switch to SS7

Wimmerer:


Wimmerer:
I'm going to have to test this on a 32-bit VM to get anywhere on this

ViralBShah:
In case you don't already have one, the easiest thing is to get the ubuntu 32 bit docker image on a linux box.

ViralBShah:
doctests need updating: https://buildkite.com/julialang/julia-master/builds/22783#01877398-57ae-43e1-8707-50d39367ca4b

Wimmerer:
Okay that's good info, I'm not surprised QR is a bit worse, I made much more cursory changes there. I'll update. 

Will do doctests too.

Wimmerer:
The newest version of SPQR will introduce a small regression for us, we have to convert HPinv from Int32->Int64->Int32 when Ti === Int64. There are no `_l_` vs `_` methods for int32_t in SPQR as far as I can tell.

ViralBShah:
Is this breaking for the Julia user, or is it just that we need to adjust SPQR bindings to do some extra conversions so that tests pass?

Wimmerer:
Just extra conversions.

Wimmerer:
@ViralBShah this might have worked to switch to Int64 temporarily in SPQR. But this should *only* be used temporarily until Tim updates upstream. We should not release like this.

ViralBShah:
I suppose we should wait for SuiteSparse 7.1.

test/spqr.jl
```diff
@@ -9,8 +9,8 @@ using LinearAlgebra: I, istriu, norm, qr, rank, rmul!, lmul!, Adjoint, Transpose
 using SparseArrays: SparseArrays, sparse, sprandn, spzeros, SparseMatrixCSC
 using Random: seed!
 
-if Base.USE_GPL_LIBS
-
+# TODO REMOVE SECOND PREDICATE WITH SS7.1
+if Base.USE_GPL_LIBS && Int === Int64
```
ViralBShah, line 13: @Wimmerer Just a reminder - that this should be removed in order to test the new SuiteSparse 7.2 fork in Julia CI.

PR #382, opened 2023-04-11
Don't test on nightly, build docs on v1.9

dkarrasch:


ViralBShah:
We should re-enable on nightly once the in-flight PRs are merged.

PR #384, opened 2023-04-11
improve cat design

vtjnash:
This is type-piracy, but we cannot change that (until JuliaLang/julia#2326), so at least do not make these method intersections unnecessary slow and complicated for everyone who does not care about SparseArrays and does not load it, and unreliable for everyone who does load it.


vtjnash:
bump

ViralBShah:
Does this need to get merged and backported to 1.9?

cc @Wimmerer @dkarrasch

vtjnash:
It is for v1.10. We need to stop backporting stuff to v1.9 and release it

ViralBShah:
@KristofferC Are you able to review and merge this?

vtjnash:
bump

ViralBShah:
Seems like Julia doesn't like this: https://buildkite.com/julialang/julia-master/builds/25732#01893d2f-42d7-49b6-8641-699e5ccc7101/602-913

vtjnash:
There is a complementary PR for base to this that I have rebased locally, once this package is fixed and functional again I can push those

ViralBShah:
We need the PR to get this package merged into Base.

ViralBShah:
It won't revert cleanly.

PR #388, opened 2023-05-11
Clean up some multiplication code

dkarrasch:
This is a continuation of the effort of reducing `mul!` methods and thereby speed-up package loading time. The general approach is to have "a few" generic `mul!`s that catch specific cases, like triangular, banded, or matvec multiplication, and forward them to helper functions. If necessary, disambiguation can happen at this level, which should be easier because it doesn't compete with the entire `mul!` list. For symmetric sparse matrices I've actually added a few `mul!` methods so that hopefully method insertion is faster (it mirrors the LinearAlgebra methods and differs only in one type parameter, no unions involved).


dkarrasch:
This PR seems to reduce package load time by more than one half. Comparing load times from one month ago with this one, load time decreases from 0.7 seconds to 0.19 seconds on my machine.

PR #392, opened 2023-05-26
bring coverage of fixed SparseMatrixCSC to 100%

SobhanMP:


test/fixed.jl
```diff
@@ -69,7 +69,16 @@ struct_eq(A::AbstractSparseVector, B::AbstractSparseVector) =
     @test typeof(B) == typeof(F)
     @test struct_eq(B, F)
 end
+@testset "SparseMatrixCSC conversions" begin
+    A = sprandn(10, 10, 0.3)
+    F = fixed(copy(A))
+    B = SparseMatrixCSC(F)
+    @assert A == B
```
dkarrasch, line 76: Shouldn't this be

```suggestion
    @test A == B
```
? At least, I've never seen people testing with `@assert`, and I'm not sure this gets counted in code coverage.

SobhanMP: :laughing: i'm sorry, i think it gets counted but yeah, i shouldn't

dkarrasch:
Seems like there are still two untested methods:

```julia
SparseMatrixCSC(m, n, colptr::ReadOnly, rowval::ReadOnly, nzval::Vector)
FixedSparseCSC{Tv,Ti}(x::AbstractSparseMatrixCSC)
```

ViralBShah:
@SobhanMP Any chance we can add coverage for the other two methods? Would be nice to get this PR in time for 1.10.

SobhanMP:
@ViralBShah change in codecov seems unrelated, this should be 100%

Issue #265, opened 2022-09-22
AbstractSparseMatrixCSC not exported


j-fu:
Hi, I would like to use the AbstractSparseMatrixCSC interface with my sparse matrix struct, however  it seems that it is not exported. 

Also, it is not completely clear what needs to be implemented to conform to the interface. 
I currently assume it is `getcolptr`, `getrowval`, `nonzeros`, `nnz`, `size`.

Wimmerer:
That should be sufficient for most operations yes. There will definitely be some things missing, there are not many other subtypes of AbstractSparseMatrixCSC in the ecosystem to test it out.

You can qualify the name `SparseArrays.AbstractSparseMatrixCSC`, but I'll look into adding it to exports.

j-fu:
... bump...

ViralBShah:
Is it possible to provide a PR? We should also document the interface.

j-fu:
Ok will try. Can get on this in a couple of days.

j-fu:
What the `sparse(A)` method should do ?
Currently, we have
https://github.com/JuliaSparse/SparseArrays.jl/blob/0d2d30afb156b48d329ed212309b56a31c4a84a4/src/sparsematrix.jl#L1000 and
https://github.com/JuliaSparse/SparseArrays.jl/blob/0d2d30afb156b48d329ed212309b56a31c4a84a4/src/sparsematrix.jl#L729 (and following).

All of them copy the AbstractSparseMatrixCSC. IMHO all of them could return a SparseMatrixCSC via [this](https://github.com/JuliaSparse/SparseArrays.jl/blob/0d2d30afb156b48d329ed212309b56a31c4a84a4/src/sparsematrix.jl#L734) constructor,
and I think the wordl would be a better place if they would. 


So the question here is: is a `sparse(A::AbstractSparseMatrixCSC)::SparseMatrixCSC` method to be seen as a part of the AbstractSparseMatrixCSC interface ?

ViralBShah:
If you have your own concrete implementation of AbstractSparseMatrixCSC, then `sparse` should give you a copy of  the matrix in that concrete type, I believe. That way, I believe, your type will propagate as much as possible through the implementation.

j-fu:
Ok, that may be the philosophy behind `sparse`. But then why currently some of the constructors of `SparseMatricCSC` from `AbstractSparseMatrixCSC` do copy too ?

ViralBShah:
[...]

PR #395, opened 2023-06-01
Export AbstractSparseMatrixCSC + interface

j-fu:
See discussion in https://github.com/JuliaSparse/SparseArrays.jl/issues/265


src/sparsematrix.jl
```diff
@@ -725,11 +749,13 @@ function Base.sizehint!(S::SparseMatrixCSC, n::Integer)
 end
 
 # converting between SparseMatrixCSC types
-SparseMatrixCSC(S::AbstractSparseMatrixCSC) = copy(S)
+
 AbstractMatrix{Tv}(A::AbstractSparseMatrixCSC) where {Tv} = SparseMatrixCSC{Tv}(A)
-SparseMatrixCSC{Tv}(S::AbstractSparseMatrixCSC{Tv}) where {Tv} = copy(S)
+SparseMatrixCSC{Tv}(S::AbstractSparseMatrixCSC) where {Tv} = SparseMatrixCSC{Tv}(S)
 SparseMatrixCSC{Tv}(S::AbstractSparseMatrixCSC) where {Tv} = SparseMatrixCSC{Tv,eltype(getcolptr(S))}(S)
-SparseMatrixCSC{Tv,Ti}(S::AbstractSparseMatrixCSC{Tv,Ti}) where {Tv,Ti} = copy(S)
```
SobhanMP, line 732: isn't this function needed?

j-fu: I am not so sure.  I even think it is a bug - if some subtype `T<:AbstractSparseMatrixCSC` implements `copy(m::T)::T`, then the we would have for `typeof(m)==T`:

`typeof(SparseMatrixCSC(m))==T`. 

Just verified this under 1.10...

j-fu: IMHO  `SparseMatrixCSC(::AbstractSparseMatrixCSC)` should always return a `SparseMatrixCSC` .

j-fu:
Similar question for `sparse(m::AbstractSparseMatrixCSC)`. In 1.10 it returns a `copy(m)`. Which may be reasonable. One could argue as well (I tend to prefer this) that the return value should be a `SparseMatrixCSC` like with the other methods of `sparse`.

src/sparsematrix.jl
ViralBShah: That seems reasonable.

ViralBShah:
Generally, the thought with `sparse` has been that it shouldn't be seen as SparseMatrixCSC specific.

src/sparsematrix.jl
j-fu: If we consider the current implementation as a bug to be fixed, we might define

`SparseMatrixCSC(A::AbstractSparseMatrixCSC)=SparseMatrixCSC(size(A)..., getcolptr(A), rowvals(A), nonzeros(A))`, in particular  as `sparse` is not supposed to return a `SparseMatrixCSC`.

The alternative is to keep this undefined, and assume that in order to have this one should write code like [this](https://github.com/SciML/LinearSolve.jl/blob/270b56).

ViralBShah:
My concern with exporting was that wouldn't we better off with something more general? AbstractSparseMatrixCSC feels too specialized a thing as an abstract type. But of course, it isn't a huge change per se.

@rayegun Any thoughts?

src/sparsematrix.jl
SobhanMP: Seems reasonable!

ViralBShah:
Pinging @KristofferC and @fredrikekre on any thoughts here. Another set of eyes would be good here. While I don't see a major issue, a public API would be a big commitment going forward, and AbstractSparseMatrixCSC feels like a bit of a specialized abstract class.

j-fu:
... I'd like to come back to this - Something more general would of course be appreciated, as e.g. discussed in #538 . OTOH, as a matter of fact the interface is used in LinearSolve.jl, see e.g. [here](https://github.com/SciML/LinearSolve.jl/blob/43fc8d37ba039e18e11f4ae83fa8962d440ff908/src/LinearSolve.jl#L13) and [here](https://github.com/SciML/LinearSolve.jl/blob/43fc8d37ba039e18e11f4ae83fa8962d440ff908/src/factorization.jl#L113).
So IMHO it would be good to have this fixed as public API. May be instead of exporting it could be made `public` now.

j-fu:
@ChrisRackauckas, @oscardssmith, what do you think about this ?

ChrisRackauckas:
I think we need to write down what that expected interface is if you're going to consider it public. Like https://docs.julialang.org/en/v1/manual/interfaces/ . Can you try doing a first draft of it?

Maybe @mbauman and @timholy would be good to have on-board as well since they drafted the original array interface and I know they have comments on what a sparse array interface would need to have.

j-fu:
I started a discussion in #559. Not sure if this is the right place. See also #538.

timholy:
Do `Diagonal`, `Bidiagonal`, `Tridiagonal`, and `SymTridiagonal` all support this interface? They're examples of sparse matrices even if they are not subtypes of `AbstractSparseMatrix`. Would a CSR matrix be expected to support `getcolptr`? What would that even mean?

The deepest thinking I've done on a real interface for sparse arrays of all kinds is in https://github.com/timholy/ArrayIteration.jl (unfinished and effectively abandoned for lack of time). That interface is not so tightly coupled to a particular internal representation.

j-fu:
... Certainly, as CSR matrix should have getrowptr and not getcolptr... 
Very interesting discussion in ArrayIterators.jl !

I see two strands of discussion here (as I tried to outline in #559): 
- "what is needed to write algorithms on sparse matrices even if I don't know how they are stored?" -- which needs time (which I might not have as well...) and dedication
- Low level interfaces for coupling to all kinds of (often binary) packages which rely upon the concrete CSC or CSR structure. In this context, the AbstractSparseMatrixCSC interface (getcolptr etc. ) is currently used as a matter of fact though it is not exported or documented as such. 

IMHO, both levels of interfaces should (and can) coexist.

j-fu:
I guess mentioning "interface" in the title of this PR triggered a rather general discussion about what is the right "interface" in the sense of ["interfaces"](https://docs.julialang.org/en/v1/manual/interfaces/). Due to lacking time and resources (including  mine...), this  discussion is hard finish. Another reason for this IMHO is the lack of formal description tools for interfaces, and we might have to wait for their  emergence.

However my main concern is much more pragmatic: I just think that `AbstractSparseMatrixCSC` and `getcolptr` should be exported or made public.
 So in this context, the title of this PR should have been "Export AbstractSparseMatrixCSC + getcolptr" and avoid mentioning "interface". 

Once we have "public" now, I foresee, that the next generation of tools like Aqua will flag many usages of AbstractSparseMatrixCSC and getcolptr e.g. in LinearSolve.jl and the like. Also, in order to give LinearSolve the full power, packages defining preconditioners etc. also might be made aware of AbstractSparseMatrixCSC. I am considering to do corresponding PRs to preconditioning packages, but it feels a bit awkward to propose to use some unexported type and method.
 

To continue, I see two ways: rename this PR, or open another one which just adds the export statements and some documentation, without mentioning "interface".

j-fu:
I removed any method overwrites, it is just export and docs now.

PR #396, opened 2023-06-11
use unwrapping mechanism for triangular matrices

dkarrasch:
This is to make use of https://github.com/JuliaLang/julia/pull/50058 after that one is merged.


dkarrasch:
Finally, finally! This not only contains modifications due to the unwrapping mechanism, but also a gentle rewiring of type unions. This was carefully tested with profiling in view of short load times.

PR #397, opened 2023-06-13
Add `sparse!(I, J, V)` and `spzeros!(I, J)`.

fredrikekre:
It is quite common when building sparse matrices that you don’t care about the COO vectors after construction. This patch adds `SparseArrays.sparse!(I, J, V)` and `SparseArrays.spzeros!(I, J)` that reuses the memory for the matrix buffers. This upstreams https://github.com/Ferrite-FEM/Ferrite.jl/blob/2fc40dbb2052b2b12e94512fbb6f139c2de47b8a/src/arrayutils.jl#L90

### DIFF
--- a/src/sparsematrix.jl
+++ b/src/sparsematrix.jl
@@ -1270,6 +1270,34 @@ function sparse!(I::AbstractVector{Ti}, J::AbstractVector{Ti},
             Vector{Ti}(undef, n+1), Vector{Ti}(), Vector{Tv}())
 end
 
+"""
+    SparseArrays.sparse!(I, J, V, [m, n, combine]) -> SparseMatrixCSC
+
+Variant of `sparse!` that re-uses the input vectors (`I`, `J`, `V`) for the final matrix
+storage. After construction the input vectors will alias the matrix buffers; `S.colptr ===
+I`, `S.rowval === J`, and `S.nzval === V` holds, and they will be `resize!`d as necessary.
+
+Note that some work buffers will still be allocated. Specifically, this method is a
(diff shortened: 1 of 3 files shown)


src/sparsematrix.jl
```diff
@@ -2116,6 +2138,26 @@ function spzeros!(::Type{Tv}, I::AbstractVector{Ti}, J::AbstractVector{Ti}, m::I
                    csrrowptr, csrcolval, cscnzval, csccolptr, cscrowval, cscnzval)
 end
 
+"""
+    SparseArrays.spzeros!(::Type{Tv}, I, J, V, [m, n]) -> SparseMatrixCSC{Tv}
```
dkarrasch, line 2142:
```suggestion
    SparseArrays.spzeros!(::Type{Tv}, I, J, [m, n]) -> SparseMatrixCSC{Tv}
```

dkarrasch (approved):
LGTM. I wasn't aware of the possibility to create a sparse matrix from `I, J, V` and reuse them. And it's good to reduce argument load for users who don't need the "full" optimization of potentially reusing all storage vectors.

src/sparsematrix.jl
```diff
@@ -1270,6 +1270,28 @@ function sparse!(I::AbstractVector{Ti}, J::AbstractVector{Ti},
             Vector{Ti}(undef, n+1), Vector{Ti}(), Vector{Tv}())
 end
 
+"""
+    SparseArrays.sparse!(I, J, V, [m, n, combine])
+
+Variant of `sparse!` that re-uses the input vectors `I`, `J`, `V` for the matrix storage.
```
KristofferC, line 1276: Maybe it is worth pointing out that some temporary arrays will still be allocated.

PR #398, opened 2023-06-19
Specialize 3-arg `dot` for sparse self-adjoint matrices

### LINKED ISSUE #83: dot with Symmetric view of sparse matrix is very slow

dkarrasch:
Fixes #83.

I only later realized that this might be resolved by the iternz iterator. But, for the time being, I suggest we fix this by explicit code, and condense later when the nz-iterator becomes available.


dkarrasch:
Can this be merged?

ViralBShah:
Should we hold back things until we get the SuiteSparse 7 PR merged, which should be soon, and get CI working reliably again?

ViralBShah:
Yes, I think it makes sense to merge for inclusion in 1.10.

PR #405, opened 2023-07-10
Intermediate test relaxation

dkarrasch:
This is an intermediate change to make https://github.com/JuliaLang/julia/pull/50058 pass CI. So we need to first merge this here, then bump the stdlib, then merge https://github.com/JuliaLang/julia/pull/50058 (and backport to v1.10), then this test relaxation can be reverted (but doesn't have to), then #396 should be merged, and again bump the stdlib including backport to v1.10... roughly.


dkarrasch:
This is the next one, then we should bump the stdlib.

PR #406, opened 2023-07-10
Revert new hvcat design temporarily

ViralBShah:
Revert #384

PR #407, opened 2023-07-10
Reland new hvcat design (#406)

vtjnash:
Opening the reland commit as a followup after https://github.com/JuliaLang/julia/pull/48977 is finished


ViralBShah:
We should have new nightlies in a couple of days and be able to rerun these tests.

dkarrasch:
We need to be careful about what we merge into `main` now. Because everything we merge now, will get backported to v1.10. So, yes, we definitely need to bump SparseArrays.

ViralBShah:
Ok, let's not merge any new features for now - and stabilize what we have including hvcat.

dkarrasch:
Well, I do think we should include #392, possibly #395, #396, #398, and #405 as a prerequisite to #396. They have been waiting while we were fixing the SuiteSparse stuff.

PR #411, opened 2023-07-12
fully qualify promote_eltypeof call

vtjnash:
Missed in https://github.com/JuliaSparse/SparseArrays.jl/pull/407, but causing doctest build to fail when we try to use this.


ViralBShah:
Please merge if ready.

PR #416, opened 2023-07-21
sparse vector views remain sparse

### LINKED ISSUE #402: Subarrays of sparse should operate like sparse
```julia
M = spzeros(4,4)

julia>  @view(M[:,i]) + @view(M[:,k])
4-element Vector{Float64}:
 0.0
 0.0
 0.0
 0.0

julia>  (M[:,i]) + (M[:,k])
4-element SparseVector{Float64, Int64} with 0 stored entries

```

matbesancon:
Closes #402 for sparse vectors. Closes #225. Fixes https://github.com/JuliaLang/julia/issues/46355.

I also relaxed the assumption that the view indexing must be a Base.OneTo, otherwise this is too restrictive for vectors

### DIFF
--- a/src/sparsevector.jl
+++ b/src/sparsevector.jl
@@ -1567,7 +1567,7 @@ for fun in (:+, :-)
         nzinds = nonzeroinds(y)
         nzvals = nonzeros(y)
         @inbounds for nzidx in eachindex(nzinds)
-            res[nzinds[nzidx]] += nzvals[nzidx]
+            res[nzinds[nzidx]] = fun(res[nzinds[nzidx]], nzvals[nzidx])
         end
         return res
     end
--- a/src/sparsevector.jl
+++ b/src/sparsevector.jl
@@ -91,7 +91,7 @@ const _SparseVectorUnion = Union{AbstractCompressedVector, _SparseColumnView, _S
 const _AdjOrTransSparseVectorUnion = AdjOrTrans{<:Any,<:_SparseVectorUnion}
 # the following aliases are unused internally, but widespread in packages
 const SparseColumnView{Tv,Ti}  = SubArray{Tv,1,<:AbstractSparseMatrixCSC{Tv,Ti},Tuple{Base.Slice{Base.OneTo{Int}},Int},false}
-const SparseVectorView{Tv,Ti}  = SubArray{Tv,1,<:AbstractSparseVector{Tv,Ti},Tuple{AbstractUnitRange},false}
+const SparseVectorView{Tv,Ti}  = SubArray{Tv,1,<:AbstractSparseVector{Tv,Ti},<:Tuple{AbstractUnitRange},false}
 const SparseVectorUnion{Tv,Ti} = Union{AbstractCompressedVector{Tv,Ti}, SparseColumnView{Tv,Ti}, SparseVectorView{Tv,Ti}}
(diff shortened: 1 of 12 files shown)


matbesancon:
Alright this is good for review on my side

matbesancon:
@ViralBShah this could be a step towards fixing the views: https://github.com/JuliaSparse/SparseArrays.jl/pull/225#issuecomment-1218070104

src/sparsevector.jl
```diff
@@ -1558,6 +1596,22 @@ for (fun, mode) in [(:+, 1), (:-, 1), (:*, 0), (:min, 2), (:max, 2)]
     end
 end
 
+for fun in (:+, :-)
+    @eval @propagate_inbounds function $(fun)(x::Union{SparseVectorUnion{Tx},SparseVectorPartialView{Tx}}, y::Union{SparseVectorUnion{Ty},SparseVectorPartialView{Ty}}) where {Tx, Ty}
+        @boundscheck size(x) == size(y)
```
fredrikekre, line 1601: This doesn't throw on mismatching size, and I think what you should check is `axes`.

fredrikekre: Still doesn't throw, you need something like `axes(x) == axes(y) || throw(...)`.

matbesancon:
still haven't figured out what is causing the issue here

src/sparsevector.jl
```diff
@@ -114,6 +117,9 @@ function nnz(x::SparseColumnView)
     return length(nzrange(parent(x), colidx))
 end
 nnz(x::SparseVectorView) = nnz(x.parent)
+function nnz(x::SparseVectorPartialView)
```
SobhanMP, line 120: some bikeshedding :P but write this as a one liner IMHO

src/sparsevector.jl
```diff
@@ -135,6 +141,12 @@ function nonzeros(x::SparseColumnView)
 end
 nonzeros(x::SparseVectorView) = nonzeros(parent(x))
 
+function nonzeros(x::SparseVectorPartialView)
+    (first_idx, last_idx) = _partialview_end_indices(x)
+    nzvals = nonzeros(parent(x))
+    return @view(nzvals[first_idx:last_idx])
```
SobhanMP, line 147: I vaguely remember people saying that you should use the view function directly.

src/sparsevector.jl
```diff
@@ -145,10 +157,36 @@ function nonzeroinds(x::SparseColumnView)
 end
 nonzeroinds(x::SparseVectorView) = nonzeroinds(parent(x))
 
+# return the first and last nonzero indices of the parent that belong to the view
+# return end+1:end if no nonzero in the parent
+function _partialview_end_indices(x::SparseVectorPartialView)
+    p = parent(x)
+    nzinds = nonzeroinds(p)
+    if isempty(nzinds)
+        first_idx = nzinds[begin]
```
SobhanMP, line 166: I'm confused, shouldn't this be out of bounds? It's empty.

SobhanMP:
Could you also update the

src/sparsevector.jl
matbesancon: good catch, I added a test to cover the case too

src/sparsevector.jl
matbesancon: done, might be linked to compile time?

src/sparsevector.jl
matbesancon: done indeed

dkarrasch:
I copied over the tests from #225 and thereby found a little oversight. I'll push the easy fix right after, and then we can merge after tests pass.

PR #417, opened 2023-07-22
various cleanups, gitignore

matbesancon:
Added a gitignore with a Manifest.toml, which otherwise gets generated when running tests.
Made the include in the tests more robust with a joinpath

### DIFF
--- a/test/higherorderfns.jl
+++ b/test/higherorderfns.jl
@@ -10,7 +10,7 @@ using Test
 using SparseArrays
 using LinearAlgebra
 using Random
-include("forbidproperties.jl")
+include(joinpath(@__DIR__, "forbidproperties.jl"))
 function test_map_and_map!(A, alloc_tests)
     # --> test map entry point
     fA = Array(A)
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+Manifest.toml
--- a/test/forbidproperties.jl
+++ b/test/forbidproperties.jl
@@ -1,5 +1,5 @@
 # This file is a part of Julia. License is MIT: https://julialang.org/license
 
 using SparseArrays
-Base.getproperty(S::SparseMatrixCSC, ::Symbol) = error("use accessor function")
-Base.getproperty(S::SparseVector, ::Symbol) = error("use accessor function")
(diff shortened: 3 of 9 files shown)


fredrikekre:
Include is always relative the current file, so why are those changes needed,

matbesancon:
usually yes, but testing stdlibs is a bit of an odd case (except if I missed an obvious workflow). To be sure the currently developed SparseArrays is being tested, I include the test file directly to test explicitly `Main.SparseArrays` which I have loaded before, in which case I always need to `cd test` before

matbesancon:
@fredrikekre what's your opinion given the answer above?

test/higherorderfns.jl
```diff
@@ -10,7 +10,7 @@ using Test
 using SparseArrays
 using LinearAlgebra
 using Random
-include("forbidproperties.jl")
+include(joinpath(@__DIR__, "forbidproperties.jl"))
```
vtjnash, line 13: This seems unnecessary complication, since `include` is already relative to `@__DIR__`

matbesancon: see this: https://github.com/JuliaSparse/SparseArrays.jl/pull/417#issuecomment-1646678186

SobhanMP: The workflow already checks that the correct version is loaded tho. See https://github.com/JuliaSparse/SparseArrays.jl/blob/main/.ci/test_and_change_uuid.jl

matbesancon: by workflow I meant local workflow, not github actions

ViralBShah:
Should we merge?

vtjnash:
I don't think it is clear why this is an improvement. It isn't how any other package does this.

matbesancon:
for now it is not as any other package, since there is no good technique (to the best of my knowledge) to test locally

SobhanMP:
but changing the local hash works fine?

matbesancon:
> but changing the local hash works fine?

how do you do this?

matbesancon:
in the meantime, it seems it's a problem on my side so I replaced it with https://github.com/JuliaSparse/SparseArrays.jl/pull/426

PR #418, opened 2023-07-22
bump version

dkarrasch:
I think we have everything included for the v1.10 release, so it's time to bump the minor version number and keep SparseArrays.jl in-sync with Julia Base.


ViralBShah:
I thought we bump these to keep in sync with Julia itself.

PR #425, opened 2023-08-23
Test suite: activate a temp project if we need to install Aqua.jl during the test suite

DilumAluthge:
Ref https://github.com/JuliaLang/julia/issues/50055

### DIFF
--- a/test/ambiguous.jl
+++ b/test/ambiguous.jl
@@ -5,6 +5,7 @@
 original_depot_path = copy(Base.DEPOT_PATH)
 original_load_path = copy(Base.LOAD_PATH)
 original_env = copy(ENV)
+original_project = Base.active_project()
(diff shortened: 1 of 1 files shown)


Wimmerer (approved):

DilumAluthge:
@Wimmerer How do you handle backports here? Is it sufficient for me to add the label to the PR?

Wimmerer:
The last time we had backports was 1.8 I believe and Kristoffer handled them. I can look at his script and see if that will work here. We should standardize across all the external stdlibs

DilumAluthge:
We should document somewhere the process for doing backports on an external stdlib.

@IanButterworth You've been taking care of backports for Pkg.jl for the past few releases. Would you be able to write up some instructions?

PR #433, opened 2023-09-01
Correctly set zeros with `fill!(::SubArray)` and fix its return value

sunoru:
This PR includes two fixes:

## 1
It is a simple obvious fix to the problem when one tries to do something like the following

```julia
using SparseArrays, StaticArrays
const Vec2 = SVector{2, Float64}
A = rand(Vec2, 2, 2) |> sparse
fill!(view(A, :, :), zero(Vec2))
```

Without the fix, it throws

```
ERROR: MethodError: Cannot `convert` an object of type Int64 to an object of type SVector{2, Float64}

Closest candidates are:
  convert(::Type{T}, ::Factorization) where T<:AbstractArray
   @ LinearAlgebra /usr/share/julia/stdlib/v1.9/LinearAlgebra/src/factorization.jl:59
  convert(::Type{SVector{N, T}}, ::CartesianIndex{N}) where {N, T}
   @ StaticArrays ~/.julia/packages/StaticArrays/jA1zK/src/SVector.jl:13
  convert(::Type{SA}, ::Tuple) where SA<:StaticArray
   @ StaticArrays ~/.julia/packages/StaticArrays/jA1zK/src/convert.jl:179
  ...

Stacktrace:
 [1] setindex!(A::Vector{SVector{2, Float64}}, x::Int64, i1::Int64)
   @ Base ./array.jl:969
 [2] _spsetz_setindex!
   @ /usr/share/julia/stdlib/v1.9/SparseArrays/src/sparsematrix.jl:3033 [inlined]
 [3] fill!(V::SubArray{SVector{2, Float64}, 1, SparseMatrixCSC{SVector{2, Float64}, Int64}, Tuple{Base.Slice{Base.OneTo{Int64}}, Int64}, false}, x::SVector{2, Float64})
   @ SparseArrays /usr/share/julia/stdlib/v1.9/SparseArrays/src/sparsematrix.jl:3001
 [4] top-level scope
   @ REPL[61]:1
```

## 2

`fill!(A, ...)` should always return `A`, but without the fix, it returns `A.parent` for sub arrays of the sparse arrays, if zeros are being set.

### DIFF
--- a/src/sparsematrix.jl
+++ b/src/sparsematrix.jl
@@ -3179,7 +3179,8 @@ function Base.fill!(V::SubArray{Tv, <:Any, <:AbstractSparseMatrixCSC{Tv}, <:Tupl
     else
         _spsetnz_setindex!(A, convert(Tv, x), I, J)
     end
-    return _checkbuffers(A)
+    _checkbuffers(A)
+    V
 end
 """
 Helper method for immediately preceding fill! method. For all (i,j) such that i in I and
--- a/src/sparsematrix.jl
+++ b/src/sparsematrix.jl
@@ -3207,7 +3207,7 @@ function _spsetz_setindex!(A::AbstractSparseMatrixCSC,
                 kI > lengthI && break
                 entrykIrow = I[kI]
             else # entrykArow == entrykIrow
-                nonzeros(A)[kA] = 0
+                nonzeros(A)[kA] = zero(eltype(A))
                 kA += 1
                 kI += 1
                 (kA > coljAlastk || kI > lengthI) && break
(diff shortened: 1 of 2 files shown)


dkarrasch:
Do you think you could add a test for this case? Maybe there is some number type that is not convertible to from an integer `0`? Perhaps some unitful number type should work?

sunoru:
OK I will try to add the tests.

Actually I just found another issue related to the `fill!` function for sub arrays that it should always returns the given array (the first argument) instead of the parent of that array. Should I submit another PR or can I also fix it here since I think they are both small fixes?

sunoru:
There is no number type except `Irrational` that is not convertible to from an integer 0, but it also doesn't have a zero value.

The only common use case I can imagine is for `SVector`s, but for the test purpose I can just declare a custom struct.

dkarrasch:
> Actually I just found another issue related to the `fill!` function for sub arrays

Fix here, but for easier review, please submit first a commit with a test that currently fails, and then another commit with the fix.

PR #437, opened 2023-09-02
Hook into new factorization dispatch mechanisms

dkarrasch:
This is a companion PR to https://github.com/JuliaLang/julia/pull/51161 and https://github.com/JuliaLang/julia/pull/51220. It should catch `lu`, `qr` and `cholesky` calls for wrapped sparse matrices (or, more generally, anything whose `lucopy` is an `AbstractSparseMatrixCSC`).

Now also depends on https://github.com/JuliaLang/julia/pull/51245.


dkarrasch:
I tested locally, and the speed-up for larger wrapped sparse matrices is, not surprisingly, gigantic.

ViralBShah:
Are we ready to merge this?

PR #438, opened 2023-09-03
Code quality cleanup

dkarrasch:
This PR contains a few small edits that should be backported to v1.10.

* resolve a method ambiguity in a rare case (previously, we didn't check method ambiguity because there was no way to distinguish the methods in the sysimage from those same in this package)
* apparently, an update of Aqua.jl allows us to tighten some code quality checks


matbesancon (approved):

matbesancon:
good to merge?

PR #444, opened 2023-09-26
Adapt to Documenter v1

dkarrasch:
Let's see if we need to exclude parts of the docs from the now default strict behavior.


dkarrasch:
@mortenpi Is there a way to cross-reference docstrings from `LinearAlgebra` and `Base` (methods)? Is there a way to cross-reference to the "Random Numbers" section somewhere in the `Base` documentation? Perhaps an URL would work. But is that stable against new releases etc?

mortenpi:
Nope, and that's non-trivial: https://github.com/JuliaDocs/Documenter.jl/issues/688

dkarrasch:
Thank you very much for the quick response. Then I went in the right direction removing these references, many of which were somewhat overambitious (like `Float64`).

ViralBShah:
I would defer to @mortenpi - since I am not familiar with the Documenter 1.0 changes yet. The docs do build, and the PR does look ok.

mortenpi:
Oh.. so this is a standard library, right? Those links do work when it gets built as part of the Julia manual: https://docs.julialang.org/en/v1/stdlib/SparseArrays/ So I think it would be better to leave them in for now?

mortenpi:
I don't think there's an easy workaround here. I think the best thing we can do is to make them warnings here by adding `:cross_references` to `warnonly`. But some bad link then may still break the combined manual build.

ViralBShah:
I think that would be fine. At least it would allow us to catch more egregious problems sooner. Links are relatively easy to fix.

ViralBShah (approved):

matbesancon:
good to go?

matbesancon (approved):

PR #447, opened 2023-10-01
Add JL_UMFPACK_PIVOT_TOLERANCE to umfpack.jl

sjdaines:
Looks like this was just omitted for some reason

See UMFPACK manual, increasing from default value is possibly helpful for ill-conditioned matrices:

*Control [UMFPACK_PIVOT_TOLERANCE]: relative pivot tolerance for threshold partial pivoting with row interchanges. In any given column, an entry is numerically acceptable if its absolute value is greater than or equal to Control [UMFPACK_PIVOT_TOLERANCE] times the largest absolute value in the column. A value of 1.0 gives true partial pivoting. If less than or equal to zero, then any nonzero entry is numerically acceptable as a pivot. Default: 0.1.*

*Smaller values tend to lead to sparser LU factors, but the solution to the linear system can become inaccurate. Larger values can lead to a more accurate solution (but not always), and usually an increase in the total work.*


ViralBShah:
Either accidentally missed, or was introduced after the major work on this wrapper was done. Thank you.

Issue #448, opened 2023-10-01
Document usage of UMFPACK control vector


sjdaines:
The documentation for UMFPACK `lu` could use:
- an explicit note/warning that Julia disables iterative refinement, which means that the Julia default is not the same as the defaults given in the UMFPACK manual.  This is an intentional change,  https://github.com/JuliaSparse/SparseArrays.jl/issues/122, but isn't documented
- an example of how to set the control vector and display it

eg 

    umfpack_control = SparseArrays.UMFPACK.get_umfpack_control(Float64, Int64) # read defaults for a Float64 sparse matrix
    SparseArrays.UMFPACK.show_umf_ctrl(umfpack_control) # display values
    umfpack_control[SparseArrays.UMFPACK.JL_UMFPACK_IRSTEP] = 2.0 # reenable iterative refinement (2 is UMFPACK default max iterative refinement steps, Julia modifies this default to disable iterative refinement)

PR #449, opened 2023-10-01
Add example for UMFPACK control vector

sjdaines:
Fixes https://github.com/JuliaSparse/SparseArrays.jl/issues/448

Add a note to the documentation for `lu`, explaining that Julia modifies the UMFPACK defaults to disable iterative refinement.

Add an example to illustrate usage of the UMFPACK control vector to reenable iterative refinement


ViralBShah:
@mortenpi Would you know what we can do to fix the failing doctests?

PR #463, opened 2023-10-19
Backports to Julia v1.10, part 2

dkarrasch:


dkarrasch:
What's the issue? Is SuiteSparse v7.2.1 not released yet?

```
possible versions are: 7.2.0 or uninstalled
```

KristofferC:
I think the reason is that on 1.10 the version of SuiteSparse is 7.2.0. It is not clear to me why the compat here need to be bumped though.

dkarrasch:
Wait, there was another `Manifest.toml` in the `gen` folder, that had a v7.2.0 hardcoded. Does that need bumping?

Generally, I don't know. That's how I understood @ViralBShah, that something needs to be backported to the release branch, without "bumping SparseArrays". But you're right, the compat entry is open to nonbreaking new releases...

ViralBShah:
What I meant was that SparseArrays master has a lot more on it than what we need in 1.10. But the whole SuiteSparse dependency is locked down (probably more than it needs to). There is no API change between 7.2 and 7.2.1, but it is possible wrappers need regenerating.

ViralBShah:
We should probably remove the Manifest.toml in the gen folder.

dkarrasch:
> The julia compat in the main SparseArrays.jl Project.toml on backports should become 1.10 (which is 1.11 on master).

It is already: https://github.com/JuliaSparse/SparseArrays.jl/blob/d8aae6daa0943d719b4f6dfb0e906193b65928fa/Project.toml#L14

So, do we need to do anything or can we just close this? Will SparseArrays simply pull SuiteSparse v7.2.1 by the compat?

ViralBShah:
@KristofferC How should we proceed?

KristofferC:
If we update e.g. the Suitesparse jll on 1.10 SparseArrays.jl should use that.

ViralBShah:
Ok - the SuiteSparse_jll bump will have to happen on the Julia stdlib side as part of the backport of my PR. So this one can just be merged here.

KristofferC:
Is the thinking here that bumping the compat here is what upgrades SuiteSparse? If so, that is not the case. I don't see what use this PR has (and now CI fails from it and will keep failing until there is a new 1.10 release).

ViralBShah:
Yes, it is to upgrade suitesparse to the release version.

ViralBShah:
The version of Julia used for the actions is older than the bump of SuiteSparse to 7.2.1. So, if we can update the Julia version in the actions here (perhaps RC1?), then this CI will start working.

PR #464, opened 2023-10-23
fix h/vcat invoke dispatch arguments

vtjnash:
Fixes https://github.com/JuliaLang/julia/issues/51826

(looks like hvcat was already correct, but somehow these got copied wrong from where ever they came from)


ViralBShah:
cc @KristofferC @dkarrasch

PR #468, opened 2023-11-08
test: relax allocated tests

vtjnash:
Several of these are failing being slightly over the threshold on some CI runs. Relax all the thresholds, since the intent is to make sure allocations are completely dynamic, not to prohibit them entirely.

C.f. https://buildkite.com/julialang/julia-master/builds/29821#018bab25-9f95-4abd-aabf-6eaa8c05859e for example

### DIFF
--- a/test/higherorderfns.jl
+++ b/test/higherorderfns.jl
@@ -62,11 +62,11 @@ end
         # --> test map! entry point
         fX = map(+, fA, fB); X = sparse(fX)
         map!(+, X, A, B); X = sparse(fX) # warmup for @allocated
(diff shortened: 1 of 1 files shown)


ViralBShah:
Should we bump SparseArrays on master to pull this in?

vtjnash:
eventually yes

vtjnash:
it can be done here https://github.com/JuliaLang/BumpStdlibs.jl/actions/workflows/BumpStdlibs.yml

ViralBShah:
Yeah, my question was whether this was urgent and if we should do it right away.

PR #471, opened 2023-11-15
Fix compat. Allow newer Aqua. Only run Aqua on repo tests.

IanButterworth:


test/runtests.jl
```diff
@@ -1,6 +1,10 @@
 # This file is a part of Julia. License is MIT: https://julialang.org/license
 using Test, LinearAlgebra, SparseArrays
 
+if Base.get_bool_env("SPARSEARRAYS_AQUA_TEST", false)
```
fredrikekre, line 4: Might be easier to just use
```suggestion
if get(ENV, "SPARSEARRAYS_AQUA_TEST", nothing) === "true"
```
since this isn't something people will configure in general (and thus not forget whether it is 1 or true).

fredrikekre (approved):

DilumAluthge:
I think it's fine to run Aqua.jl in Base Julia Buildkite CI, we just need to make sure we respect the `[compat]` entry when we do the `Pkg.add` in Buildkite. E.g. https://github.com/JuliaSparse/SparseArrays.jl/pull/472

IanButterworth:
We need to merge something. Can you make the call @DilumAluthge

DilumAluthge:
Let's do this for now.

DilumAluthge:
We can return to running Aqua.jl in Buildkite later. For now, since everything is failing on Buildkite Julia master, let's get this PR deployed.

DilumAluthge:
BumpStdlibs run: https://github.com/JuliaLang/BumpStdlibs.jl/actions/runs/6883759465

DilumAluthge:
BumpStdlibs PR: https://github.com/JuliaLang/julia/pull/52183

PR #472, opened 2023-11-15
CI: Respect `[compat]` entries for test-only deps when running in Base Julia's Buildkite CI; run the `ambiguous` testgroup in Buildkite CI; mark `Aqua.test_project_toml_formatting` as `@test_skip`

DilumAluthge:
Reverts much of https://github.com/JuliaSparse/SparseArrays.jl/pull/471

This PR introduces a framework for installing test deps during stdlib CI in a way that respects `[compat]` entries. We can likely use this same framework in other stdlib repos.


ViralBShah:
Anything further to be done here?

PR #476, opened 2023-11-23
Fix performance trap for sparse view multiplication

### LINKED ISSUE #475: Using `@view` leads to 100x performance loss
I have a dense-times-sparse matrix multiplication, but I only need some columns of the sparse matrix.
E.g., consider the following code:
```julia
D = rand(1000, 2000); X = sprand(2000, 10_000, 0.1);
w = findall(!iszero, X[1,:])
buf = zeros(size(D,1), length(w));

@benchmark buf.=D*X[:,w]
@benchmark buf.=D*@view X[:,w]
```
leads to `9ms` and `700ms`, respectively.

It took me a while to find this, as I had wrapped the code block this was originally in with a `@views` call. Any ideas where the slowdown comes from?

dkarrasch:
It turns out that for non-triangular sparse matrices, `getcolptr`, which can be easily computed from the parent array, is not used. But all the other `get*` functions don't require a unit range as the second slicing argumnet in column views, so the fast multiplication kernels also apply to that situation.

Fixes #475.


dkarrasch:
There's an opportunity here to bikeshed about names of internal aliases... just in case things can be improved for more clarity. I was not very creative.

src/sparsematrix.jl
```diff
@@ -181,13 +181,21 @@ const SparseMatrixCSCView{Tv,Ti} =
     SubArray{Tv,2,<:AbstractSparseMatrixCSC{Tv,Ti},
         Tuple{Base.Slice{Base.OneTo{Int}},I}} where {I<:AbstractUnitRange}
 const SparseMatrixCSCUnion{Tv,Ti} = Union{AbstractSparseMatrixCSC{Tv,Ti}, SparseMatrixCSCView{Tv,Ti}}
+# Define an alias for views of a SparseMatrixCSC which include all rows and a selection of the columns.
+# Also define a union of SparseMatrixCSC and this view since many methods can be defined efficiently for
+# this union by extracting the fields via the get function: getrowval, and getnzval, BUT NOT getcolptr!
+const SparseMatrixCSCView2{Tv,Ti} =
+    SubArray{Tv,2,<:AbstractSparseMatrixCSC{Tv,Ti},
+        Tuple{Base.Slice{Base.OneTo{Int}},I}} where {I<:AbstractVector{<:Integer}}
```
rayegun, line 189: Constructing `SubArrays` with `@view` this seems to always be the correct type. But is `Base.OneTo` defined to always represent the full dimension?

dkarrasch: It's `Slice` that indicates that an entire dimension was selected. The `OneTo` objects just stores the length of that dimension.

src/sparsematrix.jl
```diff
@@ -181,13 +181,21 @@ const SparseMatrixCSCView{Tv,Ti} =
     SubArray{Tv,2,<:AbstractSparseMatrixCSC{Tv,Ti},
         Tuple{Base.Slice{Base.OneTo{Int}},I}} where {I<:AbstractUnitRange}
 const SparseMatrixCSCUnion{Tv,Ti} = Union{AbstractSparseMatrixCSC{Tv,Ti}, SparseMatrixCSCView{Tv,Ti}}
+# Define an alias for views of a SparseMatrixCSC which include all rows and a selection of the columns.
+# Also define a union of SparseMatrixCSC and this view since many methods can be defined efficiently for
+# this union by extracting the fields via the get function: getrowval, and getnzval, BUT NOT getcolptr!
+const SparseMatrixCSCView2{Tv,Ti} =
```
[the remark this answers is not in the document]
dkarrasch: Probably anything is better than `SparseMatrixCSCView2`. 😄

PR #478, opened 2023-12-18
Update for SuiteSparse v7.4.0

### LINKED ISSUE #482: Wrapper generator needs updating to pick SuiteSparse headers from new location
As discussed in https://github.com/JuliaLang/julia/pull/52577#issuecomment-1874352276, the wrapper generator script needs updating to pick up the SuiteSparse headers from the new location (in stdlib) in the Julia distribtution.

@rayegun You may already have this in a PR - just making sure I note this here so it is not lost.

```
➜  gen git:(main) julia --project generator.jl
ERROR: LoadError: AssertionError: isfile(cholmod_h)
Stacktrace:
 [1] top-level scope
   @ ~/repos/SparseArrays.jl/gen/generator.jl:15
in expression starting at /Users/viral/repos/SparseArrays.jl/gen/generator.jl:15
```

giordano:
This goes with https://github.com/JuliaLang/julia/pull/52577.  All tests are passing, at least for aarch64-apple-darwin (the first platform to have finished running), but there are lots of warnings like
```
      From worker 10:	┌ Warning: CHOLMOD version incompatibility
      From worker 10:	│
      From worker 10:	│ Julia was compiled with CHOLMOD version 4.0.4. It is
      From worker 10:	│ currently linked with version 5.0.0.
      From worker 10:	│ This might cause Julia to terminate when working with
      From worker 10:	│ sparse matrix factorizations, e.g. solving systems of
      From worker 10:	│ equations with \.
      From worker 10:	│
      From worker 10:	│ It is recommended that you use Julia with the same major
      From worker 10:	│ version of CHOLMOD as the one used during the build, or
      From worker 10:	│ download the generic binaries from www.julialang.org,
      From worker 10:	│ which ship with the correct versions of all dependencies.
      From worker 10:	└ @ SparseArrays.CHOLMOD /Users/julia/.julia/scratchspaces/a66863c6-20e8-4ff4-8a62-49f30b1f605e/agent-cache/default-honeycrisp-R17H3W25T9.0/build/default-honeycrisp-R17H3W25T9-0/julialang/julia-master/julia-b3081f8082/share/julia/stdlib/v1.11/SparseArrays/src/solvers/cholmod.jl:206
```
My understanding is that we need to update these version numbers accordingly.  I have no idea whether we need to do anything else.

Close #480, fix #482.


giordano:
Note that now tests are are failing because I bumped the required version of SuiteSparse_jll, which isn't available in Julia yet, so that's expected, but tests were passing [on the previous commit](https://github.com/JuliaSparse/SparseArrays.jl/actions/runs/7254644657?pr=478) (note that macOS job was cancelled just because the job was taking longer than the other OSes, but tests were all successful to that point).

However I'd appreciate some help with figuring out this test failure in the julia repository, it only happens on 32-bit Windows: https://github.com/JuliaLang/julia/pull/52577#issuecomment-1862815917

ViralBShah:
I think we should at least regenerate the wrappers and see what happens.

giordano:
How? 🙂

giordano:
I included changes in #480, and am now running tests in https://github.com/JuliaLang/julia/pull/52577 based on this PR.  If tests are successful I suggest

* we merge this PR
* then update https://github.com/JuliaLang/julia/pull/52577 to use the `main` branch of this repository instead of this PR
* merge https://github.com/JuliaLang/julia/pull/52577
* profit

test/cholmod.jl
```diff
@@ -238,11 +238,12 @@ end
     @test_throws CHOLMOD.CHOLMODException CHOLMOD.Sparse(p)
 end
 
+# Pass 8 instead of CHOLMOD_INTLONG which no longer exists.
 @testset "illegal itype I" begin
     p = Ti == Int64 ? cholmod_l_allocate_sparse(1, 1, 1, true, true, 0, CHOLMOD_REAL, getcommon(Ti)) :
         cholmod_allocate_sparse(1, 1, 1, true, true, 0, CHOLMOD_REAL, getcommon(Ti))
     puint = convert(Ptr{UInt32}, p)
-    unsafe_store!(puint, CHOLMOD_INTLONG, 3*div(sizeof(Csize_t), 4) + 5*div(sizeof(Ptr{Cvoid}), 4) + 2)
+    unsafe_store!(puint, 123, 3*div(sizeof(Csize_t), 4) + 5*div(sizeof(Ptr{Cvoid}), 4) + 2)
```
mmuetzel, line 246: The above comment suggests that this should be "8" (not "123"). (But admittedly, I don't know what any of this means...)

giordano: @rayegun you changed this in #480

imciner2: I believe this test is checking that we get the right exception thrown from the wrapper when the int value is incorrect. In old versions, `INT_LONG` would throw that exception, but now that was removed, this looks to just be a random number that will throw the exception still.

giordano: Should we just remove this testset then?

mmuetzel: If there is no other test that checks that the correct exception is thrown, it might be worth keeping this.
But maybe, use the same (bogus) integer in the comment above this test to avoid confusion in the future. It might also help to point out that the integer is one that isn't valid input.

imciner2: > But maybe, use the same (bogus) integer in the comment above this test to avoid confusion in the future.

Which comment are you referring to?

mmuetzel: The one in the new line 241:
```
# Pass 8 instead of CHOLMOD_INTLONG which no longer exists.
```

rayegun: Sorry this was definitely a "get it work" change. There should just be a comment above saying that we're using a bogus Itype to ensure an exception is correctly thrown. We still want to test it, I'm not sure if it should be `8` or some random integer like `123`.

giordano: I improved the comments (I hope).  How does it look now?

mmuetzel: Looks good to me. Thanks! 👍

giordano:
@ViralBShah @rayegun can you please review and merge if/when ready?  As a reminder, my suggested plan is outlined at https://github.com/JuliaSparse/SparseArrays.jl/pull/478#issuecomment-1875826215

test/cholmod.jl
giordano: Thanks for the help!

ViralBShah:
Thanks @giordano. Going forward, we should probably always upgrade the wrappers with every minor version update to SuiteSparse. Even longer term, it would be nice to move all this out of stdlib, although I wonder if upgradable stdlibs will actually make the upgrading experience a bit more pleasant.

PR #480, opened 2024-01-02
Wrappers for v7.4.0

rayegun:
Now we play a game of chicken. Which do we merge first? This or https://github.com/JuliaLang/julia/pull/52577


rayegun:
Do not merge yet regardless, I need to test this locally and be sure there weren't technically breaking changes.

giordano:
I think this should go together with #478.

giordano:
Ah, test failures on nightly are the same as https://github.com/JuliaLang/julia/pull/52577#issuecomment-1862815917

rayegun:
I really don't understand this permutation thing. I will try again to find the cause

giordano:
Good news is that it seems to be reproducible on 32-bit Linux, which should be easier to debug than 32-bit Windows, as you can run the 32-bit build on a 64-bit Linux system.

rayegun:
Where can I download the 32-bit build?

giordano:
Since the failure in CI is reproducible with a plain nightly build, you can use https://julialangnightlies-s3.julialang.org/bin/linux/i686/julia-latest-linux-i686.tar.gz (from https://julialang.org/downloads/nightlies/)

PR #484, opened 2024-01-06
Reduce number of `*` methods by adopting `matprod_dest`

dkarrasch:


dkarrasch:
Windows tests seemed to be screwed. Anyway, these changes are not platform-dependent, so I take the passing Linux tests.

PR #485, opened 2024-01-07
Reduce allocation of dense arrays on-the-fly in linalg tests

dkarrasch:


dkarrasch:
Doesn't seem to have an overall impact in terms of runtime of the tests... but still nice to have, I'd say.

PR #486, opened 2024-01-09
Add generic matmatmul for inplace sparse x sparse

dkarrasch:
Fixes https://github.com/JuliaLang/julia/issues/52429.

I have no idea which branches are really used and whether this can be further reduced, especially in favor of compile time, which was a big issue over at LinearAlgebra.jl. Currently, this is just a copy-paste from the old LinAlg code.

@dmbates @palday


ViralBShah:
We should merge if ready - the windows issue is separate.

palday:
can you point to commit + lines of the old LinearAlgebra code this is copied from?

dkarrasch:
It's the one you found here: https://github.com/JuliaLang/julia/issues/52429#issuecomment-1847654946, i.e., https://github.com/JuliaLang/julia/commit/0cf2bf143552c2f425150fcb2a60225c67ea042d. It's in matmul.jl, lines 786ff.

dkarrasch:
Shall we merge this? I believe there is room for further improvement by providing specialized ` copyto!` and `copy_transpose` methods, but I don't have the time to do that now.

ViralBShah:
Agree - we should merge as is for now.

PR #487, opened 2024-01-10
Add Support for Single Precision

rayegun:
Incomplete, posting now for codecov purposes. There are some outstanding issues on how to do promotion and conversion, since the CHOLMOD built-in capabilities are limited.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,5 +1,5 @@
 name = "SparseArrays"
-uuid = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
+uuid = "2f01184e-e22b-5af5-ae63-d93ebab69eaf"
 version = "1.11.0"
 
 [deps]
--- a/Project.toml
+++ b/Project.toml
@@ -1,5 +1,5 @@
 name = "SparseArrays"
-uuid = "2f01184e-e22b-5af5-ae63-d93ebab69eaf"
+uuid = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
 version = "1.11.0"
 
 [deps]
--- a/src/solvers/cholmod.jl
+++ b/src/solvers/cholmod.jl
@@ -637,7 +637,7 @@ for TI ∈ IndexTypes
     end
 
     function ssmult(A::Sparse{Tv, $TI}, B::Sparse{Tv, $TI}, stype::Integer,
-        values::Bool, sorted::Bool) where Tv<:VRealTypes
+        values::Integer, sorted::Bool) where Tv<:VTypes
         lA = unsafe_load(pointer(A))
         lB = unsafe_load(pointer(B))
         if lA.ncol != lB.nrow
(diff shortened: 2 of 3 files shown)


ViralBShah:
Wow that is a huge PR!

rayegun:
The vast majority of it is trying to handle promotion at least somewhat sanely. Since mixed precision is not supported (and probably never will be).

rayegun:
Curses, I have no idea what's going on with Float32 on 32-bit plats. Are there arcane architectural differences afoot here? Or just a bad wrap somewhere :/

rayegun:
This seems to be an artifact of using `sprand` here perhaps? Does that make sense to you @ViralBShah? I suppose it's possible that the matrix generated here is close to non-posdef, and that minor architectural changes are causing the issue?

If I run 
```julia
julia> Ftmp = SparseMatrixCSC{Float32, Int32}(sprandn(Tv, 100, 100, 0.1)); Ftmp = Ftmp'Ftmp + I;

julia> cholesky(Ftmp)
```
There's a 1 in 4 chance it fails, but not so with `Float64` on 32-bit Linux. This *never* occurs on 64 bit platforms, no matter which datatypes or index types are used.

ViralBShah:
Just add a heavy diagonal:

`SparseMatrixCSC{Float32, Int32}(sprandn(Tv, 100, 100, 0.1)); Ftmp = Ftmp'Ftmp + 100*I`

Float32 can end up being non-posdef to numerical precision.

ViralBShah:
There's a 20min timeout.

ViralBShah:
Does the failing test on ComplexF32 suggest that there is a bug somewhere in the cholesky for single precision just added in the complex case?

rayegun:
I need to trim the tests. I think I just located a bug in the wrapper, I'm running the tests on Ubuntu-32 right now. There was an assumption that only held for 64-bit case by accident.

ViralBShah:
The tests are timing out on Windows after 6 hours. I have triggered the x64 test again. I believe it was going through successfully earlier in this PR, but perhaps the hanging is a sign of memory corruption.

dkarrasch:
I think Windows tests have been hanging for a while, also in other PRs.

rayegun:
The tests are currently massive for cholesky. I will work on reducing, we don't really need a complete product of Float32, Float64 x Real, Complex x Int32, Int64 for every test.

I just did that to naively check everything while making the changes

ViralBShah:
The hangs since SuiteSparse 7.4 are a bit worrisome.

rayegun:
Are those hangs really from SuiteSparse? The times for the SuiteSparse tests don't look that terrible. They're faster than my MacBook. Which definitely doesn't take 5 hours.

rayegun:
Oh looks like indeed they happen after SuiteSparse 7.4 update. I wonder why 🤔  because the actual times don't look terrible.

ViralBShah:
It's not straightforward to do this - but it would be great to have a PR for SuiteSparse 7.5 and see if that has the same issue.

ViralBShah:
With #490 clearly identifying the issue on Windows, I think it is safe to merge this.

ViralBShah (approved):

PR #490, opened 2024-01-15
Disable threading tests which seem to fail on Windows with SuiteSparse 7.4

ViralBShah:


ViralBShah:
@rayegun @dkarrasch The threading tests are the culprits on windows with SuiteSparse 7.4.

rayegun:
Let me think a bit about what could be going wrong here but yeah we should move forward to 7.5+

ViralBShah:
#497 bumps to 7.5.1 and nightly Julia builds starting tomorrow should include it.

ViralBShah:
@rayegun Seems like it just fails in Github CI and not on Julia CI.

PR #503, opened 2024-02-06
Move `fkeep!` docstring to the right function

lgoettgens:
Handles the `SparseArrays` part of https://github.com/JuliaLang/julia/issues/52725.

The docstring was already there, it was just attached to the wrong function.

In which file should I put the test?
```julia
@test Docs.undocumented_names(SparseArrays) == []
```

cc @stevengj

### DIFF
--- a/src/sparsematrix.jl
+++ b/src/sparsematrix.jl
@@ -1763,35 +1763,6 @@ end
 
 ## fkeep! and children tril!, triu!, droptol!, dropzeros[!]
 
-"""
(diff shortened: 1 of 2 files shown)


stevengj:
I would just put the test in the top-level file https://github.com/JuliaSparse/SparseArrays.jl/blob/main/test/runtests.jl

lgoettgens:
> Can you put it in issues.jl with a link back to the original issue in Julialang? I will merge as soon as that's done

Done. The windows CI failure seems to be unrelated.

PR #504, opened 2024-02-06
Update ci.yml  - add codecov token

ViralBShah:


ViralBShah:
The windows tests continue to be problematic but only on github actions (and not in Julia CI).

PR #509, opened 2024-02-07
Fix error message in getcommon() in cholmod

ViralBShah:
@rayegun I am not sure if this is right. I couldn't find `cholmod_common_l` anywhere.

### DIFF
--- a/src/solvers/cholmod.jl
+++ b/src/solvers/cholmod.jl
@@ -166,7 +166,7 @@ end
 function newcommon(; print = 0) # no printing from CHOLMOD by default
     common = finalizer(cholmod_finish, Ref(cholmod_common()))
     result = cholmod_start(common)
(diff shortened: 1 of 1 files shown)


rayegun:
Yeah that's just to ensure they stay separated.

What was this PR for?

ViralBShah:
I was looking for anything that might help resolve the threaded tests that fail in github actions. They don't fail on the julia buildbots but they always fail in github actions - and I was wondering if that was a real bug on our side. Specifically making sure that structs are correctly created and mirror what CHOLMOD expects (no overflows, etc.).

rayegun:
Can you point me to the logs that fail?

ViralBShah:
Yeah, see the failing windows tests in the CI right here in this PR. It times out in the threading tests.

PR #511, opened 2024-02-14
Speedup for `findmin()/findmax()` on sparse arrays (#510)

### LINKED ISSUE #510: `findmin(A; dims=1)` is much slower than manually looping over.
Some demo:
```julia
using SparseArrays

# setups
n = 10000
halfhalf = a -> SparseMatrixCSC(a.m, a.n, a.colptr, a.rowval, a.nzval .- 0.5);
symmetrize = a -> (a + a')/2;
A = halfhalf(symmetrize(sprand(n, n, 0.1)));  # nnz(A) == 19004033
function manualmincol(a)
   local mincols = zeros(Int, a.m)
   local minvals = zeros(Float64, a.m)
   for c in 1:a.m
       local rb, re = a.colptr[c], a.colptr[c+1]
       if rb == re
           continue
       end
       local minval::Float64, row_::Int64 = Inf, 0
       for r in rb:(re-1)
           local val, row = a.nzval[r], a.rowval[r]
           if val < minval
               minval, row_ = val, row
           end
       end
       mincols[c], minvals[c] = row_, minval
   end
   return mincols, minvals
end

# test
@time bob = manualmincol(A)[1];
# output: 0.138503 seconds (5 allocations: 156.375 KiB)

@time bob2 = findmin(A, dims=1)[2];
# output: 2.414430 seconds (9 allocations: 391.094 KiB)

# check that they are the same
bob2 = Vector(first.(Tuple.(bob2))'[:,1]);
bob2 == bob
# output: true
```

I believe the reason is that with `dims=...` it goes through a non-specialised version of `findmin()`:
```julia
julia> @which findmin(A, dims=1)
kwcall(::NamedTuple, ::typeof(findmin), A::AbstractArray)
     @ Base reducedim.jl:1130
```

Does it make sense to "intercept" the `dims` argument and provide a faster implementation? (I understand that the performance for row and column aggregations would be different, but currently both are much slower than spelling out the loops)

pratyai:
1. Parameterised `_findr()` on the element type of the sparse array. Otherwise, every single `op()` comparison allocates, every time `_findr()` is called (not sure if there is another way to optimise those allocations away).
2. Added a keyword argument `dims` to the existing functions, with default values to preserve current behaviours when `dims` is not provided.
3. Changed the type of `_findr()` a bit. It's not used anywhere else, and the change simplifies the previous changes.

(Example: https://pastebin.com/raw/47xxMkHa shows different allocation and timing behaviour at different stages of the 2 included commits. The direct `_findr()` needs to be adjusted to account for the type change)

Fixes #510.

### DIFF
--- a/src/sparsematrix.jl
+++ b/src/sparsematrix.jl
@@ -2697,7 +2697,7 @@ function _findr(op, A::AbstractSparseMatrixCSC{Tv}, region) where {Tv}
                 Iv = CartesianIndex(rowval[j], i)
             end
         end
-        return (fill(Sv,1,1), fill(Iv,1,1))
+        return Sv, Iv
     else
         throw(ArgumentError("invalid value for region; must be 1, 2, or (1,2)"))
     end
@@ -2710,8 +2710,8 @@ findmin(A::AbstractSparseMatrixCSC{Tv}, region::Union{Integer,Tuple{Integer},NTu
     _findr(_isless_fm, A, region)
 findmax(A::AbstractSparseMatrixCSC{Tv}, region::Union{Integer,Tuple{Integer},NTuple{2,Integer}}) where {Tv} =
     _findr(_isgreater_fm, A, region)
-findmin(A::AbstractSparseMatrixCSC) = (r=findmin(A,(1,2)); (r[1][1], r[2][1]))
-findmax(A::AbstractSparseMatrixCSC) = (r=findmax(A,(1,2)); (r[1][1], r[2][1]))
+findmin(A::AbstractSparseMatrixCSC; dims::Union{Integer,Tuple{Integer},NTuple{2,Integer}}=(1,2)) = findmin(A, dims)
+findmax(A::AbstractSparseMatrixCSC; dims::Union{Integer,Tuple{Integer},NTuple{2,Integer}}=(1,2)) = findmax(A, dims)
 
 argmin(A::AbstractSparseMatrixCSC) = findmin(A)[2]
 argmax(A::AbstractSparseMatrixCSC) = findmax(A)[2]
--- a/src/sparsematrix.jl
+++ b/src/sparsematrix.jl
@@ -2629,7 +2629,7 @@ function _findz(A::AbstractSparseMatrixCSC{Tv,Ti}, rows=1:size(A, 1), cols=1:siz
     return CartesianIndex(0, 0)
 end
 
-function _findr(op, A, region, Tv)
+function _findr(op, A::AbstractSparseMatrixCSC{Tv}, region) where {Tv}
     require_one_based_indexing(A)
     Ti = eltype(keys(A))
     i1 = first(keys(A))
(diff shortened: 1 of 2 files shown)


dkarrasch:
So, one issue was that when you call `findmin(A; dims=1)` this got dispatched to the generic method, which calls `_findmin(A, dims)` which is not overloaded here by SparseArrays.jl, right? And once you fixed that, then you need that the `Tv` type is compiled into the `findr` code rather than just provided as the fourth element to avoid the allocations? But why can't we infer the return type anymore?

pratyai:
> So, one issue was that when you call findmin(A; dims=1) this got dispatched to the generic method, which calls _findmin(A, dims) which is not overloaded here by SparseArrays.jl, right?

Yes; that's what I started with.

> And once you fixed that, then you need that the Tv type is compiled into the findr code rather than just provided as the fourth element to avoid the allocations?

Yes; otherwise I couldn't avoid the allocations.

> But why can't we infer the return type anymore?

Those 4 line changes were unnecessary indeed; I took them out now. I initially had another change in `_findr()` which didn't work with `@inferred`. I removed that change but forgot to undo the corresponding change in the test. Thanks for pointing out.

dkarrasch:
Awesome. Could you please quickly post timings for the original problem before and after the change?

pratyai:
Common setup code (in REPL):
```julia
using SparseArrays

# setups
function manualmin(a)
   local minat = (0, 0)
   local minval = length(a.nzval) < a.n*a.m ? 0.0 : Inf
   for c in 1:a.m
       local rb, re = a.colptr[c], a.colptr[c+1]
       if rb == re
           continue
       end
       for r in rb:(re-1)
           local val, row = a.nzval[r], a.rowval[r]
           if val < minval
               minval, minat = val, (r, c)
           end
       end
   end
   return minat, minval
end
function manualmincol(a)
   local mincols = zeros(Int, a.m)
   local minvals = zeros(Float64, a.m)
   for c in 1:a.m
       local rb, re = a.colptr[c], a.colptr[c+1]
       if rb == re
           continue
       end
       local minval::Float64, row_::Int64 = Inf, 0
       for r in rb:(re-1)
           local val, row = a.nzval[r], a.rowval[r]
           if val < minval
               minval, row_ = val, row
           end
       end
       mincols[c], minvals[c] = row_, minval
   end
   return mincols, minvals
end

using SparseArrays
n = 10000
halfhalf = a -> SparseMatrixCSC(a.m, a.n, a.colptr, a.rowval, a.nzval .- 0.5);
symmetrize = a -> (a + a')/2;
A = halfhalf(symmetrize(sprand(n, n, 0.1)));  # nnz(A) == 19004033
```

Before (in new instance):
```julia
julia> @time findmin(A; dims=1);
  3.680312 seconds (498.03 k allocations: 25.737 MiB, 11.85% compilation time)
julia> @time findmin(A; dims=1);
  3.076859 seconds (17 allocations: 822.953 KiB, 0.13% gc time)
julia> @time findmin(A; dims=2);
  6.395606 seconds (16 allocations: 744.781 KiB)
julia> @time findmin(A; dims=2);
  6.299757 seconds (16 allocations: 744.781 KiB)
julia> @time findmin(A; dims=(1,2));
  3.134760 seconds (33.66 k allocations: 1.674 MiB, 1.76% compilation time)
julia> @time findmin(A; dims=(1,2));
  3.023886 seconds (11 allocations: 1.203 KiB)
```

After (in new instance):
```julia
julia> @time findmin(A);
  0.165648 seconds (324.78 k allocations: 16.157 MiB, 82.11% compilation time)
julia> @time findmin(A);
  0.026682 seconds (5 allocations: 208 bytes)
julia> @time findmin(A; dims=1);
  0.163995 seconds (275.03 k allocations: 13.885 MiB, 78.06% compilation time)
julia> @time findmin(A; dims=1);
  0.034279 seconds (7 allocations: 234.562 KiB)
julia> @time findmin(A; dims=2);
  0.034301 seconds (7 allocations: 234.562 KiB)
julia> @time findmin(A; dims=2);
  0.033291 seconds (7 allocations: 234.562 KiB)
julia> @time findmin(A; dims=(1,2));
  0.029890 seconds (2.96 k allocations: 148.984 KiB, 12.98% compilation time)
julia> @time findmin(A; dims=(1,2));
  0.025771 seconds (5 allocations: 208 bytes)
```

ViralBShah:
What is the codecov issue in the CI? Not sure I fully understand what to do.

dkarrasch:
LGTM, up to the unrelated codecov issue.

PR #527, opened 2024-04-02
Revert "Don't fail CI if codecov upload fails."

DilumAluthge:
Reverts JuliaSparse/SparseArrays.jl#525


DilumAluthge:
I believe that we've identified the Codecov issue as specific to Dependabot's access to secrets: https://github.com/JuliaSparse/SparseArrays.jl/pull/524#issuecomment-2032353608

Therefore, we can revert #525.

ViralBShah:
FWIW, I don't like codecov upload failing blocking PRs even otherwise. But for now, let's see how it goes.

PR #532, opened 2024-04-08
Sparse versions of repeat for matrices and vectors

mjacobse:
Calling `repeat` for sparse matrices or vectors currently calls the `Base` function, which ends up building the result rather inefficiently with indexing brackets.

This would add sparse versions that use the info that inputs are in CSC format to provide more efficient implementations. To keep it simple it is limited to the straightforward cases of outer repetition along the first two dimensions. Row-wise repetition is intentionally kept close to the implementation for `vcat`, also using its helper function `stuffcol!` (with slight modifications). Column-wise repetition is kept simple by deferring to `Base.repeat` on the `colptr`, `rowval`, and `nzval` components of the CSC format.

### DIFF
--- a/src/sparsematrix.jl
+++ b/src/sparsematrix.jl
@@ -3908,11 +3908,12 @@ function vcat(X::AbstractSparseMatrixCSC...)
         ptr_res = colptr[c]
         for i = 1 : num
             colptrXi = getcolptr(X[i])
+            rowvalXi = rowvals(X[i])
+            nzvalXi = nonzeros(X[i])
             col_length = colptrXi[c + 1] - colptrXi[c]
             ptr_Xi = colptrXi[c]
 
-            stuffcol!(X[i], colptr, rowval, nzval,
-                      ptr_res, ptr_Xi, col_length, mX_sofar)
+            stuffcol!(rowval, nzval, ptr_res, rowvalXi, nzvalXi, ptr_Xi, col_length, mX_sofar)
 
             ptr_res += col_length
             mX_sofar += mX[i]
@@ -3922,12 +3923,8 @@ function vcat(X::AbstractSparseMatrixCSC...)
     SparseMatrixCSC(m, n, colptr, rowval, nzval)
 end
 
-@inline function stuffcol!(Xi::AbstractSparseMatrixCSC, colptr, rowval, nzval,
-                           ptr_res, ptr_Xi, col_length, mX_sofar)
-    colptrXi = getcolptr(Xi)
-    rowvalXi = rowvals(Xi)
-    nzvalXi  = nonzeros(Xi)
-
+@inline function stuffcol!(rowval, nzval, ptr_res, rowvalXi, nzvalXi, ptr_Xi,
+                           col_length, mX_sofar)
     for k=ptr_res:(ptr_res + col_length - 1)
         @inbounds rowval[k] = rowvalXi[ptr_Xi] + mX_sofar
         @inbounds nzval[k]  = nzvalXi[ptr_Xi]
--- a/src/sparsematrix.jl
+++ b/src/sparsematrix.jl
@@ -3911,9 +3911,8 @@ function vcat(X::AbstractSparseMatrixCSC...)
             col_length = colptrXi[c + 1] - colptrXi[c]
             ptr_Xi = colptrXi[c]
 
-            stuffcol!(rowval, nzval, ptr_res, rowvals(X[i]), nonzeros(X[i]), ptr_Xi, col_length, mX_sofar)
-
-            ptr_res += col_length
+            ptr_res = stuffcol!(rowval, nzval, ptr_res, rowvals(X[i]), nonzeros(X[i]), ptr_Xi,
+                                col_length, mX_sofar)
             mX_sofar += mX[i]
         end
         colptr[c + 1] = ptr_res
@@ -3928,6 +3927,7 @@ end
         @inbounds nzval[k]  = nzvalXi[ptr_Xi]
         ptr_Xi += 1
     end
+    return ptr_res + col_length
 end
 
 function hcat(X::AbstractSparseMatrixCSC...)
@@ -3984,9 +3984,8 @@ function Base.repeat(A::AbstractSparseMatrixCSC, m)
         col_length = getcolptr(A)[c + 1] - ptr_source
         for index_repetition = 0 : (m - 1)
             row_offset = index_repetition * size(A, 1)
-            stuffcol!(rowval, nzval, ptr_res, rowvals(A), nonzeros(A), ptr_source,
-                      col_length, row_offset)
-            ptr_res += col_length
+            ptr_res = stuffcol!(rowval, nzval, ptr_res, rowvals(A), nonzeros(A), ptr_source,
+                                col_length, row_offset)
         end
         colptr[c + 1] = ptr_res
     end
--- a/src/sparsematrix.jl
+++ b/src/sparsematrix.jl
@@ -3970,6 +3970,47 @@ function hcat(X::AbstractSparseMatrixCSC...)
     SparseMatrixCSC(m, n, colptr, rowval, nzval)
 end
 
+
+# Efficient repetition of sparse matrices
+
+function Base.repeat(A::AbstractSparseMatrixCSC, m)
+    colptr_source = getcolptr(A)
+    rowval_source = rowvals(A)
+    nzval_source = nonzeros(A)
+
+    nnz_new = nnz(A) * m
+    colptr = similar(colptr_source, length(colptr_source))
+    rowval = similar(rowval_source, nnz_new)
+    nzval = similar(nzval_source, nnz_new)
+
+    colptr[1] = 1
+    for c = 1 : size(A, 2)
+        ptr_res = colptr[c]
+        ptr_source = colptr_source[c]
+        col_length = colptr_source[c + 1] - ptr_source
+        for index_repetition = 0 : (m - 1)
+            row_offset = index_repetition * size(A, 1)
+            stuffcol!(rowval, nzval, ptr_res, rowval_source, nzval_source, ptr_source,
+                      col_length, row_offset)
+            ptr_res += col_length
+        end
+        colptr[c + 1] = ptr_res
+    end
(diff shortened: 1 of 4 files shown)


SobhanMP:
was `vcat` using the wrong length?

mjacobse:
> was `vcat` using the wrong length?

Are you referring to `col_length` and the change in 60eff16? No, in effect it was used correctly. Basically the helper `stuffcol!` iterates from `0:(n-1)` where `n` is the number of nonzeros in the column. Before 60eff16, it expected you to pass in `n-1` for the argument called `col_length` and then iterated `0:col_length`. After 60eff16 it expects you to pass in `n` for the argument called `col_length` and then iterates `0:(col_length-1)`. So the behavior does not change, but to my mind it is much clearer and consistent with the name of the argument. And as I intended to reuse the function in the new `repeat`, I thought it would be better to make this clarifying change instead of reproducing the somewhat confusing usage.

ViralBShah:
@SobhanMP Can you add your review to this PR?

src/sparsematrix.jl
```diff
@@ -3908,27 +3908,24 @@ function vcat(X::AbstractSparseMatrixCSC...)
         ptr_res = colptr[c]
         for i = 1 : num
             colptrXi = getcolptr(X[i])
-            col_length = (colptrXi[c + 1] - 1) - colptrXi[c]
+            rowvalXi = rowvals(X[i])
```
SobhanMP, line 3911: I think at some point we agreed not to create temporary variables for this use `rowvals(X[i])` directly it gets optimized away by the compiler

src/sparsematrix.jl
```diff
@@ -3908,27 +3908,24 @@ function vcat(X::AbstractSparseMatrixCSC...)
...
 end
 
-@inline function stuffcol!(Xi::AbstractSparseMatrixCSC, colptr, rowval, nzval,
-                           ptr_res, ptr_Xi, col_length, mX_sofar)
-    colptrXi = getcolptr(Xi)
-    rowvalXi = rowvals(Xi)
-    nzvalXi  = nonzeros(Xi)
-
-    for k=ptr_res:(ptr_res + col_length)
+@inline function stuffcol!(rowval, nzval, ptr_res, rowvalXi, nzvalXi, ptr_Xi,
```
SobhanMP, line 3926: add an alias for the original method on AbstractSparsematrixCSC

src/sparsematrix.jl
```diff
@@ -3973,6 +3970,47 @@ function hcat(X::AbstractSparseMatrixCSC...)
...
+    SparseMatrixCSC(size(A, 1) * m, size(A, 2), colptr, rowval, nzval)
+end
+
+function Base.repeat(A::AbstractSparseMatrixCSC, m, n)
+    B = repeat(A, m)
+    nnz_per_column = diff(getcolptr(B))
+    colptr = cumsum(vcat(1, repeat(nnz_per_column, n)))
+    rowval = repeat(rowvals(B), n)
+    nzval = repeat(nonzeros(B), n)
+    SparseMatrixCSC(size(B, 1), size(B, 2) * n, colptr, rowval, nzval)
```
SobhanMP, line 4010: cool implementation

src/sparsematrix.jl
```diff
@@ -3973,6 +3970,47 @@ function hcat(X::AbstractSparseMatrixCSC...)
     SparseMatrixCSC(m, n, colptr, rowval, nzval)
 end
 
+
+# Efficient repetition of sparse matrices
+
+function Base.repeat(A::AbstractSparseMatrixCSC, m)
+    colptr_source = getcolptr(A)
```
SobhanMP, line 3977: same thing here, don't assign getcolptr,rowvals, etc to variables

src/sparsematrix.jl
```diff
@@ -3973,6 +3970,47 @@ function hcat(X::AbstractSparseMatrixCSC...)
...
+    nzval = similar(nzval_source, nnz_new)
+
+    colptr[1] = 1
+    for c = 1 : size(A, 2)
+        ptr_res = colptr[c]
+        ptr_source = colptr_source[c]
+        col_length = colptr_source[c + 1] - ptr_source
+        for index_repetition = 0 : (m - 1)
+            row_offset = index_repetition * size(A, 1)
+            stuffcol!(rowval, nzval, ptr_res, rowval_source, nzval_source, ptr_source,
```
SobhanMP, line 3993: would it make sense for `stuffcol!` to return the `new ptr_Xi`?

src/sparsevector.jl
```diff
@@ -1308,6 +1308,38 @@ hvcat(rows::Tuple{Vararg{Int}}, n1::Number, ns::Vararg{Number}) = invoke(hvcat,
 hvcat(rows::Tuple{Vararg{Int}}, n1::N, ns::Vararg{N}) where {N<:Number} = invoke(hvcat, Tuple{typeof(rows), Vararg{N}}, rows, n1, ns...)
 
 
+### Efficient repetition of sparse vectors
+
+function Base.repeat(v::AbstractSparseVector, m)
+    nnz_source = nnz(v)
+    nnz_new = nnz_source * m
+
+    nzind_source, nzval_source = findnz(v)
```
SobhanMP, line 1317: avoid using `findnz`, it's allocating. How about
```julia
    nzind = similar(nonzeroinds(v), nnz_new)
    nzval = similar(nonzeros(v), nnz_new)

    ptr_res = 1
    for index_repetition = 0:(m-1)
        row_offset = index_repetition * length(v)
        stuffcol!(nzind, nzval, ptr_res, nonzeroinds(v), nonzeros(v), 1, nnz_source, row_offset)
        ptr_res += nnz_source
    end
```
it uses 4 fewer allocations

src/sparsevector.jl
```diff
@@ -1308,6 +1308,38 @@ hvcat(rows::Tuple{Vararg{Int}}, n1::Number, ns::Vararg{Number}) = invoke(hvcat,
...
+    @assert ptr_res == nnz_new + 1
+
+    SparseVector(length(v) * m, nzind, nzval)
+end
+
+function Base.repeat(v::AbstractSparseVector, m, n)
+    w = repeat(v, m)
+    nzind_source, nzval_source = findnz(w)
+
+    colptr = Vector{eltype(nzind_source)}(1 .+ nnz(w) * (0:n))
```
SobhanMP, line 1336: same thing as the one dim version.

src/sparsematrix.jl
mjacobse: Addressed by b961a7f. I kept the `getcolptr` stuff in `vcat` as it was before, to avoid touching lines unnecessarily. Let me know if you disagree and would like it changed by this PR.

src/sparsevector.jl
mjacobse: Thanks a lot, I did not realize `findnz` does a copy, I thought it's just a convenient way to get essentially both `nonzeroinds` and `nonzeros`. Fixed in b961a7f

src/sparsematrix.jl
mjacobse: I initially decided against this to keep the changes to `stuffcol!` to a minimum in this PR. But I agree, it's a simple enough change and makes the call sites simpler, so since you suggested it I added this in 6db2712

src/sparsematrix.jl
mjacobse: Sorry, I don't really understand what you mean by that, could you elaborate?

src/sparsematrix.jl
mjacobse: As mentioned I intentionally kept the column-wise repetition very simple for now. It certainly does leave room for performance improvements, but I'd imagine they would in most cases be small compared to the improvement that this PR brings over the non-sparse version of `repeat`. In the end, the result data `colptr`, `rowval` and `nzval` will require the largest allocations (unless `n <= 1`) and they will always be necessary. But if you prefer, I can look into a more optimized implementation that avoids the smaller allocations for the helper `B`.

src/sparsematrix.jl
SobhanMP: It might make sense to add a method that makes any code that uses the old signature of `stuffcol!` work (and would call the new function),

you decided if it makes sense.

src/sparsematrix.jl
SobhanMP: That sounds good; let's leave it as is, then.

SobhanMP: sounds good.

src/sparsevector.jl
SobhanMP: perfect

src/sparsematrix.jl
SobhanMP: thanks!

src/sparsematrix.jl
SobhanMP: Please add an overload to stuffcol! that would keep all the existing calls to that function working.

src/sparsematrix.jl
SobhanMP: since it's jus style, let's keep the pull request focused on just `repeat`

src/sparsematrix.jl
mjacobse: Oh I see, thanks. Inside of `SparseArrays`, the only occurence of `stuffcol!` was really as this small helper function for `vcat`, hence why it was the only call site that had to be adapted. Or should we be concerned about external users of `SparseArrays` possibly depending on it? I can hardly imagine anyone using it, since it is not documented in any way as part of the API as far as I can tell. But I am not sure about the usual policy in Julia regarding such breakages, with technically any function being visible I suppose. So I could add back a version with the old signature if you prefer. However then that version should also still have the old somewhat confusing behavior in regard to the parameter `col_length` I guess. Seems a little ugly.

SobhanMP: hmmm, let's leave it as is then.

mjacobse:
Thanks for the review! Anything left that I can do to help this move this forward?

ViralBShah:
@SobhanMP Good to merge?

SobhanMP:
yes, sorry I was behind a deadline :sweat_smile: @mjacobse, thanks for the code.

PR #547, opened 2024-07-27
implement in-place `ldiv!` for Cholesky factorization

### LINKED ISSUE #123: ldiv!(Y, A::SuiteSparse.CHOLMOD.Factor, B) errors
The reason seems to be that in the 3-arg `ldiv!` its 2-arg-version is called, which is missing for sparse Cholesky factorizations.

MWE:
```
A = sprand(100, 100, 0.01) + spdiagm(0 => ones(100))
x = rand(100)
y = similar(x)
ldiv!(y, factorize(A'A), x)
```
throws
```
ERROR: MethodError: no method matching ldiv!(::SuiteSparse.CHOLMOD.Factor{Float64}, ::Array{Float64,1})
```
Asking for the employed method gives
```
julia> @which ldiv!(y, factorize(A'A), x)
ldiv!(Y::Union{AbstractArray{T,1}, AbstractArray{T,2}} where T, A::Factorization, B::Union{AbstractArray{T,1}, AbstractArray{T,2}} where T) in LinearAlgebra at /Users/osx/buildbot/slave/package_osx64/build/usr/share/julia/stdlib/v1.0/LinearAlgebra/src/factorization.jl:100
```

### LINKED ISSUE #275: Support `ldiv!` SuiteSparse.CHOLMOD.Factor
I through that it was not supported in `SuiteSparse` but it's not the case (https://github.com/DrTimothyAldenDavis/SuiteSparse/issues/136).
We could do something similar to https://github.com/SobhanMP/SuiteSparse.jl/commit/7093e4ac176787983b56f8f157c294cae28ca034 and add the additional two vectors `Y` and `E` in the Julia structure.

[CHOLMOD_UserGuide.pdf](https://github.com/JuliaSparse/SparseArrays.jl/files/9793362/CHOLMOD_UserGuide.pdf) <- pages 119-120.

### LINKED ISSUE #319: missing `ldiv!` overload for sparse Cholesky
It seems like this should be supported:
```

julia> A = sprand(10,10,0.1); A = A'A + I;

julia> F = cholesky(A)
SuiteSparse.CHOLMOD.Factor{Float64}
type:    LLt
method:  simplicial
maxnnz:  11
nnz:     11
success: true

julia> b = rand(10); x = similar(b);

julia> ldiv!(x, F, b)
ERROR: MethodError: no method matching ldiv!(::SuiteSparse.CHOLMOD.Factor{Float64}, ::Vector{Float64})
```

Similar to #242, except in this case CHOLMOD *does* seem to provide an in-place API (whereas currently [we are calling](72827cd31f0aa346cfc54158f9c6db1738ecd9e4) the out-of-place `cholmod_l_spsolve`):

image

cc @ChrisRackauckas, since this just came up in https://github.com/SciML/LinearSolve.jl/pull/246

ranocha:
I have implemented an in-place `ldiv!` for Cholesky factorizations of sparse matrices. 

Closes #319 
Closes #275
Closes #123
Xref https://discourse.julialang.org/t/is-there-no-ldiv-for-choleski-factorization-of-a-sparse-matrix/114806

### Caveats

The method works locally but is not as elegant as the UMFPack/LU version or the other code in the CHOLMOD wrapper:
- UMFPack accepts arrays/pointers directly whereas CHOLMOD uses `cholmod_dense` structs.
- The `Dense` wrappers in the Julia CHOLMOD module allocate new arrays.

Thus, the code looks a bit different from the remaining code in the file.

The current method still allocates some memory on my system since we do not provide the workspace vectors `Handle_Y` and `Handle_E`. We could store them in the `CHOLMOD.Factor` struct. I tried that, but they were still `C_NULL` after the call to `solve2` (when initialized with `C_NULL`). Thus, re-using them is not working as I understand the docs. Thus, I have not added this complexity to this PR.
Maybe related: https://github.com/DrTimothyAldenDavis/SuiteSparse/issues/45

### Questions

Shall we announce this somewhere in some form of NEWS.md so that people can adopt to this change?

### DIFF
--- a/test/cholmod.jl
+++ b/test/cholmod.jl
@@ -13,7 +13,7 @@ using Random
 using Serialization
 using LinearAlgebra:
     I, cholesky, cholesky!, det, diag, eigmax, ishermitian, isposdef, issuccess,
-    issymmetric, ldlt, ldlt!, logdet, norm, opnorm, Diagonal, Hermitian, Symmetric,
+    issymmetric, ldiv!, ldlt, ldlt!, logdet, norm, opnorm, Diagonal, Hermitian, Symmetric,
     PosDefException, ZeroPivotException, RowMaximum
 using SparseArrays
 using SparseArrays: getcolptr
--- a/test/cholmod.jl
+++ b/test/cholmod.jl
@@ -284,6 +284,26 @@ end
     end
 end
 
+@testset "ldiv! $Tv $Ti" begin
+    local A, x, x2, b, X, X2, B
+    A = sprand(10, 10, 0.1)
+    A = I + A * A'
+    A = convert(SparseMatrixCSC{Tv,Ti}, A)
+    factor = cholesky(A)
+
+    x = fill(Tv(1), 10)
+    b = A * x
+    x2 = zero(x)
+    @inferred ldiv!(x2, factor, b)
+    @test x2 ≈ x
+
+    X = fill(Tv(1), 10, 5)
+    B = A * X
+    X2 = zero(X)
+    @inferred ldiv!(X2, factor, B)
+    @test X2 ≈ X
+end
+
 end #end for Ti ∈ itypes
 
 for Tv ∈ (Float32, Float64)
--- a/test/cholmod.jl
+++ b/test/cholmod.jl
@@ -302,6 +302,17 @@ end
     X2 = zero(X)
     @inferred ldiv!(X2, factor, B)
     @test X2 ≈ X
+
+    c = fill(Tv(1), size(x, 1) + 1)
+    C = fill(Tv(1), size(X, 1) + 1, size(X, 2))
+    y = fill(Tv(1), size(x, 1) + 1)
+    Y = fill(Tv(1), size(X, 1) + 1, size(X, 2))
+    @test_throws DimensionMismatch ldiv!(y, factor, b)
+    @test_throws DimensionMismatch ldiv!(Y, factor, B)
+    @test_throws DimensionMismatch ldiv!(x2, factor, c)
+    @test_throws DimensionMismatch ldiv!(X2, factor, C)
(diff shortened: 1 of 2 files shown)


ViralBShah:
cc @dkarrasch @rayegun

ranocha:
@ViralBShah @dkarrasch @rayegun Does any of you have some time to review this PR?

ViralBShah:
I will take a look today. Also hoping someone else can in addition.

test/cholmod.jl
```diff
@@ -281,6 +284,37 @@ end
     end
 end
 
+@testset "ldiv! $Tv $Ti" begin
```
rayegun, line 287: Does this test Dense input?

rayegun (commented):
With the exception that I'm not clear if Dense inputs are tested sgtm. Will merge once that has an answer

ViralBShah:
This looks straightforward. Merging. The PR just looked much bigger because of all the whitespace changes.

ViralBShah (approved):

test/cholmod.jl
ranocha: This tests standard dense Julia `Vector`s as input right-hand sides. `Dense` structs are tested above in the new lines 141-143.

ranocha:
Thanks a lot! I'm sorry for the confusing with the whitespace changes - VSCode did all that for me and I didn't notice before

ViralBShah:
Is it possible to add a docstring?

ViralBShah:
Yeah good point. Not necessary.

ranocha:
Thanks a lot for your reviews! Do you know which Julia version will include this patch?

ViralBShah:
This will go into 1.12.

ranocha:
Thanks!

PR #552, opened 2024-08-05
Keep sparse solvers docs as before

ViralBShah:


ViralBShah:
We don't necessarily need to reproduce the solver docs in the SparseArrays section, since they are already in LinearAlgebra. But they are scattered all over LinearAlgebra. So even if we could just have a listing here, that links to the SparseMatrixCSC specific solver docs  in LinearAlgebra, that should suffice.

PR #560, opened 2024-09-19
fix a broken link in docs

ctarn:


dkarrasch (approved):

fredrikekre:
Where do you see a broken link?

ctarn:
should we use a more specific id instead? I am not so familiar with `Documentor.jl` and am worried that the docs of julia may have used the id, and may raise a conflict.

fredrikekre:
Okay, looks like it has already been fixed in https://github.com/JuliaSparse/SparseArrays.jl/pull/430 so I think that PR should just be backported to 1.10

ctarn:
Yes! It has been fixed in the docs of 1.12-dev. I am going to close the PR.
BTW, jump to `[#man-csc](https://docs.julialang.org/en/v1/stdlib/SparseArrays/#man-csc)` may be a bug of Documentor.jl. I noticed that you are a member of JuliaDocs, and maybe I don't have to create another issue in JuliaDocs :).
Thanks!

PR #565, opened 2024-10-11
added specialized method for 3-argument dot with diagonal matrix

### LINKED ISSUE #563: slow 3-argument dot with sparse vectors
`dot(a,D,b)` with `a`, `b` sparse vector and D diagonal is slow, it specializes on Diagonal but without exploiting the sparsity of the vectors

matbesancon:
Closes #563

### DIFF
--- a/src/linalg.jl
+++ b/src/linalg.jl
@@ -630,12 +630,12 @@ function dot(A::AbstractSparseMatrixCSC, B::Union{DenseMatrixUnion,WrapperMatrix
 end
 
 function dot(x::AbstractSparseVector, Q::Diagonal, y::AbstractVector)
-    if length(x) != length(y)
+    d = Q.diag
+    if length(x) != length(y) || length(y) != length(d)
         throw(
-            DimensionMismatch("Vector x has a length $(length(x)) but y has a length $(length(y))")
+            DimensionMismatch("Vectors and matrix have different dimensions, x has a length $(length(x)), y has a length $(length(y)), Q has side dimension $(length(d))")
         )
     end
-    d = Q.diag
     nzvals = nonzeros(x)
     nzinds = nonzeroinds(x)
     s = zero(Base.promote_eltype(x, Q, y))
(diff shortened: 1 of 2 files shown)


src/linalg.jl
```diff
@@ -629,6 +629,61 @@ function dot(A::AbstractSparseMatrixCSC, B::Union{DenseMatrixUnion,WrapperMatrix
     return conj(dot(B, A))
 end
 
+function dot(x::AbstractSparseVector, Q::Diagonal, y::AbstractVector)
+    if length(x) != length(y)
```
dkarrasch, line 633: I think you also need to check the length of `Q.diag`. Perhaps you could replace `Q` with `D` and continue to use `d = D.diag`?

src/linalg.jl
```diff
@@ -629,6 +629,61 @@ function dot(A::AbstractSparseMatrixCSC, B::Union{DenseMatrixUnion,WrapperMatrix
...
+        throw(
+            DimensionMismatch("Vector x has a length $(length(x)) but y has a length $(length(y))")
+        )
+    end
+    d = Q.diag
+    nzvals = nonzeros(x)
+    nzinds = nonzeroinds(x)
+    s = zero(Base.promote_eltype(x, Q, y))
```
dkarrasch, line 641: In `LinearAlgebra`, we use

```julia
s = zero(typeof(dot(first(x), first(A), first(y))))
```

because `dot` acts recursively and technically, arrays of arrays are allowed.

dkarrasch (changes requested):
Nice. The comments refer to both methods actually. As for the length checks, perhaps you could add tests for all mismatching cases?

src/linalg.jl
```diff
@@ -629,6 +629,61 @@ function dot(A::AbstractSparseMatrixCSC, B::Union{DenseMatrixUnion,WrapperMatrix
     return conj(dot(B, A))
 end
 
+function dot(x::AbstractSparseVector, Q::Diagonal, y::AbstractVector)
+    d = Q.diag
+    if length(x) != length(y) || length(y) != length(d)
+        throw(
+            DimensionMismatch("Vectors and matrix have different dimensions, x has a length $(length(x)), y has a length $(length(y)), Q has side dimension $(size(Q, 1))")
+        )
+    end
```
dkarrasch, line 638:
```suggestion
    length(x) == length(y)  == length(d) ||
        throw(
            DimensionMismatch("Vectors and matrix have different dimensions, x has a length $(length(x)), y has a length $(length(y)), Q has side dimension $(size(Q, 1))")
        )
```
?

Same below, since you're not reusing `n = length(x)` there.

src/linalg.jl
matbesancon: both done thank you!

src/linalg.jl
matbesancon: done!

src/linalg.jl
matbesancon: just changed

src/linalg.jl
```diff
@@ -629,6 +629,60 @@ function dot(A::AbstractSparseMatrixCSC, B::Union{DenseMatrixUnion,WrapperMatrix
     return conj(dot(B, A))
 end
 
+function dot(x::AbstractSparseVector, D::Diagonal, y::AbstractVector)
```
dkarrasch, line 632: Do you want to add a specialization for when `y` is sparse? Like

```julia
dot(x::AbstractVector, D::Diagonal, y::AbstractSparseVector) = adjoint(dot(y, D', x))
```

? Though this may introduce an infinite loop when `D'` returns an `Adjoint`. Or write it out?

matbesancon: > Though this may introduce an infinite loop when D' returns an Adjoint. Or write it out?

It shouldn't no, D' returns a diagonal without wrapping an Adjoint type on top

dkarrasch: But for arrays of arrays it wraps an `Adjoint` around, and 3-arg `dot` with a middle `Adjoint` unwraps the adjoint again. I'm never sure anyone uses array of arrays "in production", but technically one could, and `dot` highlights that it acts recursively in such situations. It's just a few lines of code, so I think it's safer to spell it out.

matbesancon: > But for arrays of arrays it wraps an Adjoint around, and 3-arg dot with a middle Adjoint unwraps the adjoint again

I don't get the issue, if `D'` is always guaranteed not to return an adjoint?

`adjoint(D::Diagonal) = Diagonal(adjoint.(D.diag))` is the only potentially implementation for recursive arrays, and it still returns a diagonal, it could potentially wrap around an adjoint diagonal but this doesn't trigger a recursive loop?

dkarrasch: You're right, sorry, I misread it (or misremembered).

dkarrasch (approved):

Issue #548, opened 2024-07-30
CHOLMOD default ordering options: METIS vs AMD


gerw:
Consider the following snippet
```julia
using SparseArrays
using LinearAlgebra

function setup(n)
	e = ones(n)
	e2 = -ones(n-1)
	K = spdiagm(n, n, 0 => 2.0e, 1 => e2, -1 => e2)
	Id = sparse(1.0*I, n, n)
	return kron(K, Id) + kron(Id, K)
end

function solve(n)
	A = setup(n)
	F = ones(n^2)
	@time A \ F
end
```
It just builds a discretization of a PDE and solves it. Now let us execute
```julia
solve(1156);
solve(1157);
```
Using `julia 1.9.3` I get
```
  2.095628 seconds (66 allocations: 1.339 GiB, 0.08% gc time)
  2.134111 seconds (66 allocations: 1.339 GiB, 0.08% gc time)
```
and all is fine. With `julia 1.10.4` instead I get
```
  1.966967 seconds (69 allocations: 1.339 GiB, 0.17% gc time)
  7.775528 seconds (1.99 M allocations: 9.856 GiB, 2.97% gc time)
```
Thus, the second solve on a slightly larger system is suddenly much slower and allocates memory. The profiling shows that `cholmod_l_analyze` runs much longer. The same happens with the RC of 1.11.

Digging a little bit deeper with
```
A = setup(1157);
B = SparseArrays.CHOLMOD.Sparse(A);
@time F = SparseArrays.CHOLMOD.symbolic(B, perm=nothing);
@time cholesky!(F, A; shift = 1., check = true);
```
and looking at the output of `top`, one can see that `symbolic` runs only on a single core and allocates memory, `cholesky!` seems to be fine (and seems to run in parallel).

I also tried to fiddle around with
```
SparseArrays.CHOLMOD.getcommon()[].nthreads_max
```
(which is `1`) after startup, but it does not seem to change anything.

ViralBShah:
It seems the first order of business should be to see what is driving those allocations and fix that.

gerw:
Can you reproduce this behaviour?

If I see it correctly, `symbolic` just boils down to a ccall to `cholmod_l_analyze`...

ViralBShah:
Yes, I see the same thing with 1.10.4.

ViralBShah:
@KristofferC Any ideas?

KristofferC:
[...]

PR #572, opened 2024-10-21
Backports to v1.10

dkarrasch:


dkarrasch:
Do we have more to add here? Otherwise, we should merge this and then bump it into https://github.com/JuliaLang/julia/pull/56381.

ViralBShah:
I think we want #560

ViralBShah:
Should we update the CI on this branch to run on 1.10?

dkarrasch:
Oh, I forgot about our label, so there's a lot marked for backport, actually. I'll go through the list. CI seems to be running on v1.10?

I can't backport #550. It has quite some heavy merge conflicts that I don't know how to resolve.

dkarrasch:
> There are several others marked as backport to 1.10, but many probably have already been backported. Perhaps we should remove the labels?

I have cleaned up the list. Everything (up to #550) is included and I removed the labels.

ViralBShah:
> I have cleaned up the list. Everything (up to #550) is included and I removed the labels.

Thank you!

ViralBShah:
I isolated the commit for disabling nested dissection and pushed it. Unfortunately, instead of a branch I pushed directly to release-1.10. Also updated the CI. I might need to push a few more to make sure it all ends up working out.

ViralBShah:
Would it be ok to remove those commits and force push on the 1.10 branch? I am creating a new PR against the `backports-release-1.10` branch.

ViralBShah:
cc @KristofferC

ViralBShah:
Verified that the issue reported in #548 is fixed on this branch.

dkarrasch:
This seems like done. I'm merging this.

PR #575, opened 2024-10-25
Fix keytype and zero #574

### LINKED ISSUE #574: zero & keytype always return Int64 for the index type
```julia
julia> using SparseArrays

julia> v=spzeros(Int,Int32,10)
10-element SparseVector{Int64, Int32} with 0 stored entries

julia> keytype(v)
Int64

julia> zero(v)
10-element SparseVector{Int64, Int64} with 0 stored entries
```

dpinol:
Fixes https://github.com/JuliaSparse/SparseArrays.jl/issues/574

### DIFF
--- a/src/SparseArrays.jl
+++ b/src/SparseArrays.jl
@@ -22,7 +22,7 @@ import LinearAlgebra: mul!, ldiv!, rdiv!, cholesky, adjoint!, diag, eigen, dot,
 
 import Base: adjoint, argmin, argmax, Array, broadcast, circshift!, complex, Complex,
     conj, conj!, convert, copy, copy!, copyto!, count, diff, findall, findmax, findmin,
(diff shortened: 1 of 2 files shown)


fredrikekre:
`keytype` isn't the type of the stored indices though, it should match `eltype(keys(...))`:
```
julia> v = spzeros(Int, Int32, 10)
10-element SparseVector{Int64, Int32} with 0 stored entries

julia> eltype(keys(v))
Int64
```

fredrikekre:
No, I think `keys` is correct, and thus think that `keytype` should always be `Int`.

dpinol:
> No, I think `keys` is correct, and thus think that `keytype` should always be `Int`.

why?

ViralBShah:
What should do further here?

ViralBShah:
@fredrikekre Should this be closed?

araujoms:
Closing in favour of #710

PR #576, opened 2024-10-26
doc: move solvers doc to `src\solvers.md`

inkydragon:
When building julia with `USE_GPL_LIBS:=0`, the documentation builds fails because the documentation for the solver cannot be found.

Only `docs/src/index.md` is copied when building the documentation, so we can move the solver documentation to `docs/src/solver.md` to avoid this problem.

See: https://github.com/JuliaLang/julia/issues/56344


ViralBShah:
@dkarrasch It would be good to pull this into the 1.10 PR you already have.

PR #583, opened 2024-11-18
Make `allowscalar` a macro with auto-world-age-increment

Keno:
Goes along with https://github.com/JuliaLang/julia/pull/56509


ViralBShah:
Does this need to be backported? I assume not.

Keno:
No

PR #592, opened 2025-01-09
More adhereance to 1.10 error types (hopefully)

wsmoses:
It looks like in base that error was only enabled on number inputs, lets see if this fixes it?

cc @gbaraldi @KristofferC


wsmoses:
okay @dkarrasch would you mind merging this? According to @gbaraldi on a local test this fixed the issue in https://github.com/JuliaLang/julia/pull/56653#issuecomment-2579358102 blocking the 1.10.8 release

PR #595, opened 2025-01-12
Solve with `AdjointFactorization` of Cholmod and `Adjoint` rhs

### LINKED ISSUE #594: Cannot solve with `AdjointFactorization` of Cholesky factorizations and `Adjoint` right-hand side
Consider the following MWE:

```julia

n = 10
A = sprand(n, n, 1/n)
A = A + A' + 10I

B = rand(n, 2)
Bt = Matrix(B')

F = factorize(A')
F \ B   # works just fine
F \ Bt' # breaks
```

```
ERROR: MethodError: no method matching ldiv!(::LinearAlgebra.AdjointFactorization{Float64, SparseArrays.CHOLMOD.Factor{Float64, Int64}}, ::Matrix{Float64})

Closest candidates are:
  ldiv!(::BunchKaufman{T, <:StridedMatrix{T} where T}, ::StridedVecOrMat{T}) where T<:Union{Float32, Float64}
   @ LinearAlgebra ~/.julia/juliaup/julia-1.10.5+0.x64.linux.gnu/share/julia/stdlib/v1.10/LinearAlgebra/src/bunchkaufman.jl:329
  ldiv!(::LinearAlgebra.QRCompactWY{T, M, C} where {M<:AbstractMatrix{T}, C<:AbstractMatrix{T}}, ::AbstractMatrix{T}) where T
   @ LinearAlgebra ~/.julia/juliaup/julia-1.10.5+0.x64.linux.gnu/share/julia/stdlib/v1.10/LinearAlgebra/src/qr.jl:523
  ldiv!(::UpperHessenberg, ::AbstractVecOrMat; shift)
   @ LinearAlgebra ~/.julia/juliaup/julia-1.10.5+0.x64.linux.gnu/share/julia/stdlib/v1.10/LinearAlgebra/src/hessenberg.jl:179
  ...

Stacktrace:
 [1] \(F::LinearAlgebra.AdjointFactorization{Float64, SparseArrays.CHOLMOD.Factor{Float64, Int64}}, B::Adjoint{Float64, Matrix{Float64}})
   @ LinearAlgebra ~/.julia/juliaup/julia-1.10.5+0.x64.linux.gnu/share/julia/stdlib/v1.10/LinearAlgebra/src/factorization.jl:143
 [2] top-level scope
   @ REPL[27]:1
 [3] top-level scope
   @ none:1
```

A similar situation happens in [one of my codes](https://github.com/mpimd-csc/DifferentialRiccatiEquations.jl/blob/35f357021be677f0ea9bd31213e3cccc1ea01c94/src/LowRankUpdate.jl#L79) that implements the Sherman-Morrison-Woodbury formula. What is the proper way to fix this?

1. Define `adjoint(F::SparseArrays.CHOLMOD.Factor) = F`? 
2. Make `Adjoint{T, Matrix{T}}` a `StridedMatrix`?
3. Define the missing method of `ldiv!`?
4. ...?

Option 1 would work, but seems out of place given [this method of `(\)`](https://github.com/JuliaSparse/SparseArrays.jl/blob/248d9f94cf31d9668e5b07f237f13a239472b2ec/src/solvers/cholmod.jl#L1716-L1719) (permalink created from `release-1.10`). I didn't test options 2 or 3. What would be the best option to have this work with the current LTS?

dkarrasch:
Fixes #594.

### DIFF
--- a/src/solvers/cholmod.jl
+++ b/src/solvers/cholmod.jl
@@ -15,7 +15,7 @@ import Base: (*), convert, copy, eltype, getindex, getproperty, show, size,
 using Base: require_one_based_indexing
 
 using LinearAlgebra
-using LinearAlgebra: RealHermSymComplexHerm, AdjOrTrans
+using LinearAlgebra: RealHermSymComplexHerm, AdjOrTrans, AdjOrTransStridedMat
 import LinearAlgebra: (\), AdjointFactorization,
                  cholesky, cholesky!, det, diag, ishermitian, isposdef,
                  issuccess, issymmetric, ldiv!, ldlt, ldlt!, logdet,
@@ -1928,6 +1928,7 @@ function \(adjL::AdjointFactorization{<:VTypes,<:Factor}, B::StridedMatrix)
     L = adjL.parent
     return Matrix(solve(CHOLMOD_A, L, Dense(B)))
 end
+(\)(adjL::AdjointFactorization{<:VTypes,<:Factor}, B::LinearAlgebra.AdjOrTransStridedMat) = adjL \ copy(B)
 
 const RealHermSymComplexHermSSL{Ti, Tr} = Union{
     Symmetric{Tr, SparseMatrixCSC{Tr, Ti}},
--- a/src/solvers/cholmod.jl
+++ b/src/solvers/cholmod.jl
@@ -15,7 +15,7 @@ import Base: (*), convert, copy, eltype, getindex, getproperty, show, size,
 using Base: require_one_based_indexing
 
 using LinearAlgebra
-using LinearAlgebra: RealHermSymComplexHerm, AdjOrTrans, AdjOrTransStridedMat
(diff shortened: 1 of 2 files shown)


jonas-schulze (commented):
Thanks for the swift solution! I wonder why the new method of `(\)` is not covered by the tests. :eyes: 

Could this be backported to Julia 1.10, please?

ViralBShah (approved):

dkarrasch:
> Could this be backported to Julia 1.10, please?

I'm not sure about that. @ViralBShah @KristofferC This was failing due to a "missing method". I guess this counts more like a "new feature" rather than "bugfix".

@jonas-schulze If we decide against backporting, you could commit type piracy in your package and define this one method in a version-dependent branch...? I know it's not optimal, but I personally have used all my credits (if I ever had some) for arguing in favor of a semver-breaking action. :sweat_smile:

ViralBShah:
I think it is ok to backport it, but @KristofferC should be the person to ask.

Issue #585, opened 2024-11-20
Broken dtype detection in CHOLMOD in 1.11


pawel-tarasiuk-quantumz:
Computing `qr` factorization on `SparseMatrixCSC{Float32}` causes an error with Julia 1.11:

```Julia
using SparseArrays, LinearAlgebra
A = sprandn(Float32, 10, 10, 0.5)
qr(Float64.(A))  # works proprely
qr(A)   # !!!
```

Message:
```
ERROR: SparseArrays.CHOLMOD.CHOLMODException("dtype=0 not supported")
Stacktrace:
 [1] SparseArrays.CHOLMOD.Dense{Float32}(ptr::Ptr{SparseArrays.LibSuiteSparse.cholmod_dense_struct})
   @ SparseArrays.CHOLMOD ~/.julia/juliaup/julia-1.11.1+0.x64.linux.gnu/share/julia/stdlib/v1.11/SparseArrays/src/solvers/cholmod.jl:276
 [2] qr(A::SparseMatrixCSC{Float32, Int64}; tol::Float32, ordering::Int32)
   @ SparseArrays.SPQR ~/.julia/juliaup/julia-1.11.1+0.x64.linux.gnu/share/julia/stdlib/v1.11/SparseArrays/src/solvers/spqr.jl:209
 [3] qr(A::SparseMatrixCSC{Float32, Int64})
   @ SparseArrays.SPQR ~/.julia/juliaup/julia-1.11.1+0.x64.linux.gnu/share/julia/stdlib/v1.11/SparseArrays/src/solvers/spqr.jl:195
 [4] top-level scope
   @ REPL[4]:1
```

Tested on clean Julia 1.11 setup with juliaup on Ubuntu 22.04, x86_64. Eqivalent setup with Julia 1.10 works for either element type.

It would seem that https://github.com/JuliaSparse/SparseArrays.jl/blob/1b4933ccc7b1f97427ff88bd7ba58950021f2c60/src/solvers/spqr.jl#L200 does not indicate the proper `dtype`, as `unsafe_load(HTau[]).dtype` is `0` when run for either type. This is incidentally equal to `CHOLMOD_DOUBLE`, which allows the computation for `Float64`.

pawel-tarasiuk-quantumz:
Since `dtype` mismatch is possible, it might be worth to support varied results with no assumptions, as proposed in https://github.com/JuliaSparse/SparseArrays.jl/pull/586.

[...]

PR #597, opened 2025-02-07
cleanup support for Float32 and ComplexF32

araujoms:
I fixed the same bug as #585, without realizing that it had already been done. It wasn't a complete waste of time, because the PR #586 left some cleanup to do: the lines
```julia
LinearAlgebra.qr(A::SparseMatrixCSC{<:Union{Float16,Float32}}; tol=_default_tol(A)) =
    qr(convert(SparseMatrixCSC{Float64}, A); tol=tol)
LinearAlgebra.qr(A::SparseMatrixCSC{<:Union{ComplexF16,ComplexF32}}; tol=_default_tol(A)) =
    qr(convert(SparseMatrixCSC{ComplexF64}, A); tol=tol)
```
would have prevented it from working if they actually worked! Luckily they don't, as the correct method is more specific and gets called, but it's still worth fixing. I also changed them to convert `Float16` to `Float32` instead of `Float64`, and the analogous change for complex numbers.

I also fixed the error messages one gets. They were of the form "dtype=$(s.dtype) not supported", which in my case translated to saying that `Float64` wasn't supported when I tried calling it with `Float32`. `s.dtype` is the type that actually got computed, so it clearly is supported. I changed them to say "dtype=$(dtyp(Tv)) not supported".

### DIFF
--- a/src/solvers/cholmod.jl
+++ b/src/solvers/cholmod.jl
@@ -270,10 +270,10 @@ mutable struct Dense{Tv<:VTypes} <: DenseMatrix{Tv}
         s = unsafe_load(ptr)
         if s.xtype != xtyp(Tv)
             free!(ptr)
-            throw(CHOLMODException("xtype=$(s.xtype) not supported"))
+            throw(CHOLMODException("xtype=$(xtyp(Tv)) not supported"))
         elseif s.dtype != dtyp(Tv)
             free!(ptr)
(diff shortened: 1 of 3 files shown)


ViralBShah (approved):

ViralBShah:
LGTM. Let's give it a couple of days for comments.

PR #598, opened 2025-02-08
Explicitly declare type constructor imports

Keno:
These would otherwise give a warning after JuliaLang/julia#57311, but are a good change even independently.

### DIFF
--- a/src/SparseArrays.jl
+++ b/src/SparseArrays.jl
@@ -15,6 +15,7 @@ using LinearAlgebra: AdjOrTrans, AdjointFactorization, TransposeFactorization, m
 
 
 import Base: +, -, *, \, /, ==, zero
+import Base: Matrix, Vector
 import LinearAlgebra: mul!, ldiv!, rdiv!, cholesky, adjoint!, diag, eigen, dot,
     issymmetric, istril, istriu, lu, tr, transpose!, tril!, triu!, isbanded,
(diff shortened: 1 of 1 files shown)


ViralBShah:
This PR is included in SparseArrays on julia master, but it still seems to get the warnings

```
┌ SparseArrays
│  WARNING: Constructor for type "Matrix" was extended in `CHOLMOD` without explicit qualification or import.
│    NOTE: Assumed "Matrix" refers to `Base.Matrix`. This behavior is deprecated and may differ in future versions.`
│    NOTE: This behavior may have differed in Julia versions prior to 1.12.
│    Hint: If you intended to create a new generic function of the same name, use `function Matrix end`.
│    Hint: To silence the warning, qualify `Matrix` as `Base.Matrix` or explicitly `import Base: Matrix`
│  WARNING: Constructor for type "Vector" was extended in `CHOLMOD` without explicit qualification or import.
│    NOTE: Assumed "Vector" refers to `Base.Vector`. This behavior is deprecated and may differ in future versions.`
│    NOTE: This behavior may have differed in Julia versions prior to 1.12.
│    Hint: If you intended to create a new generic function of the same name, use `function Vector end`.
│    Hint: To silence the warning, qualify `Vector` as `Base.Vector` or explicitly `import Base: Vector`
└  
```

Issue #609, opened 2025-03-16
Usage of `_iszero`, `_isnotzero`


OlivierHnt:
We are currently working on releasing the 1.0 version of [IntervalArithmetic.jl](https://github.com/JuliaIntervals/IntervalArithmetic.jl) (see https://github.com/JuliaIntervals/IntervalArithmetic.jl/pull/698). One important decision we made is to be very strict with the usage of boolean comparison functions from Base. For instance, `iszero(x::Interval)` throws if `x` contains zero but is not thin, like $[-1, 1]$.

I noticed that you have two internal functions `_iszero` and `_isnotzero` which seems to be used only in the context "is the number exactly zero".

I found (on 1.11, but maybe also true on 1.10) that this is not used everywhere, e.g.

https://github.com/JuliaSparse/SparseArrays.jl/blob/242035184c0d539bdb5e64bf26eb7726b123db14/src/sparsevector.jl#L2384C8-L2384C21

Is it possible to replace all instances of `x == zero(x)` (and similar) with `_iszero` (or `_isnotzero`) ? If so, could this be backported to 1.10 and 1.11 as well?

ViralBShah:
Should be ok in principle. Are you able to put a PR?

SobhanMP:
I think it's probably worth reading the comments in https://github.com/JuliaSparse/SparseArrays.jl/pull/296, in short beware that -0 is a bit weird.

OlivierHnt:
Thx for the heads up 🙂. I'll make a PR, we can discuss the backporting procedure then.

LilithHafner:
I expect that you'll need to return `missing` or `IntervalArithmatic.Unknown()` or some such or declare a weak dependency on SpareArrays because `iszero(x) === true` still throws if `iszero(x)` throws.

PR #610, opened 2025-03-17
Replace `v == zero(v)` with  `_iszero`

OlivierHnt:
The purpose of this PR is to not call `==` directly, and use `_iszero` instead. This is to help integration with [IntervalArithmetic.jl](https://github.com/JuliaIntervals/IntervalArithmetic.jl) since `==` for our `Interval` type does not always return a Boolean.

Closes https://github.com/JuliaSparse/SparseArrays.jl/issues/609.

I did not change two checks using `===` since this would break the behaviour for `-0.0`. If I understood correctly, it should not be an issue with `Interval`, since it will return `false` consistently (it may be sub-optimal, but not wrong)
cc @dpsanders, @Kolaru
cc also @kalmarek, since you raised https://github.com/JuliaIntervals/IntervalArithmetic.jl/issues/628 you may have more experience mixing SparseArrays and IntervalArithmetic

I do not know how to the backporting procedure works, but it'd be great to have this for 1.10 and 1.11.

### DIFF
--- a/src/sparsematrix.jl
+++ b/src/sparsematrix.jl
@@ -2283,7 +2283,7 @@ function (+)(A::Array, B::SparseMatrixCSCUnion)
     for j in axes(B,2)
         for i in nzrange(B, j)
             rowidx = rowinds[i]
-            C[rowidx,j] = A[rowidx,j] + nzvals[i] 
+            C[rowidx,j] = A[rowidx,j] + nzvals[i]
         end
     end
     return C
@@ -2452,7 +2452,7 @@ function Base._mapreduce(f, op::Union{typeof(Base.mul_prod),typeof(*)}, ::Base.I
     else
         v = f(zero(T))^(nzeros)
         # Bail out early if initial reduction value is zero or if there are no stored elements
-        (v == zero(T) || nnzA == 0) ? v : v*Base._mapreduce(f, op, nzvalview(A))
+        (_iszero(v) || nnzA == 0) ? v : v*Base._mapreduce(f, op, nzvalview(A))
     end
 end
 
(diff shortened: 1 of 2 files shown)


ViralBShah:
@SobhanMP

LilithHafner:
> I did not change two checks using `===` since this would break the behaviour for `-0.0`. If I understood correctly, it should not be an issue with `Interval`, since it will return `false` consistently (it may be sub-optimal, but not wrong)

You're right not to change those checks and why. They are actually necessary for correctness in the case of Intervals, too! For example, with `_iszero`, if you set an entry to be an ungarunteed zero interval and then accessed that entry it would have the isgarunteed flag incorrectly set. (Also, the `===` check will return true sometimes (e.g. on `I"0"`)).

I'm happy to merge this PR but hesitant to mark it for back porting because of how many edge cases sparse arrays tend to have. I'm not opposed to someone else backporting if they are more confidant than I that there won't be unintended consequences.

OlivierHnt:
@LilithHafner 

Thanks for reviewing the PR. Indeed we must be cautious with such changes.
My argument in favour of backporting this PR is that the `_iszero` and `_isnotzero` mechanism is already implemented in 1.10 and 1.11. So backporting this PR is really just about merging the same changes in the same functions for 1.10 and 1.11 (if these changes are safe here, they are safe to backport).
See e.g. on 1.11
https://github.com/JuliaSparse/SparseArrays.jl/blob/242035184c0d539bdb5e64bf26eb7726b123db14/src/sparsematrix.jl#L2443

I might be overlooking something, but it seems that these 6 instances of using `==` to check whether the value is zero were unintentionally left out when `_iszero` and `_isnotzero` were introduced.

LilithHafner:
For folks considering backporting, the semantic changes here are, AFAICT:
- Change error to false in the case of non-`true` result from `== zero`
- Enable user specialization on `iszero` (previously we bypassed that by using `== zero` directly)

The latter could theoretically break things, though that seems implausible. I'm not opposed to another maintainer choosing to backport.

SobhanMP:
@LilithHafner Thanks for reviewing this!

PR #617, opened 2025-04-24
Delete lock interface implementation from UmfpackLU

fredrikekre:


ViralBShah (approved):

SobhanMP:
I'm not critical of the patch, but why? IMHO, it's fair to warn people that they are waiting for locks to do sparse operations.

fredrikekre:
The motivation wasn't really to remove the warning but rather that it is kind of weird that a struct `UmfpackLU` is lockable, it isn't an `AbstractLock` for example.

Regarding the warning, since it isn't really actionable by the user, and safe to ignore, it seems kind of pointless?

SobhanMP:
> The motivation wasn't really to remove the warning but rather that it is kind of weird that a struct UmfpackLU is lockable, it isn't an AbstractLock for example.

Makes sense

> Regarding the warning, since it isn't really actionable by the user, and safe to ignore, it seems kind of pointless?

It is tho, you can pass a different workspace.

KristofferC (approved):
I agree with both things here. It is weird to implement these lock methods for `UmfpackLU`, if anything a user should probably use `Lockable` and for the warning message it isn't something that is typically done in Julia libraries.

PR #621, opened 2025-05-08
Clarify pros, cons and limitations of Cholesky and LDLt

danielwe:
I find it worth pointing out explicitly in the docs that LDLt, which mathematically looks like a drop-in replacement for Cholesky that does away with the positive definiteness requirement, comes with the following caveats:

* It fails for a lot of matrices (for example, `ldlt(Symmetric(sprandn(1000, 1000, p)))` basically never succeeds for any relevant sparsity `p`) due to the requirement that all leading principal minors be well-conditioned
* In CHOLMOD, `ldlt` is significantly slower than `cholesky` as it does not have a supernodal implementation

So I made some docstring edits to clarify the relationship and tradeoffs between `cholesky` and `ldlt`.

Citation for these claims: pages 106-107 in the CHOLMOD user guide at https://github.com/DrTimothyAldenDavis/SuiteSparse/blob/v7.10.3/CHOLMOD/Doc/CHOLMOD_UserGuide.pdf

### DIFF
--- a/src/solvers/cholmod.jl
+++ b/src/solvers/cholmod.jl
@@ -1563,6 +1563,9 @@ Setting the optional `shift` keyword argument computes the factorization of
 it should be a permutation of `1:size(A,1)` giving the ordering to use
 (instead of CHOLMOD's default AMD ordering).
 
+See also [`ldlt`](@ref) for a similar factorization that does not require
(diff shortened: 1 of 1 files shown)


dkarrasch (approved):
Seems concise and helpful.

PR #622, opened 2025-05-08
Update some wording in the docstring of cholmod.jl

WalterMadelim:
https://discourse.julialang.org/t/what-is-a-recommended-way-to-solve-a-linear-equation/128420/31?u=waltermadelim

### DIFF
--- a/src/solvers/cholmod.jl
+++ b/src/solvers/cholmod.jl
@@ -1504,8 +1504,8 @@ end
 
 Compute the Cholesky (``LL'``) factorization of `A`, reusing the symbolic
 factorization `F`. `A` must be a [`SparseMatrixCSC`](@ref) or a [`Symmetric`](@ref)/
-[`Hermitian`](@ref) view of a `SparseMatrixCSC`. Note that even if `A` doesn't
-have the type tag, it must still be symmetric or Hermitian.
+[`Hermitian`](@ref) view of a `SparseMatrixCSC`. Note that if `A` doesn't
+have the type tag, itself must be symmetric or Hermitian.
 
 See also [`cholesky`](@ref).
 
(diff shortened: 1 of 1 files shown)


ViralBShah:
Thanks for the suggestions, but the original wording is better.

WalterMadelim:
Well, I think the existing docstring itself is a bit inconsistent.
https://github.com/JuliaSparse/SparseArrays.jl/blob/b225a853f12cc2aca1f55138d41776b0670a2808/src/solvers/cholmod.jl#L1538-L1545

Observing line 1539, the arg `A` is restricted to be a `SparseMatrixCSC`. But in line 1542 it suggests `A` can be a `Symmetric` view.
But essentially, were `A` a view, it will not be dispatched to this method, strictly speaking, (am I right?).
```julia
julia> A::SparseMatrixCSC
3×3 SparseMatrixCSC{Float64, Int64} with 7 stored entries:
 10.0        0.519584   0.169799
  0.519584  10.0         ⋅
  0.169799    ⋅        10.0

julia> tagged_A = Symmetric(A)
3×3 Symmetric{Float64, SparseMatrixCSC{Float64, Int64}}:
 10.0        0.519584   0.169799
  0.519584  10.0         ⋅
  0.169799    ⋅        10.0

julia> tagged_A::SparseMatrixCSC
ERROR: TypeError: in typeassert, expected SparseMatrixCSC, got a value of type Symmetric{Float64, SparseMatrixCSC{Float64, Int64}}
Stacktrace:
 [1] top-level scope
   @ REPL[28]:1
```
A larger issue concerns the top-level design of LinearAlgebra.
I think `LinearAlgebra` and `SparseArrays` are not 2 separate objects. From the standpoint of a user, he should feel some sort of consistency. A matrix can have numerous "advantageous" structures---Symmetric, Banded, Sparse, etc.
From a text book sense, a Banded (e.g. Tridiagonal) matrix is surely also a Sparse matrix.
But the `TriDiagonal` belongs to `LinearAlgebra.jl` whereas `SparseMatrixCSC` is in `SparseArrays.jl`.

In `LinearAlgebra.jl`, we even have `SymTriDiagonal`, which is both Symmetric and TriDiagonal (which can be readily recognized).
But in this package, it seems we have to use a `Symmetric` view, as the `tagged_A` above shows. Therefore `tagged_A::SparseMatrixCSC` will ERROR.

Here are some of my thoughts (far from constructive). I'm still learning these two packages (`LinearAlgebra` and `SparseArray`).  Because it's not purely about julia, it invoke 3rd party APIs. Maybe I need more time to have a deeper understanding.🙂

ViralBShah:
IIRC, we simply ignore the lower triangular part if not actually symmetric - but need to look at the code for that. In that case the documentation should be updated accordingly.

PR #620, opened 2025-05-06
Use `libsuitesparseconfig` from JLL

staticfloat:
This prevents us from being influenced by things like `LD_LIBRARY_PATH`, and ensures that we always load the correct `libsuitesparseconfig` that came with our JLL.

PR #625, opened 2025-05-17
fix libsuitesparseconfig bug

IanButterworth:
Fixes https://github.com/JuliaSparse/SparseArrays.jl/pull/620#issuecomment-2888024821


IanButterworth:
Added the same backport labels as https://github.com/JuliaSparse/SparseArrays.jl/pull/620

PR #626, opened 2025-05-17
Lazily init cholmod

IanButterworth:
On top of https://github.com/JuliaSparse/SparseArrays.jl/pull/625

### DIFF
--- a/src/solvers/cholmod.jl
+++ b/src/solvers/cholmod.jl
@@ -37,7 +37,7 @@ import SparseArrays: AbstractSparseMatrix, SparseMatrixCSC, indtype, sparse, spz
 import ..increment, ..increment!
 
 using ..LibSuiteSparse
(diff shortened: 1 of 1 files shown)


rayegun:
Hmm I wonder if we should build SuiteSparse against libjulia for the `jl_<GC>` functions in Ygg. This will now no longer use `jl_malloc` for a number of JLLs including Sundials as well as any of the other SuiteSparse packages

ViralBShah:
That would be a good idea!

rayegun:
My understanding is that `__init__` would be called by Julia always at load time, therefore cholmod will be ready for any downstream JLLs. Whereas now you must call `getcommon` which may not occur until you actually use cholmod. This is really something we should stick in a prologue on the JLL I think?

IanButterworth:
I think additionally, I did wonder if this change is safe for deserialized cholmod objects?

PR #637, opened 2025-07-19
use index_shape_nested

adienes:
would be paired with https://github.com/JuliaLang/julia/pull/59025 to make `setindex!` a bit more coherent


adienes:
I think this one should go first, that way https://github.com/JuliaLang/julia/pull/59025 can bump the hash as part of the same commit so there are no commits in `JuliaLang/julia` with an invalid SparseArrays

adienes:
I intend to finish up the plan within a week or two

PR #638, opened 2025-07-24
Overwrite fewer methods in ReadOnly

tiemvanderdeure:
I think there are lots of methods in this file that can be deleted - but I don't totally understand what ReadOnly is used for and took the conservative approach and deleted only the methods that snoop_invalidations tell me cause invalidations. Which are `eachindex` and `eltype`

`Base.resize!` also invalidates but that one at least is actually used.


oscardssmith:
I'm probably not the right person to review this since I don't know what ReadOnly is.

PR #640, opened 2025-08-08
fast quadratic form for dense matrix, sparse vectors

matbesancon:
This was missing from the specializations and thus still slow.
This is a port from the same PR on FW.jl: https://github.com/ZIB-IOL/FrankWolfe.jl/pull/607

### DIFF
--- a/src/linalg.jl
+++ b/src/linalg.jl
@@ -725,7 +725,7 @@ end
 
 function dot(
     a::AbstractSparseVector,
-    Q::Union{DenseMatrixUnion,WrapperMatrixTypes{<:Real,<:DenseMatrixUnion}},
+    Q::Union{DenseMatrixUnion,WrapperMatrixTypes{<:Any,<:DenseMatrixUnion}},
     b::AbstractSparseVector,
 )
     return _dot_quadratic_form(a, Q, b)
--- a/src/linalg.jl
+++ b/src/linalg.jl
@@ -728,6 +728,30 @@ function dot(
     Q::Union{DenseMatrixUnion,WrapperMatrixTypes{<:Any,DenseMatrixUnion}},
     b::AbstractSparseVector,
 )
+    return _dot_quadratic_form(a, Q, b)
+end
+
+function dot(
+    a::AbstractSparseVector,
+    Q::LinearAlgebra.Transpose{<:Real, Union{StridedMatrix, BitMatrix}},
+    b::AbstractSparseVector,
+)
+    return _dot_quadratic_form(a, Q, b)
+end
+
+function dot(
+    a::AbstractSparseVector,
+    Q::Union{
+        LinearAlgebra.Hermitian{<:Real, Union{StridedMatrix, BitMatrix}},
+        LinearAlgebra.Hermitian{Complex{<:Real}, Union{StridedMatrix, BitMatrix}},
+        LinearAlgebra.Symmetric{<:Real, Union{StridedMatrix, BitMatrix}}
+    },
+    b::AbstractSparseVector)
+    return _dot_quadratic_form(a, Q, b)
+end
+
+# actual function implementation called by the method dispatch
+function _dot_quadratic_form(a, Q, b)
     n = length(a)
     m = length(b)
     if size(Q) != (n, m)
--- a/src/linalg.jl
+++ b/src/linalg.jl
@@ -725,7 +725,7 @@ end
 
 function dot(
     a::AbstractSparseVector,
-    Q::Union{DenseMatrixUnion,WrapperMatrixTypes{<:Real,DenseMatrixUnion}},
+    Q::AbstractMatrix{<:Number},
     b::AbstractSparseVector,
 )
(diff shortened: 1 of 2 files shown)


src/linalg.jl
```diff
@@ -723,6 +723,69 @@ function dot(x::AbstractSparseVector, D::Diagonal, y::AbstractSparseVector)
...
+    a::AbstractSparseVector,
+    Q::Union{DenseMatrixUnion,WrapperMatrixTypes{<:Real,DenseMatrixUnion}},
+    b::AbstractSparseVector,
+)
+    return _dot_quadratic_form(a, Q, b)
+end
+
+function dot(
+    a::AbstractSparseVector,
+    Q::LinearAlgebra.Transpose{<:Real, DenseMatrixUnion},
```
dkarrasch, line 736: This case is included in the previous one. Is it needed for disambiguation?

src/linalg.jl
```diff
@@ -723,6 +723,69 @@ function dot(x::AbstractSparseVector, D::Diagonal, y::AbstractSparseVector)
...
+function dot(
+    a::AbstractSparseVector,
+    Q::LinearAlgebra.RealHermSymComplexHerm{<:Real, DenseMatrixUnion},
+    b::AbstractSparseVector)
+    return _dot_quadratic_form(a, Q, b)
+end
+
+function dot(
+    a::AbstractSparseVector,
+    Q::LinearAlgebra.Hermitian{<:Real, DenseMatrixUnion},
```
dkarrasch, line 759: Same here.

dkarrasch (commented):
The doubly layered case (transpose of wrapper of dense) begs the question if that really occurs, and if yes, then what about the adjoint of wrappers of dense? Conversely, would it be too generic to have

```julia
function dot(a::AbstractSparseVector, Q::AbstractMatrix, b::AbstractSparseVector)
    return _dot_quadratic_form(a, Q, b)
end
```

matbesancon:
> begs the question if that really occurs

that's the problem with method ambiguities, there is no way to prove Aqua that this is never the case

matbesancon:
About the suggestion of

`dot(a::AbstractSparseVector, Q::AbstractMatrix, b::AbstractSparseVector)`

The issue is precisely
`dot(x::AbstractVector{T1}, A::AbstractSparseMatrixCSC{T2}, y::AbstractVector{T3}) where {T1,T2,T3}`

because this needs to be disambiguated again

matbesancon:
Making it general was making a whole lot of ambiguities pop:
https://github.com/JuliaSparse/SparseArrays.jl/actions/runs/16826540584/job/47664254978?pr=640#step:6:118

matbesancon:
ok all settled on my end, @dkarrasch can you take a look to see if things are all good for you?

matbesancon:
Specifically this function:
https://github.com/JuliaSparse/SparseArrays.jl/blob/main/src/linalg.jl#L1245

matbesancon:
The test will fail if we haven't overwritten the function

dkarrasch:
A fix is coming in  #643. Let's merge that one first, then update here and make sure we don't end up with ambiguities.

matbesancon:
The tests are passing, good to merge I'd say

test/linalg.jl
```diff
@@ -952,6 +952,13 @@ end
         y = sprand(ComplexF64, 15, 0.5)
         @test dot(x, A, y) ≈ dot(Vector(x), A, Vector(y)) ≈ (Vector(x)' * Matrix(A)) * Vector(y)
         @test dot(x, A, y) ≈ dot(x, Av, y)
+        @test dot(x, collect(A), y) ≈ dot(x, A, y)
+        @test dot(y, collect(A)', x) ≈ dot(y, A', x)
+        @test dot(y, transpose(collect(A)), x) ≈ dot(y, transpose(A), x)
+        @test dot(y, Hermitian(collect(A)' * collect(A)), y) ≈ dot(y, Hermitian(A' * A), y)
+        @test dot(y, Symmetric(collect(A)' * collect(A)), y) ≈ dot(y, Symmetric(A' * A), y)
+        B = BitMatrix(rand(Bool, 10, 15))
+        @test dot(x, A, y) ≈ dot(x, Matrix(A), y)
```
dkarrasch, line 961:
```suggestion
        @test dot(x, A, y) ≈ dot(x, Matrix(A), y)
        @test_throws DimensionMismatch dot([x, x], A, y)
        @test_throws DimensionMismatch dot(x, A, [y, y])
        @test iszero(dot(spzeros(length(x)), A, y)
```

test/linalg.jl
```diff
@@ -952,6 +952,16 @@ end
...
+        @test_throws DimensionMismatch dot([x, x], A, y)
+        @test_throws DimensionMismatch dot(x, A, [y, y])
+        @test iszero(dot(spzeros(length(x)), A, y)
```
dkarrasch, line 964:
```suggestion
        @test iszero(dot(spzeros(length(x)), A, y))
```

PR #645, opened 2025-08-27
Improve CHOLMOD friendliness

danielwe:
Follow-up from a discussion on Slack #linear-algebra.

* Implement `propertynames(::Factor)`
* Improve error message for things like `sparse(F.U)`
* Add comprehensive list of properties, and brief discussion of materializing components, to docstrings

### DIFF
--- a/test/cholmod.jl
+++ b/test/cholmod.jl
@@ -589,9 +589,13 @@ end
 
     @testset "cholesky, no permutation $Tv" begin
         Fs = cholesky(As, perm=[1:3;])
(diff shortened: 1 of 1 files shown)


danielwe:
@ViralBShah, thanks for dropping by and marking as draft while I was fixing the tests. CI is all green with 100 % diff coverage now, so this would be ready for review from my side.

ViralBShah:
It would be nice if @rayegun or @andreasnoack can review.

PR #652, opened 2025-09-14
`ReadOnly`: slightly restrict the dispatch on the `setindex!` method

nsajko:
Stricten up `Vararg{Any}` to `Vararg{Integer}`.

The stricter dispatch still satisfies the `AbstractArray` interface, for both cases of `IndexStyle`.

The benefit is that `Integer` excludes some common types, such as `CartesianIndex`. As a result, loading SparseArrays causes only $24$ invalidations in the sysimage after this change, as opposed to $26$ before this change.


nsajko:
merge?

PR #653, opened 2025-09-15
`ReadOnly`: do not add single-argument methods to `eachindex`, etc

nsajko:
Do not add single-argument methods to:

* `eachindex`

* `firstindex`

* `lastindex`

The above have fallbacks that work as intended as long as the two two-argument methods of `eachindex` are defined.

So add the two-argument methods of `eachindex` instead of adding single-argument methods of the three functions.

On nightly Julia, together with PR #652, this PR decreases the amount of sysimage invalidations that happen when loading SparseArrays from $26$ to $17$.

Without PR #652, this PR happens to increase the amount of invalidations, to $28$.


nsajko:
merge?

PR #659, opened 2025-11-25
Update CI workflow for Julia and checkout action

ViralBShah:
Updated CI configuration to use actions/checkout@v6 and added macOS-15-intel for macos x86 tests.


dkarrasch:
This was merged after the v1.13 split. Should this be backported?

PR #661, opened 2025-12-20
Remove more bounds checks

yuyichao:
Mostly for `nzrange` but also added more `@inbounds` as I went through the files. All of the functions that has additional `@inbounds` either takes no user index input or have already performed explicit checks on the index (e.g. `isstored`).


dkarrasch:
This was merged after the v1.13 split. Should this be backported?

ViralBShah:
Yes, we should backport this.

PR #666, opened 2026-01-01
Assorted optimizations for sparse dense matrix multiplication

yuyichao:
The change is mainly based on testing of the `_spmul!(C::StridedMatrix, X::DenseMatrixUnion, A::SparseMatrixCSCUnion2, α::Number, β::Number)` function which is also the function used for my performance numbers below. I've then also applied the improvements to other similar functions though the performance for some of them may not be as big (since not all functions touched are amendable to vectorization).

The main improvement is to hoist matrix size and pointer access out of the loop to work around https://github.com/JuliaLang/julia/issues/60409 . This change have as much as 2x performance impact for complex numbers (it actually be even more on armv8.3-a (i.e. including all apple processors) and above by better triggering LLVM's complex number multiplication pattern matching with https://github.com/llvm/llvm-project/pull/173818).

Adding `muladd` is the second most important change which affect mostly complex number and bigfloat since the cost of operations saved is more significant compared to the bare memory access.

And then there are other minor tweaks that are mostly useful for small matrices (~20% impact for ~10x10 matrix). These were included mainly because I was working on another optimization which may not work well for small cases. I then make these optimizations to the small matrix cases so that I can make a fair comparison for the effect caused by the other change. I've still not done with testing the other change yet but these small fixes are ready so I've included them here.

### DIFF
--- a/src/linalg.jl
+++ b/src/linalg.jl
@@ -83,7 +83,7 @@ function _spmatmul!(C, A, B, α, β)
         throw(DimensionMismatch("second dimension of B, $(size(B,2)), does not match the second dimension of C, $(size(C,2))"))
     nzv = nonzeros(A)
     rv = rowvals(A)
-    β != one(β) && LinearAlgebra._rmul_or_fill!(C, β)
+    isone(β) || LinearAlgebra._rmul_or_fill!(C, β)
     for k in axes(C, 2)
         @inbounds for col in axes(A,2)
             αxj = B[col,k] * α
@@ -104,7 +104,7 @@ function _At_or_Ac_mul_B!(tfun::Function, C, A, B, α, β)
         throw(DimensionMismatch("second dimension of B, $(size(B,2)), does not match the second dimension of C, $(size(C,2))"))
     nzv = nonzeros(A)
     rv = rowvals(A)
-    β != one(β) && LinearAlgebra._rmul_or_fill!(C, β)
+    isone(β) || LinearAlgebra._rmul_or_fill!(C, β)
     for k in axes(C, 2)
         @inbounds for col in axes(A,2)
             tmp = zero(eltype(C))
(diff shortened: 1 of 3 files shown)


src/linalg.jl
```diff
@@ -4,6 +4,41 @@ using LinearAlgebra: AbstractTriangular, StridedMaybeAdjOrTransMat, UpperOrLower
...
+    # An immutable fixed size wrapper for matrices to work around
+    # the performance issue caused by https://github.com/JuliaLang/julia/issues/60409
+    struct _FixedSizeMatrix{Trans,R}
+        ref::R
+        nrow::Int
+        ncol::Int
+        function _FixedSizeMatrix{Trans}(ref::R, nrow, ncol) where {Trans,R}
+            new{Trans,R}(ref, nrow, ncol)
+        end
+    end
```
ViralBShah: BTW, can we use the existing `FixedSparseCSC` type that we already have?

yuyichao: The `FixedSparseCSC` is for sparse matrix only while this is for dense matrix so I don't think so. It could probably use `FixedSizeArray` though if not for dependency issue (I don't want to pull in a new package as a dependency for a standard library package).

src/sparsevector.jl
```diff
@@ -2085,9 +2083,9 @@ function _At_or_Ac_mul_B!(tfun::Function,
         "Matrix A has $n columns, but vector x has a length $(length(x))"))
     length(y) == n || throw(DimensionMismatch(
         "Matrix A has $m rows, but vector y has a length $(length(y))"))
-    n == 0 && return y
```
dkarrasch: Why this type of change? In `LinearAlgebra`, the style seems to be to return the result array. That's also what the docstring suggests.

dkarrasch: Aha, I see it is tested that `mul!(C, ...) === C`. This seems to be handled by upper-level functions then.

Issue #671, opened 2026-01-27
init_once makes it possible to allocate with one malloc and free with another


gbaraldi:
This was found in https://github.com/JuliaLang/julia/pull/60568 when running the MinFEM tests with that PR, specifically the poisson_test.jl file. Basically what seems to happen is that `umfpack_dl_numeric` calls SuiteSparse_malloc which before calling into CHOLMOD points to the libc allocator, but after we initialize it in `init_once` it points to the jl_malloc allocator. Currently this doesn't cause segfaults but will cause the GC to miscount, but with mimalloc this segfaults. That initialization needs to either be a __init__ or be called whenever any function may call any of the SuiteSparse_malloc and related

gbaraldi:
Right, the only issue is that the init seems to happen a bit too late.

PR #677, opened 2026-02-16
Unify initialization of SuiteSparse libraries

rayegun:
Create a unified initialization function for all SuiteSparse libraries (both in the stdlib and elsewhere like KLU). This remains lazily initialized. Should address https://github.com/JuliaSparse/SparseArrays.jl/issues/671. 

I wonder whether this should be truly `public` but KLU, AMD, and ParU, SPEX will rely on this to function correctly so probably should be. 

Also bumps wrapper generation. 

@gbaraldi not sure how to test this with your PR but tests on MinFEM.jl seem to pass locally for me.


rayegun:
This is ready, although I have no idea how backporting works

ViralBShah:
Feels pretty straightforward to me. For backporting, just tag with the right label. @dkarrasch has been helping prepare the backport PR.

PR #678, opened 2026-02-21
Move `nonzeroinds` to `LinearAlgebra`

jishnub:
This builds on https://github.com/JuliaLang/LinearAlgebra.jl/pull/1534.

One issue with this would be that direct uses of `SparseArrays.nonzeroinds` would lead to warnings:
```julia
julia> s = sparsevec([2, 6], [2,3])
6-element SparseVector{Int64, Int64} with 2 stored entries:
  [2]  =  2
  [6]  =  3

julia> SparseArrays.nonzeroinds(s)
[ Warning: nonzeroinds is defined in LinearAlgebra and is not public in SparseArrays
2-element Vector{Int64}:
 2
 6
```
This may be remedied by making the function public, but then this had not been declared as `public` anyway.

The other alternative might be to retain the function in `SparseArrays` and just add a method to the one from `LinearAlgebra`, as in the PR linked above.


araujoms (approved):
LGTM

PR #682, opened 2026-03-29
fix infinite loop in cond

### LINKED ISSUE #680: There is a bug in the calculation conditions for sparse matrices
using LinearAlgebra,SparseArrays
A = sparse([1.0 2.0;3.0 4.0])
SparseArrays.cond(A,1)

Has anyone noticed this issue? The code above seems to consistently fail to exit. `cond(collect(A),1)` works fine.

araujoms:
Closes #680

For `n==2` this code was trying to find three non-parallel +-1 vectors, which is mathematically impossible.

### DIFF
--- a/src/linalg.jl
+++ b/src/linalg.jl
@@ -1742,7 +1742,7 @@ function opnormestinv(A::AbstractSparseMatrixCSC{T}, t::Integer = min(2,maximum(
                             repeated = true
                         end
                     end
-                    if !repeated && 2^(n-1) ≥ 2t #we need enough ±1 non-parallel vectors to exist
+                    if !repeated && 2^(n-1) ≥ 2t #we need enough non-parallel ±1 vectors
                         saux2 = S[1:n,j]' * S_old[1:n,1:t]
                         if _any_abs_eq(saux2,n)
                             repeated = true
--- a/src/linalg.jl
+++ b/src/linalg.jl
@@ -1742,7 +1742,7 @@ function opnormestinv(A::AbstractSparseMatrixCSC{T}, t::Integer = min(2,maximum(
                             repeated = true
                         end
                     end
-                    if !repeated
+                    if !repeated && n > 2
                         saux2 = S[1:n,j]' * S_old[1:n,1:t]
                         if _any_abs_eq(saux2,n)
                             repeated = true
--- a/src/linalg.jl
+++ b/src/linalg.jl
@@ -1742,7 +1742,7 @@ function opnormestinv(A::AbstractSparseMatrixCSC{T}, t::Integer = min(2,maximum(
                             repeated = true
                         end
                     end
-                    if !repeated && n > 2
+                    if !repeated && 2^(n-1) ≥ 2t #we need enough ±1 non-parallel vectors to exist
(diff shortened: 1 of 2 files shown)


JamesWrigley (dismissed):
Weak LGTM!

(I have a tenuous understanding of the problem, but the logic and new tests make sense to me)

araujoms:
I can't say I fully understand `opnormestinv` either. I'm certain my fix doesn't affect the previously working cases, and improves the situation for `n == 2`, so I'm comfortable with merging.

Nevertheless, it would be better if somebody who does understand it could comment. I did some archaeology, and who originally wrong this function was @mfasi, maybe he'll spare us a glance?

mfasi:
Thanks for fixing this issue!

The function estimates the 1-norm of $A^{-1}$ without forming $A^{-1}$ explicitly.

Your diagnosis is correct, and the fix is not to enforce orthogonality with `S_old` when `t == n`, as you are doing here.

I would perhaps change the condition to `n > t` rather than `n > 2`, as a user could in principle request a `t` larger than the default `2`, if this function is exported in the future.

araujoms:
Thanks for chiming in! I don't think that's quite right, though. We need `2t` non-parallel ±1 vectors to exist for the loop to be finite. In dimension `n` we have `2^(n-1)` such vectors, so for example `n == t == 4` works fine.

But `n == t == 3` indeed fails, I hadn't noticed, fixed now.

araujoms:
What I'd like to understand is why does it matter whether the columns of `S` are parallel to the columns of `Sold`. Can't we get rid of this check in all cases?

mfasi:
> Thanks for chiming in! I don't think that's quite right, though. We need `2t` non-parallel ±1 vectors to exist for the loop to be finite. In dimension `n` we have `2^(n-1)` such vectors, so for example `n == t == 4` works fine.
> 
> But `n == t == 3` indeed fails, I hadn't noticed, fixed now.

Right! I agree that the current condition is the right one.

> What I'd like to understand is why does it matter whether the columns of `S` are parallel to the columns of `Sold`. Can't we get rid of this check in all cases?

The algorithm is mathematically correct without that check, which I think is an engineering solution to an observed problem. As I understand it, the only goal of enforcing that condition is to ensure that the algorithm doesn't get stuck in a cycle where it loops back and forth without any optimisation progress.

Removing it may affect convergence, but I would say that it's mostly useful for small `n`, where the set of possible sign directions is small.

araujoms:
Thanks for the explanation, I'll leave it as it is then.

araujoms:
@dkarrasch could you backport this one as well?

dkarrasch:
Sure. Please look carefully when merging: for normal PRs, squash-merge is preferred over a merge commit. We needed to allow for merge commits in the settings for the backport PRs.

araujoms:
Thanks! I was sure I did a squash-merge, though.

dkarrasch:
https://github.com/JuliaSparse/SparseArrays.jl/commits/main/ shows all three commits. It's somewhat annoying that the different merge options are not distinguished by color, so one really needs to read carefully. The browser memorizes the last choice, though.

araujoms:
Oh, ok, sorry. I'll pay more attention next time.

PR #689, opened 2026-04-04
Fix `mul!` for `HermOrSym` sparse matrices

### LINKED ISSUE #688: `mul!` incorrect for `Symmetric` and `Hermitian` matrices.
```julia-repl

julia> A = [1 3; 2 4]
2×2 Matrix{Int64}:
 1  3
 2  4

julia> B = [1 1; 1 1]
2×2 Matrix{Int64}:
 1  1
 1  1

julia> mul!(copy(B), B, Hermitian(A), true, true)
2×2 Matrix{Int64}:
 5  8
 5  8

julia> mul!(copy(B), B, Hermitian(sparse(A)), true, true)
2×2 Matrix{Int64}:
 5  7
 5  7
```

dkarrasch:
Fixes #688 on `main`. I need to check if this can be backported to v1.12. We need a slightly different fix for v1.10.


dkarrasch:
Yes, should be easily backportable to v1.12.

PR #693, opened 2026-04-05
Make destination array type in sparse multiplication consistent

dkarrasch:
This intends to fix the unexpected matrix return type behaviour observed in https://github.com/JuliaSparse/SparseArrays.jl/issues/685#issue-4179979925. The goal is determine the return type by the (potentially) non-sparse factor. For `BandedMatrix` types, however, the return type should be sparse, since these are themselves "sparse".

I can't seem to be able to run tests, but it seems that this yields a major speed-up in multiplication, and so fixes #685 (to be confirmed).

### DIFF
--- a/test/linalg.jl
+++ b/test/linalg.jl
@@ -110,12 +110,13 @@ end
 end
 
 @testset "destination array density in multiplication" begin
-    for tA in (adjoint, transpose, Hermitian, Symmetric, UpperTriangular, UpperHessenberg)
+    wrappers = (adjoint, transpose, Hermitian, Symmetric, UpperTriangular, LowerTriangular, UnitUpperTriangular, UnitLowerTriangular, UpperHessenberg)
+    for tA in wrappers
         A = randn(5,5)
         At = tA(A)
         S = sprandn(5,5,0.3)
         St = tA(S)
-        for tB in (adjoint, transpose, Hermitian, Symmetric, UpperTriangular, UpperHessenberg)
+        for tB in wrappers
             B = sprandn(5,5, 0.3)
             Bt = tB(B)
             C = At*Bt
--- a/test/linalg.jl
+++ b/test/linalg.jl
@@ -120,14 +120,14 @@ end
             Bt = tB(B)
             C = At*Bt
             @test C ≈ Matrix(At) * Matrix(Bt)
-            @test C isa DenseVecOrMat
+            @test !issparse(C)
             D = St*Bt
             @test D ≈ Matrix(St) * Matrix(Bt)
-            @test D isa SparseMatrixCSC
+            @test issparse(D)
         end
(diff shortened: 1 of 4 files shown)


dkarrasch:
This obviously changes return types and is therefore not for backport (except for v1.13?).

dkarrasch:
With this PR, I get (after pre-compilation)

```julia

julia> A = Hermitian(sprand(5000, 5000, .01));

julia> X = rand(5000, 10);

julia> Y = rand(10, 5000);

julia> @time A * X;
  0.002859 seconds (6 allocations: 390.789 KiB)

julia> @time Y * A;
  0.001683 seconds (3 allocations: 390.711 KiB)

julia> @time A * Y';
  2.896842 seconds (4 allocations: 390.727 KiB)

julia> @time X' * A;
  0.001743 seconds (4 allocations: 390.727 KiB)
```

The case `A*Y` is fixed by #692. With that fix applied, I get (after pre-compilation)

```julia

julia> A = Hermitian(sprand(5000, 5000, .01));

julia> X = rand(5000, 10);

julia> Y = rand(10, 5000);

julia> @time A * X;
  0.002663 seconds (6 allocations: 390.789 KiB)

julia> @time Y * A;
  0.001571 seconds (3 allocations: 390.711 KiB)

julia> @time A * Y';
  0.002957 seconds (8 allocations: 390.820 KiB)

julia> @time X' * A;
  0.001666 seconds (4 allocations: 390.727 KiB)
```

EDIT: The `X'A` timings are nonsense, I can't reproduce them. Not sure what I was measuring.

dkarrasch:
I wonder if we really need the `StructuredMatrix` methods over at LinearAlgebra.jl. They require massive disambiguation here, but since the fallback is to call 3-arg `similar`, I don't think they add anything that the fallback wouldn't give already. @jishnub

dkarrasch:
After https://github.com/JuliaLang/LinearAlgebra.jl/pull/1572, the current state of this PR should be enough to create dense target arrays as long as one factor is dense and the other is "quasi-sparse", meaning sparse or a wrapper of a sparse array. Sparse times banded would yield sparse arrays.

dkarrasch:
Hallelujah!

test/linalg.jl
```diff
@@ -109,6 +109,43 @@ end
     end
 end
 
+@testset "destination array density in multiplication" begin
+    for tA in (adjoint, transpose, Hermitian, Symmetric, UpperTriangular, UpperHessenberg)
```
araujoms: No `LowerTriangular`?

araujoms (approved):
I hope you don't mind if I just merge it; I just added a simple test, and the lack of this PR is ruining CI for everyone.

dkarrasch:
Yeah, I was on vacation, offline. Thanks for taking over.

PR #697, opened 2026-04-12
Use generic_trimatdiv!() in ldiv!() for QRSparse

### LINKED ISSUE #696: QR performance regression
See https://github.com/JuliaLang/julia/pull/61452 for history. This reduces to the following:

```julia
# Minimal reproducer: SubArray of SparseMatrixCSC loses specialized triangular solve
#
# `ldiv!(UpperTriangular(R), x)` is fast when R is a SparseMatrixCSC,
# but falls back to a generic O(n^2) method when R is a SubArray view.
#
# Introduced in JuliaSparse/SparseArrays.jl#676 (commit 95b6ac4).
# The new ldiv! uses a @view of F.R instead of indexing:
#   New (slow): https://github.com/JuliaSparse/SparseArrays.jl/blob/95b6ac4b0fa4d99e17920505dd36d7b95d91a0ab/src/solvers/spqr.jl#L513
#   Old (fast): https://github.com/JuliaSparse/SparseArrays.jl/blob/4500d8656d26a9328f50a48eaf9eca2f1fabc8ef/src/solvers/spqr.jl#L440

n = 9000
R = spdiagm(0 => fill(2.0, n), 1 => fill(1.0, n - 1))
x = ones(n)

R_copy = R[Base.OneTo(n), Base.OneTo(n)]   # SparseMatrixCSC
R_view = @view R[Base.OneTo(n), Base.OneTo(n)]  # SubArray

# Warmup
ldiv!(UpperTriangular(R_copy), copy(x))
ldiv!(UpperTriangular(R_view), copy(x))

t_copy = @elapsed for _ in 1:10; ldiv!(UpperTriangular(R_copy), copy(x)); end
t_view = @elapsed for _ in 1:10; ldiv!(UpperTriangular(R_view), copy(x)); end

println("SparseMatrixCSC:  $(round(t_copy/10*1e6, digits=1)) μs")
println("SubArray view:    $(round(t_view/10*1e6, digits=1)) μs")
println("Slowdown:         $(round(t_view/t_copy, digits=0))x")
```

JamesWrigley:
This avoids having to go through UpperTriangular, which requires making a square matrix out of our rectangular F.R matrix. Taking a view of F.R would hit a slow fallback for ldiv!(), and making a new square matrix would cause allocations.

Fixes #696. Written with help from Claude :robot: 

Benchmark script adapted from BaseBenchmarks:
```julia
m = 10000
n = 9000
getA() = spdiagm(0 => fill(2.0, m),
                 -1 => fill(1.0, m - 1),
                 1 => fill(1.0, m - 1),
                 360 => fill(1.0, m - 360))[:, 1:n]
getB()   = ones(m, 2)

@benchmark qr(A)\B setup=(A=$getA(); B=$getB())

# main
BenchmarkTools.Trial: 28 samples with 1 evaluation per sample.
 Range (min … max):  175.511 ms … 218.146 ms  ┊ GC (min … max): 0.41% … 21.76%
 Time  (median):     178.119 ms               ┊ GC (median):    0.85%
 Time  (mean ± σ):   181.623 ms ±  10.391 ms  ┊ GC (mean ± σ):  2.55% ±  5.57%

  ██▅  ▂     ▂                                                   
  ███▁▁██▁▅▅▅█▁▁▁▅▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▅▁▁▅ ▁
  176 ms           Histogram: frequency by time          218 ms <

 Memory estimate: 45.61 MiB, allocs estimate: 197.

# PR
BenchmarkTools.Trial: 174 samples with 1 evaluation per sample.
 Range (min … max):  18.428 ms … 70.734 ms  ┊ GC (min … max):  0.00% … 66.01%
 Time  (median):     23.961 ms              ┊ GC (median):     2.37%
 Time  (mean ± σ):   28.509 ms ± 13.099 ms  ┊ GC (mean ± σ):  20.07% ± 20.51%

   █    ▂                                                      
  ▆█▅▆▅██▅▅█▇▅▄▃▃▁▁▁▁▃▂▁▂▃▁▂▁▁▁▁▁▁▁▁▁▁▁▁▁▃▁▃▁▂▃▁▂▁▁▁▁▁▁▁▃▃▂▂▃ ▂
  18.4 ms         Histogram: frequency by time        68.4 ms <

 Memory estimate: 45.61 MiB, allocs estimate: 197.
```

### DIFF
--- a/src/solvers/spqr.jl
+++ b/src/solvers/spqr.jl
@@ -518,8 +518,23 @@ function LinearAlgebra.ldiv!(X::StridedVecOrMat{T}, F::QRSparse{T}, B::StridedVe
         lmul!(adjoint(F.Q), W0)
 
         # Solve R*X = Q'*P*B
-        ldiv!(UpperTriangular(@view(F.R[Base.OneTo(rnk), Base.OneTo(rnk)])),
-              @view(W0[Base.OneTo(rnk), :]))
+        #
+        # We call generic_trimatdiv! directly instead of going through
+        # ldiv!(UpperTriangular(R_sub), ...) for two reasons:
+        # 1. UpperTriangular requires a square matrix, but F.R is m×n
+        #    so we can only take a column view R[:, 1:rnk] (which is
+        #    m×rnk, not square). A row+column view R[1:rnk, 1:rnk]
+        #    would be square but doesn't match SparseMatrixCSCView,
+        #    causing dispatch to a slow generic fallback.
+        # 2. generic_trimatdiv! is what UpperTriangular ldiv! dispatches
+        #    to anyway — calling it directly with uploc='U', isunitc='N',
+        #    tfun=identity is equivalent. The back-substitution loop
+        #    iterates over axes(B,1) = 1:rnk and searchsortedlast
+        #    excludes entries with row > j, so the extra rows in the
+        #    column view are never accessed.
+        W_rnk = @view(W0[Base.OneTo(rnk), :])
+        LinearAlgebra.generic_trimatdiv!(W_rnk, 'U', 'N', identity,
+                                         @view(F.R[:, Base.OneTo(rnk)]), W_rnk)
 
         # Apply right permutation: scatter solved rows into X using cpiv directly.
         # Zero X first so free variables (beyond rank) are zero in the basic solution.


dkarrasch:
Would a widened method signature somewhere help (also other cases)?

JamesWrigley:
Hmm I'm not sure, do you mean adding something like `SparseMatrixCSCView` but for a `UnitRange` of rows and columns?

dkarrasch:
It was just a wild guess, like could we modify the call path by changing a method signature to allow the previously used method to be used (again)? In case you know, which `ldiv!` is directing away from the specialized `_generic_trimatdiv!`, to the slow fallback? Because that specialized generic one happily accepts SparseMatrixViews.

JamesWrigley:
Ah I see what you mean :thinking: So according to Cthulhu these are the signatures for the copy and view respectively:
```julia
generic_trimatdiv!(C::Vector{Float64},
                   uplo_char(A::UpperTriangular{Float64, SparseMatrixCSC{Float64, Int64}})::Core.Const('U'),
                   isunit_char(A::UpperTriangular{Float64, SparseMatrixCSC{Float64, Int64}})::Core.Const('N'),
                   _wrapperop(parent(A))::Core.Const(identity),
                   _unwrap_at(parent(A))::SparseMatrixCSC{Float64, Int64},
                   B::Vector{Float64})

generic_trimatdiv!(C::Vector{Float64},
                   uplo_char(A::UpperTriangular{Float64, SubArray{Float64, 2, SparseMatrixCSC{Float64, Int64}, Tuple{Base.OneTo{Int64}, Base.OneTo{Int64}}, false}})::Core.Const('U'),
                   isunit_char(A::UpperTriangular{Float64, SubArray{Float64, 2, SparseMatrixCSC{Float64, Int64}, Tuple{Base.OneTo{Int64}, Base.OneTo{Int64}}, false}})::Core.Const('N'),
                   _wrapperop(parent(A))::Core.Const(identity),
                   _unwrap_at(parent(A))::SubArray{Float64, 2, SparseMatrixCSC{Float64, Int64}, Tuple{Base.OneTo{Int64}, Base.OneTo{Int64}}, false},
                   B::Vector{Float64})

```

I think the issue is that this subarray type has a `Base.OneTo{Int}` as the first axes type argument: `SubArray{Float64, 2, SparseMatrixCSC{Float64, Int64}, Tuple{Base.OneTo{Int64}, Base.OneTo{Int64}}`

Whereas `SparseMatrixCSCView` requires it to be a `Base.Slice{Base.OneTo{Int}}`, which would be the equivalent of `@view F.R[:, ...]`. That makes it not match our sparse matrix method and fallback to the LinearAlgebra one. Does it make sense to add limited support for views with a `Base.OneTo` selection of rows? I guess at least in the case of `generic_trimatdiv!()` it's safe.

dkarrasch:
I think that could be another PR. Being able to assume that the whole column is included in the view may be important, given how we store row numbers by column. So, for the time being this should be okay.

dkarrasch (approved):
LGTM, and I assume this is well covered by tests.

IanButterworth:
Could this have regressed sparse.transpose.adjoint?

image

https://github.com/JuliaCI/NanosoldierReports/blob/master/benchmark/by_date/2026-04/28/report.md

JamesWrigley:
My linear algebra is rusty but I don't think ldivs are used in adjoints? :thinking:

IanButterworth:
Sorry, ignore me, I hadn't realized the sparse arrays fix wasn't in the benchmarked commit range https://github.com/JuliaLang/julia/compare/4fdd12e2776419e205b96c3d0a407ba682c9913d...d19eac062f4083e8cb4777323904cb87fab35200

JamesWrigley:
Well, just in case :stuck_out_tongue: https://github.com/JuliaLang/julia/pull/61692

PR #700, opened 2026-04-16
Add `indtype` for `SparseMatrixCSCColumnSubset`

### LINKED ISSUE #699: Missing method definitions for column views
When using column views of sparse matrices, `sparse_view*dense` matrix multiplication works well, but `sparse_view*sparse` does not.
I suspect this is an oversight and that it would be easy to make it work. Is that right?
Here's an example.

```julia-repl
julia> S = sparse(1:4, 1:4, 1:4)
4×4 SparseMatrixCSC{Int64, Int64} with 4 stored entries:
 1  ⋅  ⋅  ⋅
 ⋅  2  ⋅  ⋅
 ⋅  ⋅  3  ⋅
 ⋅  ⋅  ⋅  4

julia> Sv = @view S[:,1:2]
4×2 view(::SparseMatrixCSC{Int64, Int64}, :, 1:2) with eltype Int64:
 1  ⋅
 ⋅  2
 ⋅  ⋅
 ⋅  ⋅

julia> Sv*zeros(2)
4-element Vector{Float64}:
 0.0
 0.0
 0.0
 0.0

julia> Sv*spzeros(2)
ERROR: MethodError: no method matching indtype(::SubArray{Int64, 2, SparseMatrixCSC{Int64, Int64}, Tuple{Base.Slice{Base.OneTo{…}}, UnitRange{Int64}}, false})
The function `indtype` exists, but no method is defined for this combination of argument types.
```

dkarrasch:
Fixes #699. Should be backported.

### DIFF
--- a/src/sparsematrix.jl
+++ b/src/sparsematrix.jl
@@ -313,6 +313,8 @@ Base.@propagate_inbounds nzrange(S::SparseMatrixCSCColumnSubset, col::Integer) =
 nzrange(S::UpperTriangular{<:Any,<:SparseMatrixCSCUnion}, i::Integer) = nzrangeup(S.data, i)
 nzrange(S::LowerTriangular{<:Any,<:SparseMatrixCSCUnion}, i::Integer) = nzrangelo(S.data, i)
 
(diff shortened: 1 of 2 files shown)


araujoms (approved):
LGTM

PR #712, opened 2026-05-06
Fix and test `reuse_symbolic=false` for UMFPACK `lu!()`

johnomotani:
The `reuse_symbolic=false` option for UMFPACK's `lu!()` is broken, due to undefined `Tv` and `Ti`. This PR fixes that bug, and adds a test that uses `reuse_symbolic=false`.


araujoms:
Instead of copy-pasting the entire test you could add another loop `for reuse in (true, false)`.

johnomotani:
> Instead of copy-pasting the entire test you could add another loop `for reuse in (true, false)`.

@araujoms - done.

araujoms (approved):
Thanks for your contribution!

PR #716, opened 2026-05-15
test: avoid method-overwrite warnings in forbidproperties.jl

PatrickHaecker:
`test/forbidproperties.jl` is `include`d from six sibling test modules (`SparseVectorTests`, `SparseLinalgTests`, `SparseTests`, `SparseIssuesTests`, `HigherOrderFnsTests`, `SparseMatrixConstructorIndexingTests`). It defines two methods on `Base.getproperty` for `SparseMatrixCSC` and `SparseVector`. With the [(hopefully) upcoming Julia default `--warn-overwrite=default`](https://github.com/JuliaLang/julia/pull/61430), the second through sixth includes each produce method-overwrite warnings.

Make the overrides idempotent by only defining them when undefined.

I used Claude Opus 4.7 while preparing this.

### DIFF
--- a/test/forbidproperties.jl
+++ b/test/forbidproperties.jl
@@ -1,5 +1,18 @@
 # This file is a part of Julia. License is MIT: https://julialang.org/license
 
 using SparseArrays
-Base.getproperty(::SparseMatrixCSC, ::Symbol) = error("use accessor function")
-Base.getproperty(::SparseVector, ::Symbol) = error("use accessor function")
+# Only define each `getproperty` override if it isn't already defined. This
+# file is `include`d from multiple sibling test modules, and without this
+# guard each subsequent include would re-set the same `Base` methods,
+# producing spurious method-overwrite warnings (depending on `--warn-overwrite`).
+# `hasmethod` is not sufficient here because the generic `Base.getproperty`
+# fallback for `Any` always exists; we instead check whether `which` returns
+# that fallback method.
+let fallback = which(Base.getproperty, Tuple{Any, Symbol})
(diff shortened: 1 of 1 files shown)


dkarrasch:
IIUC then having this is "harmless" regardless of the Julia PR, right? So merging this is safe?

PatrickHaecker:
> IIUC then having this is "harmless" regardless of the Julia PR, right? So merging this is safe?

Yes, exactly. It's already now helping with `--warn-overwrite=yes`, because the overwrites here are probably not the overwrites the user is interested in.

The reason why this is needed is interesting and I used your question as an opportunity to write it down:

It's basically the `#ifndef` from `C`. And we need it because of the same reason as `C` needs it: The `include` mechanism on its own is a mechanism to split code, not to share code. However, here we use it for sharing code, which leads to exactly the unwanted behavior we sometimes want to warn about: (accidentally) redefining something.

Luckily, opposed to `C`, `Julia` has a more advanced mechanism with its `module`s. However, `include`ing a `module` is still an `include`, as the module would end up as separate submodules in the (sub)module hierarchy.

So we would need to make it a package to use Julia's even more powerful mechanism. This would work, but at this point we are talking about a larger refactor than I was comfortable doing as a small PR. It would be cleaner, however.

dkarrasch:
I think this is really just for internal usage in the tests, so this has nothing to do with the actual productive code. This is used to guard SparseArrays.jl of introducing (specific) `getproperty` calls, and use accessor functions that then need to be specialized for concrete sparse array subtypes.

PatrickHaecker:
Yes, exactly. That's why adding this "guard logic" is the pragmatic solution, even if it is not the "cleanest".

Issue #726, opened 2026-06-19
CHOLMOD allocations


lorenzogambichler:
### Problem

Solving against an existing sparse Cholesky factorization in a loop allocates on every solve, even with a preallocated solution and RHS and a reused factorization. For workloads with many solves against a fixed factorization, e.g. time-stepping, PDE-constrained optimization, eigensolvers, these per-solve allocations add up.

Example:
```julia
using SparseArrays, LinearAlgebra
N = 5000
A = sprand(N, N, 1e-3)
A = A*A' + N*I # sparse SPD
F = cholesky(A)
b = rand(N)
x = similar(b)
ldiv!(x, F, b) # warm-up
@allocated ldiv!(x, F, b) # Allocates
```

The allocations come from CHOLMOD allocating and freeing its internal `Y/E` workspace on each solve. `cholmod_solve2` lets the caller pass persistent buffers instead.
This can be solved by calling `cholmod_solve2` or `cholmod_l_solve2` directly, see https://github.com/lorenzogambichler/CHOLMODSolve.jl as a quick proof of concept. But since the SuiteSparse version is tied to the Julia version, this solution is a bit fragile. This can be seen from `CholmodSolve2.jl`, a previous solution attempt, which has not been updated and is therefore unusable.

### Proposed direction

Possibly a reusable workspace object that `ldiv!` or `solve!` can optionally accept. Thereby the internal `Y` and `E` buffers can be pre-allocated once before solving, similar to the repo above. 

This issue has already come up a few times in the past, e.g. in the julialang discourse.

ViralBShah:
Yes, this would be good to have and a PR would be a great starting point.

bbrehm:
I think this also leaks memory https://github.com/JuliaSparse/SparseArrays.jl/blame/896723afccc0833aee2b2131ec0d444857caa8c2/src/solvers/cholmod.jl#L1966

Regardless of whether that is responsible for the observed-on-discourse memory leakage, this code looks very wrong.

stevengj:
See also this [discourse thread](https://discourse.julialang.org/t/running-out-of-ram-with-repeated-cholesky-solves/1360440) for a "pointer" to a memory leak, which should be fixed (independent of an allocation-free API): #733.

PR #727, opened 2026-06-20
Add workspace to Factor for allocation-free CHOLMOD solves

lorenzogambichler:
Addresses #726 

Makes repeated solves against a fixed sparse Cholesky factorization allocation-free by reusing CHOLMOD's `cholmod_solve2` `Y`/`E` buffers across calls. Each `Factor` stores its own reusable workspace and a `ReentrantLock`, so the existing API stays the same and eliminates all julia-sided allocations.

```julia
F = cholesky(A)
for b in rhs
    ldiv!(x, F, b) # allocates ~16 bytes from CHOLMOD, as tested
end
```

### DIFF
--- a/src/solvers/cholmod.jl
+++ b/src/solvers/cholmod.jl
@@ -1919,28 +1919,29 @@ const AbstractSparseVecOrMatInclAdjAndTrans = Union{AbstractSparseVecOrMat, AdjO
 """
     CholmodWorkspace()
 
-Reusable workspace for allocation-free calls to `ldiv!(x, L, b, ws)` with a
-CHOLMOD [`Factor`](@ref).
+Reusable workspace for allocation-free calls to `ldiv!(x, L, b, ws)`.
 
 `cholmod_solve2` allocates two temporary dense matrices (Y and E) on every solve.
 A `CholmodWorkspace` holds those buffers across calls so that CHOLMOD reuses them
 instead of reallocating. It also pre-allocates the `cholmod_dense_struct` wrappers
-for `x` and `b`, eliminating all Julia-side heap allocations per solve.
+for `x` and `b`.
+This eliminates all Julia-side heap allocations per solve.
 
-The workspace is finalizer-protected: the CHOLMOD-managed Y and E buffers are freed
-automatically when the workspace is garbage-collected. A single workspace can be
-reused across calls with different arrays or a different number of RHS columns;
-CHOLMOD will resize Y and E as needed.
+A finalizer frees Y adn E automatically when the worspace is garbage-collected.
+
+A single workspace can be reused across calls with different arrays or a different number of RHS columns since
+CHOLMOD will resize Y and E.
 
 # Example
 ```julia
-F  = cholesky(A)
+F = cholesky(A)
 ws = CholmodWorkspace()
 for b in rhs_list
     ldiv!(x, F, b, ws)
 end
 ```
 """
+
 mutable struct CholmodWorkspace
     dense_x::cholmod_dense_struct
     dense_b::cholmod_dense_struct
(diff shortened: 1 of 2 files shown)


stevengj:
We already have a `cholesky!(::CHOLMOD.Factor, ::SparseMatrixCSC)` API, couldn't we just make that allocation-free by storing the workspace in the `Factor`?

lorenzogambichler:
I see, this is a good point. It seems like that is essentially how UmfpackLU handles the workspace caching? I'll rework the PR to add a workspace on the Factor with a default getworkspace accessor, essentially trying to mirror UmfpackLU.

lorenzogambichler:
Instead of an optional workspace argument, the workspace now lives inside the Factor. I tried to keep it similar to UmfpackWS and UmfpackLU in umfpack.jl. This keeps the 3-arg ldiv! API, while making it allocation-free.

src/solvers/cholmod.jl
```diff
@@ -258,8 +258,50 @@ function Sparse(p::Ptr{cholmod_sparse})
...
+        ws = new(
+            cholmod_dense_struct(), # all fields will be written in ldiv! before use
+            cholmod_dense_struct(),
+            Ref(Ptr{cholmod_dense_struct}(C_NULL)),
+            Ref(Ptr{cholmod_dense_struct}(C_NULL)),
+            Ref(Ptr{cholmod_dense_struct}(C_NULL))
+        )
+        finalizer(ws) do w
+            w.Y[] != C_NULL && free!(w.Y[])
+            w.E[] != C_NULL && free!(w.E[])
```
stevengj:
```suggestion
            free!(w.Y[])
            free!(w.E[])
```
(Th `cholmod_free_dense` function, like `Libc.free`, [already checks](https://github.com/PetterS/SuiteSparse/blob/27e5a8516464a6ac40bd3fa0e5b46e51b11f4765/CHOLMOD/Core/cholmod_dense.c#L370-L374) for `NULL` pointers.)

src/solvers/cholmod.jl
```diff
@@ -258,8 +258,50 @@ function Sparse(p::Ptr{cholmod_sparse})
...
+for `x` and `b`, eliminating all Julia-side heap allocations per solve.
+
+Each [`Factor`](@ref) embeds its own workspace and uses it automatically in
+`ldiv!(x, F, b)`. For multi-thread solves with a single factor, use `copy(F)`
+for every thread so each copy gets its own independent workspace and lock.
+
+The workspace is finalizer-protected. The Y and E buffers are freed automatically
+when the workspace is garbage-collected.
+"""
+mutable struct CholmodWorkspace
```
stevengj: Not sure it's worth having a separate `struct` type here, as opposed to just making them fields of `Factor`

lorenzogambichler: The main arguments for the separate struct are that it keeps the finalizer for the cholmod buffers separate from the `free!`s for the `cholmod_factor` pointer and it allows using a single `:workspace` enttry inside `getproperty` instead of five. The performance overhead should be negligible too. If you prefer the inline version, I can do that though.

stevengj: What's the advantage of keeping the finalizer separate?  

Can one ever use the workspace separately from a factor, or share the workspace between multiple factors?

lorenzogambichler: It's exactly one set of buffers per factor, right. I've added the fields directly to Factor.

dkarrasch:
@lorenzogambichler Could you please resolve the merge conflicts? I was a bit afraid of touching this. The issue is to make sure there's no memory leak.

lorenzogambichler:
I included a test that checks for any leaks on the C side using CHOLMOD's memory_inuse to make sure.

PR #733, opened 2026-07-06
fix memory leak in cholmod

stevengj:
Fix bug identified [on discourse](https://discourse.julialang.org/t/running-out-of-ram-with-repeated-cholesky-solves/136044/9).

### DIFF
--- a/src/solvers/cholmod.jl
+++ b/src/solvers/cholmod.jl
@@ -1952,22 +1952,22 @@ for TI in IndexTypes
         dense_b, dense_b_ptr = wrap_dense_and_ptr(b)
 
         X_Handle = Ptr{cholmod_dense_struct}(dense_x_ptr)
-        Y_Handle = Ptr{cholmod_dense_struct}(C_NULL)
-        E_Handle = Ptr{cholmod_dense_struct}(C_NULL)
+        Y_Handle = Ref(Ptr{cholmod_dense_struct}(C_NULL))
+        E_Handle = Ref(Ptr{cholmod_dense_struct}(C_NULL))
         status = GC.@preserve x dense_x b dense_b begin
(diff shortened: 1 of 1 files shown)


dkarrasch:
Can this be tested? Does this need backporting (v1.13 for sure, but v1.12 & LTS)?

stevengj:
Presumably the existing tests from #547 (cc @ranocha) cover the code to make sure it works? Not sure how to test for a memory leak?

bbrehm:
I verified presence of some memory leak via `Base.gc_live_bytes()`. That might work as a regression test.

I have not looked into the allocation path in SuiteSparse or valgrind. To be honest, I'm somewhat surprised that this allocation gets properly logged and accounted for, at all!

ranocha:
Thanks a lot for the fix. I am sorry for introducing this bug two years ago.

The basic method should be tested, but I am not sure whether all branches will be tested completely.

This could be a explanation for

> The current method still allocates some memory on my system since we do not provide the workspace vectors `Handle_Y` and `Handle_E`. We could store them in the `CHOLMOD.Factor` struct. I tried that, but they were still `C_NULL` after the call to `solve2` (when initialized with `C_NULL`). Thus, re-using them is not working as I understand the docs. Thus, I have not added this complexity to this PR.

noticed in https://github.com/JuliaSparse/SparseArrays.jl/pull/547. I guess we should at least open an issue to check whether this works correctly with the fix from this PR, and consider adding this additional optimization in another PR.

ViralBShah:
cc @rayegun

stevengj:
Storing the workspace vectors in `CHOLMOD.Factor` is happening in https://github.com/JuliaSparse/SparseArrays.jl/pull/727

PR #734, opened 2026-07-06
Optimize row-indexing in `AdjOrTrans` of sparse matrix

### LINKED ISSUE #628: `X'[1,:]` much slower than `X[:, 1]`, despite theoretically same access pattern
## Expected behaviour
When I allocate a `SparseMatrixCSC`, indexing columns is fast and indexing rows slow. E.g., 
```julia
X = sprand(100_000, 100_000, 0.01);
@elapsed X[:, 1]  # fast
@elapsed X[1, :]  # slow
```
Further, when I create the lazy adjoint, then the behaviour should switch
```julia
Xt = X'  # lazy adjoint
@elapsed Xt[:, 1]  # slow
@elapsed Xt[1, :]  # fast
```
and in particular, I'd expect the timings for each "fast" and "slow" to approximately match.

## Actual behaviour
Unfortunately, although `Adjoint{SparseMatrixCSC}` "simulates" a `SparseMatrixCSR`, the indexing is quite slow for `Xt[1, :]`.
In particular:
```
julia> X = sprand(100_000, 100_000, 0.01);
julia> @elapsed X[:, 1]
9.018e-6
julia> @elapsed X[1, :]
0.001110558

julia> Xt = X';
julia> @elapsed Xt[:, 1]
0.009835612
julia> @elapsed Xt[1, :]
0.001390563
```
In particular, `Xt[1,:]` is about 100 times slower than `X[:, 1]` although the same elements are indexed, and the data structure should be the same.

I have already tried to understand what's happening by `Cthulhu.@descend`ing down the indexing callstack of the lazy transpose, but the generated functions are too difficult to see through for me.

## Why do I need this
I have an application where sparse matrix indexing performance is critical. For parts of the algorithm I need to retrieve columns quickly (no problem), and for some parts I need to retrieve rows quickly. It would therefore be nice to just explicitly compute the transpose once, and then index into that, i.e. `X_quickrow = copy(transpose(X))`. Then I don't need to change any math or syntax, but row lookup should be fast. However, as you can see above, it is not.

dkarrasch:
This uses both optimized column-indexing and `map!`, as suggested by https://github.com/JuliaSparse/SparseArrays.jl/issues/628#issuecomment-2898646451 @SobhanMP.

Fixes #628.

### DIFF
--- a/src/sparsevector.jl
+++ b/src/sparsevector.jl
@@ -683,8 +683,10 @@ function getindex(x::AbstractSparseMatrixCSC, I::AbstractUnitRange, j::Integer)
     return @if_move_fixed x SparseVector(length(I), [rowvals(x)[i] - first(I) + 1 for i = r1:r2], nonzeros(x)[r1:r2])
 end
 
-getindex(M::AdjOrTrans{T,<:AbstractSparseMatrixCSC{T}}, i, ::Colon) where {T} =
+getindex(M::AdjOrTrans{<:Any,<:AbstractSparseMatrixCSC}, i::Integer, ::Colon) =
     map!(wrapperop(M), M.parent[:,i])
+getindex(M::AdjOrTrans{<:Any,<:AbstractSparseMatrixCSC}, i::AbstractVector, ::Colon) =
+    copy(wrapperop(M)(M.parent[:,i]))
 
 # In the general case, we piggy back upon SparseMatrixCSC's optimized solution
 @inline getindex(A::AbstractSparseMatrixCSC, I::AbstractVector, J::Integer) =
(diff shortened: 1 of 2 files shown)


araujoms (dismissed):
LGTM

src/sparsevector.jl
```diff
@@ -683,6 +683,11 @@ function getindex(x::AbstractSparseMatrixCSC, I::AbstractUnitRange, j::Integer)
     return @if_move_fixed x SparseVector(length(I), [rowvals(x)[i] - first(I) + 1 for i = r1:r2], nonzeros(x)[r1:r2])
 end
 
+getindex(M::AdjOrTrans{<:Any,<:AbstractSparseMatrixCSC}, i::Integer, ::Colon) =
+    map!(wrapperop(M), M.parent[:,i])
+getindex(M::AdjOrTrans{<:Any,<:AbstractSparseMatrixCSC}, i::AbstractVector, ::Colon) =
+    copy(wrapperop(M)(M.parent[:,i]))
```
dkarrasch, line 689: I wonder if there is any existing situation where we return a wrapped array, so if we could perhaps leave out the `copy`?

dkarrasch: No, we can't.

PR #735, opened 2026-07-15
use <:Union{Number,Missing} as the value type

rokke-git:
restricts the value's type to be Union{Number,Missing}. currently, we're erroring on 'creation' when it gets output to the repl, but I'm going to fix the output function, so we will no longer be able to rely on that error to constrain type

basically just added a `<:Union{Number,Missing}` wherever a `Integer` was before, hopefully that catches everything. ~shouldn't effect anything~ except for giving better errors on instantiation

edit: theoretically shouldn't have affected anything... how is `((_, _, _, _, x)->x).(Int, Int, Int, Int, spzeros(3))` turning into `spzeros(::Core.TypeEgal{Any}, ::Core.TypeEgal{Int64}, ::Tuple{Int64})`?  
it's saying it's this line  https://github.com/JuliaSparse/SparseArrays.jl/blob/751387e43562a26999c69fa776cdaec86938e155/src/sparsevector.jl#L246  
but when I read it it looks like it should go to https://github.com/JuliaSparse/SparseArrays.jl/blob/751387e43562a26999c69fa776cdaec86938e155/src/sparsevector.jl#L243
after starting with https://github.com/JuliaSparse/SparseArrays.jl/blob/751387e43562a26999c69fa776cdaec86938e155/src/sparsevector.jl#L241

### DIFF
--- a/src/sparsematrix.jl
+++ b/src/sparsematrix.jl
@@ -9,14 +9,14 @@
 # Assumes that 0   <= length(nzval) < typemax(Ti)
 
 """
-    SparseMatrixCSC{Tv,Ti<:Integer} <: AbstractSparseMatrixCSC{Tv,Ti}
+    SparseMatrixCSC{Tv<:Union{Number,Missing},Ti<:Integer} <: AbstractSparseMatrixCSC{Tv,Ti}
 
 Matrix type for storing sparse matrices in the
 [Compressed Sparse Column](@ref man-csc) format. The standard way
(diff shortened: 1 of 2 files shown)


ViralBShah:
Just noting - This is a pretty fundamental change and we need to test it well beyond the existing set of SparseArrays tests - not only for breakage, but also compile time regressions if it introduces any.

rokke-git:
this isn't the change I really care about anyways, it was just about being correct. it appears to be having mysterious effects so I prob won't spend much time digging into this.

PR #737, opened 2026-07-23
Run only the Aqua tests in the aqua CI job

KristofferC:
Since https://github.com/JuliaSparse/SparseArrays.jl/pull/537 the aqua job runs the aqua-specific test *and then all the other tests as well!*. That is clearly a waste.


KristofferC:
aqua now takes ~1 min instead of full test suite time.

PR #746, opened 2026-08-16
Rename `generic_mat***mul!` to `mul!`

dkarrasch:
Follow-up to https://github.com/JuliaLang/LinearAlgebra.jl/pull/1671.

This will work only after https://github.com/JuliaLang/LinearAlgebra.jl/pull/1676 is merged and `LinearAlgebra.jl` is bumped.


araujoms (dismissed):

dkarrasch:
I can't get julia nightly to run on my machine. @araujoms Would you be able to check the call paths for the two methods that are seemingly uncovered?

araujoms:
Codecov is down for me, if you tell me which are the supposedly missed lines I can try them.

dkarrasch:
It's line 89 in linalg.jl, method

```julia
 mul!(C::StridedMatrix, tA, tB, A::SparseMatrixCSCUnion2, B::AbstractTriangular, alpha::Number, beta::Number)
```

and line 1901 in sparsevector.jl, method

```julia
mul!(y::AbstractVector, tA, A::UpperOrLowerTriangular, x::AbstractSparseVector, alpha::Number, beta::Number)
```

for which I thought I added a test. So it's basically SparseMatrix*Triangular and Triangular*SparseVector. Those should run by the above mentioned methods.

araujoms:
Codecov is right, multiplying a sparse matrix and an upper triangular goes instead here: https://github.com/JuliaLang/LinearAlgebra.jl/blob/9c4fde31a153bf54956b8fb044761f5086d3e47c/src/triangular.jl#L1235-L1243

Similarly, multiplying an upper triangular and a sparse vector goes here: https://github.com/JuliaLang/LinearAlgebra.jl/blob/9c4fde31a153bf54956b8fb044761f5086d3e47c/src/triangular.jl#L1222-L1229

dkarrasch:
Thanks for checking. Those misdirections have been there before this PR, so time to fix it.

dkarrasch:
The fix is ugly, we might just as well overload 5-arg `mul!` itself in those two cases, which somehow defeats the purpose of the whole thing. The problem is that for triangular multiplication, we catch and direct earlier in the call path, even before the "7-arg" `mul!` gets called.

araujoms:
It doesn't seem possible to do a proper fix without changing LinearAlgebra, and I think it shouldn't be changed anymore for 1.13, so I'd go with it as it is for now.

src/SparseArrays.jl
```diff
@@ -19,7 +19,7 @@ import Base: Matrix, Vector
 import LinearAlgebra: mul!, ldiv!, rdiv!, cholesky, adjoint!, diag, eigen, dot,
     issymmetric, istril, istriu, lu, tr, transpose!, tril!, triu!, isbanded, isdiag,
     cond, diagm, factorize, ishermitian, norm, opnorm, lmul!, rmul!, tril, triu,
-    matop_dest, generic_matvecmul!, generic_matmatmul!, generic_matmatmul_wrapper!, copytrito!, nonzeroinds
```
Zentrik, line 22: Claude tells me that the removal of the `generic_matmatmul_wrapper` import is the cause for large perf regressions in Julia sparse benchmarks. It says this is due to the overrides we had in linalg.jl now being unused causing a fallback to a generic implementation.
See https://rustserver.perf.julialang.org/compare.html?start=56fae408d1a9b0364fddf9c561ca06427a74cccb&end=c76331c9274c4ac31c08a019e73ca0249c01d202&stat=min-wall-time for benchmark results

araujoms: Thanks for noticing, fixed in #757

PR #747, opened 2026-08-17
Fix breakage caused by JuliaLang/LinearAlgebra.jl#1668

yuyichao:
I don't think the new name for this function is the right decision but at least this should fix the breakage.

### DIFF
--- a/src/linalg.jl
+++ b/src/linalg.jl
@@ -4,6 +4,14 @@ using LinearAlgebra: AbstractTriangular, StridedMaybeAdjOrTransMat, UpperOrLower
     RealHermSymComplexHerm, HermOrSym, checksquare, sym_uplo, wrap
 using Random: rand!
 
+if isdefined(LinearAlgebra, :_uppercase)
+    const _uppercase = LinearAlgebra._uppercase
+    const _isuppercase = LinearAlgebra._isuppercase
+else
+    const _uppercase = uppercase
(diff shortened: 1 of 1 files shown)


araujoms:
https://github.com/JuliaLang/LinearAlgebra.jl/pull/1679 fixed it, but your PR is a better solution because we then get the advantages of `_uppercase`.

I have to apologise, I didn't realise that this was a public interface used by other packages.

Which name would you prefer, though? There is still time to change it.

src/linalg.jl
```diff
@@ -4,6 +4,14 @@ using LinearAlgebra: AbstractTriangular, StridedMaybeAdjOrTransMat, UpperOrLower
     RealHermSymComplexHerm, HermOrSym, checksquare, sym_uplo, wrap
 using Random: rand!
 
+if isdefined(LinearAlgebra, :_uppercase)
+    const _uppercase = LinearAlgebra._uppercase
+    const _isuppercase = LinearAlgebra._isuppercase
+else
+    const _uppercase = uppercase
+    const _isuppercase = isuppercase
+end
```
araujoms:
```suggestion
import LinearAlgebra: _uppercase, _isuppercase
```
I don't see the point of doing this `isdefined` mess, it's not as if `SparseArrays` can be used with other versions of `LinearAlgebra`.

PR #749, opened 2026-08-25
Improve `issymmetric`

### LINKED ISSUE #748: issymmetric still incorrect for some sparse matrices: #636's fix is incomplete
The fix for #636 covers the case where the mirror column is empty, but not the
case where it is non-empty and the tracker has run past its end.

```julia

julia> S = sparse([3,3,4,1,2], [1,2,2,3,4], ones(Int,5), 4, 4);

julia> S[2,3], S[3,2]
(0, 1)

julia> S == transpose(S)
false

julia> issymmetric(S)
true

julia> issymmetric(Matrix(S))
false
```

Reproduces on 1.10.11, 1.12.6 and 1.14.0-DEV.2617. #636's own example is fixed
on 1.12 and nightly, still broken on 1.10.

`is_hermsym` keeps a per-column `tracker` and, for a stored `A[row,col]`, reads
`rowval[tracker[row]]` to find `A[col,row]`. Here `colptr = [1,2,4,5,6]` and
`rowval = [3,3,4,1,2]`:

```
col=1 row=3: reads rowval[4]=1, column 3 spans 4:4
col=2 row=3: reads rowval[5]=2, column 3 spans 4:4   <-- outside column 3
```

`rowval[5]` belongs to column 4, so it concludes `A[2,3]` is stored and matches.
It isn't; `A[2,3]` is zero. The two guards are `isempty(nzrange(A, row))`, added
for #636, which catches an entirely empty column, and `offset > length(rowval)`,
which catches the end of the whole array. Neither catches running off the end of
a non-empty column.

Adding

```julia
offset > colptr[row+1] - 1 && return false
```

right after `offset = tracker[row]` gives the correct answer on this matrix. The
`while row2 < col` loop just below advances `offset` the same way and needs the
same bound.

Filed by Claude Code on my behalf — it turned this up chasing a sparse/dense
discrepancy in a modular decomposition package, where `issymmetric` was being
used to pick between the undirected and directed algorithms.

dkarrasch:
Fixes #748. Fix taken from the original issue report.


dkarrasch:
@StefanKarpinski Do you happen to have the second of the failing 4x4 matrices at hand?

StefanKarpinski:
I had claude post the cases in the issue; it may have gone a little overboard, but at least you have them now.

dkarrasch:
Feels good to be taken by the hand by a machine. :rofl: I was happy to have something copy-pastable!

dkarrasch:
This passes additional tests, so let's go with it.

dkarrasch:
When backporting, then we should first pick this one for the additional tests, then #749 for the actual algorithmic improvement.

PR #751, opened 2026-08-27
Make `sparse` and `spdiagm` statically dispatchable / `--trim` compatible

topolarity:
This allows these basic SparseArrays constructors to be `--trim` compatible:
```julia
    A = sparse([1, 2, 3], [1, 2, 3], [1.0, 2.0, 3.0])
    B = spdiagm(0 => ones(3), 1 => ones(2))
```

This commit was written with the assistance of generative AI (Claude) 🤖

### DIFF
--- a/src/sparsematrix.jl
+++ b/src/sparsematrix.jl
@@ -1069,7 +1069,7 @@ julia> sparse(Is, Js, Vs)
  ⋅  ⋅  3
 ```
 """
-function sparse(I::AbstractVector{Ti}, J::AbstractVector{Ti}, V::AbstractVector{Tv}, m::Integer, n::Integer, combine) where {Tv,Ti<:Integer}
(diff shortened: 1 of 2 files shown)


topolarity:
Tests for trimmability will be added downstream in the Julia dep bump

dkarrasch (dismissed):

ViralBShah:
Is this good to merge?

topolarity:
It is from my POV but I do not have merge permissions. @dkarrasch ?

araujoms:
@dkarrasch is taking a break.

Can't you add an `@inferred` test to make sure `--trim` will keep working?

topolarity:
We could, but `@inferred` checks the return type which was already good before this fix

araujoms:
As far as I understand what wasn't well-inferred before was the eltype of `V`, so an `@inferred` on `spdiagm_eltype` would be useful, no?

topolarity:
`@inferred` doesn't quite work due to a technicality with `Type{...}`, but I've added a `Base.infer_return_type` test now

araujoms:
Thanks. It's not quite what we needed but it'll have to do.

PR #753, opened 2026-09-02
Specialize `mapreduce` + `sparsevec` on function arguments

topolarity:
Another small trim fix-up, follow-up to #751 . Expect ~1 or 2 more PRs in this series.

I do not think it is easy to add a test here since the problem this fixes has to do with the "compileable" specializations chosen by the compiler - those affect codegen and cause `--trim` to fail but they don't generally impact inference.

### DIFF
--- a/src/sparsematrix.jl
+++ b/src/sparsematrix.jl
@@ -2402,7 +2402,7 @@ function Base.reducedim_initarray(A::AbstractSparseMatrixCSC, region, v0, ::Type
 end
 
 # General mapreduce
-function _mapreducezeros(f, op, ::Type{T}, nzeros::Integer, v0) where T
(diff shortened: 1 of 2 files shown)


araujoms:
You could write a comment explaining why the extra type parameters are needed - they look otherwise useless and somebody else might remove them.

fredrikekre (approved):
> You could write a comment explaining why the extra type parameters are needed

I don't think that is needed. This is a well known pattern and is used all over the place (https://docs.julialang.org/en/v1/manual/performance-tips/#Be-aware-of-when-Julia-avoids-specializing).

topolarity:
I agree with @fredrikekre but I am happy to add a note if you'd like @araujoms

PR #756, opened 2026-09-02
SPQR: build factors with typed constructors, for type-stability

topolarity:
Last PR of the `--trim` set! 

Having the types spelled out explicitly here allows these constructors to be fully inferred w.r.t. dispatch. This does not affect the return type of any call, so no `@inferred` test is included.

This commit was written with the assistance of generative AI (Claude) 🤖

### DIFF
--- a/test/spqr.jl
+++ b/test/spqr.jl
@@ -116,6 +116,13 @@ end
     @test eltype(F.Q) == eltype(F.R) == eltyA
 end
 
+@testset "single-precision qr factorization works as expected: $eltyA" for eltyA in (Float32, ComplexF32, Float16, ComplexF16)
+    A = sprandn(eltyA, m, n, 0.3)
+    F = qr(A)
+    @test eltype(F.Q) == eltype(F.R) == eltyA
+    @test Matrix(F.Q) * F.R ≈ A[F.prow, F.pcol]
+end
+
 @testset "select ordering overdetermined" begin
      A = sparse([1:n; rand(1:m, nn - n)], [1:n; rand(1:n, nn - n)], randn(nn), m, n)
      b = randn(m)
--- a/src/solvers/spqr.jl
+++ b/src/solvers/spqr.jl
@@ -220,9 +220,9 @@ function LinearAlgebra.qr(A::SparseMatrixCSC{Tv, Ti}; tol=_default_tol(A), order
         C_NULL, C_NULL, C_NULL, C_NULL,
         R, E, H, HPinv, HTau)
 
-    R_ = SparseMatrixCSC{Tv, Ti}(Sparse(R[]))
-    factors = SparseMatrixCSC{Tv, Ti}(Sparse(H[]))
-    τ = vec(Array{Tv}(CHOLMOD.Dense(HTau[])))
+    R_ = SparseMatrixCSC{Tv, Ti}(Sparse{Tv, Ti}(R[]))
+    factors = SparseMatrixCSC{Tv, Ti}(Sparse{Tv, Ti}(H[]))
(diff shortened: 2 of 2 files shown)


araujoms:
Wait, why does the function accept `Tv<:CHOLMOD.VTypes` if it doesn't actually support them? It should be restricted to `Float64` and `ComplexF64`, with `Float32` and `ComplexF32` handled by an explicit conversion like `Float16` and `ComplexF16` below.

topolarity:
You're right. This is broken right now:
```julia
julia> qr(SparseMatrixCSC{Float32}(sparse([1,2,3],[1,2,3],[1.0,2.0,3.0]))).R
3×3 SparseMatrixCSC{Float32, Int64} with 1 stored entry:
 2.0   ⋅    ⋅
  ⋅    ⋅    ⋅
  ⋅    ⋅    ⋅

julia> qr(SparseMatrixCSC{Float64}(sparse([1,2,3],[1,2,3],[1.0,2.0,3.0]))).R
3×3 SparseMatrixCSC{Float64, Int64} with 3 stored entries:
 1.0   ⋅    ⋅
  ⋅   2.0   ⋅
  ⋅    ⋅   3.0
```
  
I have pushed a quick fix to the branch

test/spqr.jl
```diff
@@ -116,6 +116,16 @@ end
     @test eltype(F.Q) == eltype(F.R) == eltyA
 end
 
+@testset "single-precision qr factorization works as expected: $eltyA" for eltyA in (Float32, ComplexF32, Float16, ComplexF16)
+    seed!(4)
+    A = SparseMatrixCSC{eltyA, Int}(eltyA <: Real ? sprandn(m, n, 0.3) : sprandn(ComplexF64, m, n, 0.3))
```
araujoms:
```suggestion
    A = sprandn(eltyA, m, n, 0.3)
```
Why such a complication?

test/spqr.jl
```diff
@@ -116,6 +116,16 @@ end
     @test eltype(F.Q) == eltype(F.R) == eltyA
 end
 
+@testset "single-precision qr factorization works as expected: $eltyA" for eltyA in (Float32, ComplexF32, Float16, ComplexF16)
+    seed!(4)
```
araujoms:
```suggestion
```
Why do you need a concrete random seed?

test/spqr.jl
```diff
@@ -116,6 +116,16 @@ end
...
+    A = SparseMatrixCSC{eltyA, Int}(eltyA <: Real ? sprandn(m, n, 0.3) : sprandn(ComplexF64, m, n, 0.3))
+    F = qr(A)
+    @test eltype(F.Q) == eltype(F.R) == eltyA
+    @test Matrix(F.Q) * F.R ≈ A[F.prow, F.pcol]
+    x = ones(eltyA, n)
+    @test F \ (A * x) ≈ x
```
araujoms:
```suggestion
```
Why would you expect this to hold? It doesn't it general, you shouldn't test for it.

topolarity: I think it holds for a full rank matrix, but yeah we should just do the simple factorization identity check.

test/spqr.jl
topolarity: AI got confused - Apologies, I should have reviewed the tests more closely.

test/spqr.jl
topolarity: Ah right, `Test` handles this automatically

araujoms (approved):
I'm relieved to see that Claude hasn't ascended to Godhood yet.

PR #762, opened 2026-09-06
CHOLMOD: fix getindex for unsorted and unpacked Sparse matrices

### LINKED ISSUE #758: `getindex(::Sparse, ...)` returns incorrect results for unsorted `Sparse` objects
```julia
julia> using SparseArrays, SparseArrays.CHOLMOD

julia> A = sparse([2, 1, 2], [1, 2, 2], Float64[1, 2, 3])

2×2 SparseMatrixCSC{Float64, Int64} with 3 stored entries:
  ⋅   2.0
 1.0  3.0

julia> S = CHOLMOD.Sparse(A); P = S' * S; # P ends up unsorted

julia> Matrix(sparse(P)) # good
2×2 Matrix{Float64}:
 1.0   3.0
 3.0  13.0

julia> P[1,2]
0.0
```

this affects some conversions as well:
```julia
julia> Array(P) # bad
2×2 Matrix{Float64}:
 1.0  0.0
 3.0  0.0
```

noticed while investigating the bug in https://github.com/JuliaSparse/SparseArrays.jl/pull/756

ViralBShah:
Fixes #758.

`getindex(::Sparse, i, j)` binary-searched the row indices of a column unconditionally, but CHOLMOD only guarantees that a column's row indices are sorted when the struct's `sorted` flag is set. `S'S` returns an unsorted result, so lookups silently missed stored entries — and since `Array` and `show` go through `getindex`, those conversions were wrong too:

```julia
julia> A = sparse([2, 1, 2], [1, 2, 2], Float64[1, 2, 3]);

julia> P = CHOLMOD.Sparse(A)' * CHOLMOD.Sparse(A); # unsorted

julia> P[1, 2]      # before: 0.0, after: 3.0
3.0

julia> Array(P)     # before: [1.0 0.0; 3.0 0.0]
2×2 Matrix{Float64}:
 1.0   3.0
 3.0  13.0
```

Changes:

* Use `searchsortedfirst` only when `sorted` is set; fall back to a linear scan over the column otherwise.
* While computing the range of entries belonging to a column, also handle unpacked matrices (`packed == 0`), where a column ends after `nz[j]` entries rather than at `p[j + 1]`. Previously those reads ran past the end of the column into uninitialized storage.

Tests cover the matrix from the issue plus an explicitly unsorted *and* unpacked matrix built with `allocate_sparse`, over both index types.

Note that `Sparse` is not a public API, so this is not user-facing today; it matters if/when we start relying on this indexing path internally.

https://claude.ai/code/session_014KpwdQ6bpWQKzjGV5prtAR

### DIFF
--- a/src/solvers/cholmod.jl
+++ b/src/solvers/cholmod.jl
@@ -1307,11 +1307,24 @@ function getindex(A::Sparse{T}, i0::Integer, i1::Integer) where T
     s.stype < 0 && i0 < i1 && return conj(A[i1,i0])
     s.stype > 0 && i0 > i1 && return conj(A[i1,i0])
 
+    # in an unpacked matrix the entries of column `i1` stop after `nz[i1]`
+    # entries rather than at `p[i1 + 1]`
     r1 = Int(unsafe_load(s.p, i1) + 1)
-    r2 = Int(unsafe_load(s.p, i1 + 1))
+    r2 = s.packed != 0 ? Int(unsafe_load(s.p, i1 + 1)) :
+                         r1 + Int(unsafe_load(s.nz, i1)) - 1
     (r1 > r2) && return zero(T)
-    r1 += Int(searchsortedfirst(view(unsafe_wrap(Array, s.i, (s.nzmax,), own = false), r1:r2), i0 - 1) - 1)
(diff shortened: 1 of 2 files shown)


ViralBShah:
@rayegun A quick review would be great.

PR #767, opened 2026-09-08
Export AbstractSparseMatrixCSC, getcolptr, indtype

j-fu:
AbstractSparseMatrixCSC and getcolptr are heavily used in LinearSolve.jl.

We find uses of getcolptr in Sparspak.jl and Pardiso.jl (and yes, in my package ExtendableSparse.jl)

indtype is a useful complement to eltype which could be extended for subtypes of AbstractSparseArray.SparseMatrixCSC

I intend to replace #395 with this PR. My main concern ist "ExplicitImports sanity" of the package ecosystem.  IMHO #395 was (perceived as) too ambitious, probably because of the word 'interface' in the title. The present PR doesn't care about "interface", but just pragmatically tries to formalize what is current practice.

As for 'indtype', this is a slightly different discussion, we may just make it public, and I would also agree with keeping it internal, in order to prevent delaying of fixing the situation with AbstractSparseMatrixCSC and getcolptr.

### DIFF
--- a/src/abstractsparse.jl
+++ b/src/abstractsparse.jl
@@ -75,6 +75,11 @@ end
 issparse(A::DenseArray) = false
 issparse(S::AbstractSparseArray) = true
 
+"""
+    indtype(S)
+
+Return the type used to index sparse array entries.
+"""
 indtype(S::AbstractSparseArray{<:Any,Ti}) where {Ti} = Ti
 indtype(T::UpperOrLowerTriangular{<:Any,<:AbstractSparseArray}) = indtype(parent(T))
 
--- a/src/sparsematrix.jl
+++ b/src/sparsematrix.jl
@@ -194,10 +194,11 @@ const SparseMatrixCSCUnion2{Tv,Ti} = Union{AbstractSparseMatrixCSC{Tv,Ti}, Spars
 """
         getcolptr(S)
 
-Return the vector of column start indices in an AbstractSparseMatrixCSC pointing
-into [`nonzeros`](@ref) and [`rowvals`](@ref). Any modifications to the returned
-vector will mutate `A` as well. Providing access to the column start indices
-can be useful in preconditioners and sparse direct solvers.
+Return the vector of column start indices in an `AbstractSparseMatrixCSC` pointing
+into [`nonzeros`](@ref) and [`rowvals`](@ref). The returned vector aliases `S`,
+but implementations with fixed sparsity may make it read-only. When it is
+writable, modifications to it mutate `S`. Providing access to the column start
+indices can be useful in preconditioners and sparse direct solvers.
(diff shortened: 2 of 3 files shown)


ViralBShah:
Let’s get it done.

j-fu:
Should be ok now.

ViralBShah:
I am ok with merging this. Let's give it a few days for feedback since we are exporting a couple new things here.

ViralBShah:
@rayegun Can we get your eyes on this one since it is exporting a new API?

src/sparsematrix.jl
```diff
@@ -177,6 +177,49 @@ function _goodbuffers(m, n, colptr, rowval, nzval)
...
+const SparseMatrixCSCUnion2{Tv,Ti} = Union{AbstractSparseMatrixCSC{Tv,Ti}, SparseMatrixCSCColumnSubset{Tv,Ti}}
+
+"""
+        getcolptr(S)
+
+Return the vector of column start indices in an `AbstractSparseMatrixCSC` pointing
+into [`nonzeros`](@ref) and [`rowvals`](@ref). The returned vector aliases `S`,
+but implementations with fixed sparsity may make it read-only. When it is
+writable, modifications to it mutate `S`. Providing access to the column start
+indices can be useful in preconditioners and sparse direct solvers.
```
matbesancon, line 204: Not sure the last sentence about usefulness outside the code is necessary here / the good place for it

src/abstractsparse.jl
```diff
@@ -142,6 +142,11 @@ end
 issparse(A::DenseArray) = false
 issparse(S::AbstractSparseArray) = true
 
+"""
+    indtype(S)
+
+Return the type used to index sparse array entries.
+"""
```
ViralBShah: Given that we are using getrowval, getnzval, getcolptr etc., should we use getindtype - to be consistent for all accessor functions?

ViralBShah:
I'd like to have this in for 1.14 - for which the freeze is about 3 weeks away. We have enough time and we are getting some eyeballs now.

src/sparsematrix.jl
j-fu: This was inspired by a similar "usefulness" statement in the `rowvals` docs.  I agree that this is not particularly necessary.

src/abstractsparse.jl
j-fu: I tend to keep  `indtype` as it is like `eltype` and in this way appears to be rather intuitive.

ViralBShah:
@j-fu Let me know if you'd like me to rebase this PR given the substantial changes I have been making. I think we should merge it soon.

j-fu:
Yes, I think this can be rebased.

PR #769, opened 2026-09-08
Fix `findnext`/`findprev` for sparse arrays to extend `Base`, making `hash` fast

### LINKED ISSUE #570: hash(SparseVector) very slow
```julia
const sp=spzeros(Int,1_000_000_000)
1000000000-element SparseVector{Int64, Int64} with 0 stored entries

@btime hash(sp)
  182.608 ms (0 allocations: 0 bytes)

```

I can create a PR with the alternative implementation if you agree with it
```julia
@btime hash((sp.nzind, sp.nzval, sp.n))
  4.208 ns (0 allocations: 0 bytes)
```

ViralBShah:
Fixes #570 

Follow-up to #766 and #768, addressing the last relative of `isequal` that still visited every element: `hash`.

**Root cause.** The sparse `findnext`/`findprev` methods in `src/abstractsparse.jl` have been dead code since the JuliaLang/julia import cleanup (JuliaLang/julia#42894) dropped `findnext, findprev` from the `import Base:` list. They defined a local `SparseArrays.findprev` instead of extending `Base.findprev`, so every call fell through to the generic elementwise method. The existing tests compared against dense results and so never noticed.

**Why this matters for `hash`.** `Base.hash` on large arrays (`_hash_fib`) skips runs of equal values with `findprev(!isequal(elt), A, i)`. On a sparse array that call is where all the time goes. With the import restored it jumps straight to the previous stored entry.

| Call | Before | After |
|---|---|---|
| `hash` of a length `10^9` sparse vector, one stored entry | ~1 s | 2 µs |
| `hash` of a `10^4 x 10^4` sparse matrix, one stored entry | 81 ms | 4 µs |

Hash values are unchanged: this PR does not touch `hash` itself, so a sparse array still hashes identically to its dense counterpart, and explicitly stored zeros do not affect the value. Verified with randomized checks including NaN and signed zeros, and now covered by tests.

Changes:
- Add `findnext, findprev` back to the `import Base:` list.
- Test that the sparse methods actually extend `Base`, and that `findnext`/`findprev` with `!isequal(x)` predicates agree with dense for NaN and signed zeros, including a timing guard on a `10^9` vector.
- Test that `hash` of sparse vectors and matrices matches dense (with and without a seed, with stored zeros, NaN and `-0.0`), plus a timing guard.

Based on #768 because both touch the same test files; it will rebase cleanly onto `main` once that merges. Full test suite passes on nightly and `Test.detect_ambiguities(SparseArrays)` remains empty.

https://claude.ai/code/session_01LgBHUw9Hp7YW5ub29B4R5y


ViralBShah:
This should close https://github.com/JuliaSparse/SparseArrays.jl/issues/570.

@fredrikekre Can I ask you for a review?

PR #783, opened 2026-09-08
Reduce type-loop breadth in two slow testsets

ViralBShah:
Item 5 of the test-time review, narrowed to the two changes that remove duplicated work without touching coverage.

- **`cholmod_ops.jl` "Core functionality"**: the testset iterated `Ti ∈ itypes` even though `Ti` is already the file-level loop variable, so it ran twice per outer iteration with identical arguments (64 runs on 64-bit). It now uses the outer `Ti`: 32 runs, same coverage.
- **`linalg_products.jl` "kronecker product"**: one matrix size instead of two. The 5x5 structured-matrix product and the three wrappers inside are unchanged.

The `test_mul` alpha/beta grid, the "dimension mismatch error" constructor set and the `sparsevector.jl` `fill!` loop are left as they are on `main`.

https://claude.ai/code/session_01QrQuwcdC8fGXs3czT5KxN5

### DIFF
--- a/test/cholmod_ops.jl
+++ b/test/cholmod_ops.jl
@@ -27,11 +27,12 @@ itypes = sizeof(Int) == 4 ? (Int32,) : (Int32, Int64)
 for Ti ∈ itypes, Tv ∈ (Float32, Float64)
 Random.seed!(123)
 
+# `Ti` and `Tv` come from the file-level loop; iterating `Ti` here as well used to run
+# this testset twice per outer iteration with identical arguments.
 @testset "Core functionality ($elty, $elty2)" for
     elty in (Tv, Complex{Tv}),
     Tv2 in (Float32, Float64),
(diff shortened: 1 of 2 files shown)


test/cholmod_ops.jl
```diff
@@ -27,11 +27,12 @@ itypes = sizeof(Int) == 4 ? (Int32,) : (Int32, Int64)
 for Ti ∈ itypes, Tv ∈ (Float32, Float64)
 Random.seed!(123)
 
+# `Ti` and `Tv` come from the file-level loop; iterating `Ti` here as well used to run
```
KristofferC, line 30: Typically, you don't want comments describing some previous state of the code. 

AIs like to add these type of comments for some reason.

ViralBShah: Yes, it seems they seem fundamental to how these tools operate - but they are a bit annoying.

PR #798, opened 2026-09-09
Add a `sparse` keyword to reductions along a dimension and reduce views, adjoints and column subsets through the sparse kernels

### LINKED ISSUE #43: sum(sparse) -> dense?
```julia
julia> A = sparse([rand() < 0.01 ? 1 : 0 for _ in 1:50, _ in 1:50])
50×50 SparseMatrixCSC{Int64,Int64} with 23 stored entries:
[...]

julia> sum(A, dims=1)
1×50 Array{Int64,2}:
 0  1  0  0  0  0  0  0  1  0  1  0  0  0  0  1  0  1  0  0  1  1  0  0  0  2  3  0  0  0  0  0  0  0  0  0  0  0  1  1  1  0  0  1  1  0  2  1  0  3
```

Sometimes you have a very sparse matrix and want to sum a slice of it. The slice may potentially have very many columns or rows that are entirely zero, in which case it makes a lot of sense to preserve sparsity in the output.

In the non-slice case, where every row and column usually contains at least one nonzero value, it makes sense to keep the result dense.

It would be great if preserving sparsity in `sum` and other similar reductions were allowed via a keyword argument, eg. `sparse=true`.

cc @simonbyrne, who suggested the keyword idea on Slack a few months ago.

### LINKED ISSUE #377: New method for `sum` for `SubArray` of `SparseMatrix`
I often find myself needing to operate on a subset of a sparse array, typically a few adjacent columns. Creating a new `SparseMatrix` is relatively expensive, so being able to operate on view would be a big win.

Here's an example

```julia
s = sprand(101, 99, 0.01);

sc = s[:, 5:11];
sv = view(s, :, 5:11);

# sum parent array
@btime sum($s, dims=1);  # 400.706 ns (4 allocations: 1.02 KiB)
@btime sum($s, dims=2);  # 575.319 ns (9 allocations: 1.97 KiB)

# sum copy of columns
@btime sum($sc, dims=1);  # 244.282 ns (4 allocations: 256 bytes)
@btime sum($sc, dims=2);  # 454.545 ns (9 allocations: 1.97 KiB)

# sum view of columns
@btime sum($sv, dims=1);  # 1.179 μs (5 allocations: 416 bytes)
@btime sum($sv, dims=2);  # 6.442 μs (9 allocations: 1.88 KiB)
```

It looks like there is a specialized method for 1D column views. It would be great to extend this to groups of columns.

ViralBShah:
Fixes #43. Fixes #377.

**Issue.** `sum(A; dims = 2)` on a hypersparse matrix allocates a dense result proportional to `size(A, 1)` however few entries `A` stores (#43). Reductions of a column-range view went through Base's element-wise fallback and returned a sparse result, since Base's `similar` on a `SubArray` defers to the parent (#377).

```julia
A = sparse([1, 10^7], [1, 150], [1.0, 2.0], 10^7, 150)
@time sum(A; dims = 2)            # 15 ms, 153 MiB, for two stored entries
B = sprand(101, 99, 0.01)
sum(view(B, :, 5:11); dims = 1)   # SparseMatrixCSC, element by element; the copy gives a Matrix
```

**Fix.** The default stays a dense `Matrix`, as for dense input (see the discussion below). A sparse result is opt-in: `sum(A; dims = 2, sparse = true)`, and likewise for `prod`, `maximum`, `minimum`, `count`, `any`, `all` and `mapreduce`, with `init` where they accept it. It stores an entry for each row or column that stores one, or for every slice when an unstored slice reduces to something nonzero, with the element type of the dense result. Column-range views reduce through the sparse kernels and return the same `Matrix` as their copy; `nnz` of such a view is O(1).

**Mechanism.** Base forwards unknown keywords from `sum`, `prod`, `maximum` and `minimum` to `mapreduce`, so one `mapreduce` method carries the keyword; `any`, `all` and `count` get their own. Slices are seeded as Base seeds the dense result, with `init` or `mapreduce_first`, and unstored entries are folded in through the existing `_mapreducezeros`. Element type and the value of an empty slice come from Base's `reducedim_init` on a stand-in. Row reductions build the result column by column; column reductions use a dense workspace, or sort the stored entries by row when there are fewer than `m / 8` of them.

**Dispatch.** New `sparse`-keyword methods for `Base.mapreduce`, `any`, `all` and `count` on `SparseMatrixCSCUnion`; the existing sparse reduction methods widen from `AbstractSparseMatrixCSC` to `SparseMatrixCSCUnion`. Behaviour change: reductions along a dimension of a column-range or column-subset view, or of an adjoint or transpose, now return a `Matrix`. `extrema` with `sparse = true` throws an `ArgumentError`, since a tuple has no zero.

**Adjoints, column subsets, vectors.** `sum(A'; dims)` and reductions of `view(A, :, [1, 5, 9])` had the same defect as #377: a fully stored sparse destination from `similar`, filled through sparse `setindex!` or element by element. They now get the dense destination of their copy. The kernels walk `nzrange` rather than the column pointers, so they take `SparseMatrixCSCOrColumnSubset` (no change in time for a plain matrix or a column-range view); the whole-array reductions gather the stored positions of a non-contiguous subset. LinearAlgebra already forwards the commutative reductions of an adjoint or transpose to the parent; a `_mapreducedim!` method covers the remaining `op`s the same way. `sparse = true` is accepted for all of these (an adjoint reduces its parent along the other dimension and `permutedims` the result) and for sparse vectors, which return a `SparseVector`. The keyword methods are generated for the three existing argument types, so there is no new alias. Reducing both dimensions of an adjoint folds a copy, so that a non-commutative `op` sees the adjoint's element order. For 200000 x 2000 at density 1e-5: `sum(A'; dims = 1)` 64 ms to 0.2 ms, `sum(view(A, :, [1, 5, 9]); dims = 2)` 4.7 ms to 0.1 ms (0.4 µs with `sparse = true`). Tests count the calls to `f` to show these reach the kernels. Not covered: row-range views, `Symmetric` and triangular wrappers, and the adjoint of a sparse vector, which still reduce element by element.

**Measured** on 1.14-DEV, min over samples:

| case | dense default | `sparse = true` |
|---|---|---|
| `sum(A; dims=2)`, 10^7 x 150, 151 entries | 15.1 ms, 153 MiB | 2.4 µs, 16 KiB |
| `sum(B; dims=1)`, 10^4 x 10^4, 1e-2 | 0.16 ms | 0.19 ms |
| `sum(B; dims=2)`, 10^4 x 10^4, 1e-2 | 1.26 ms | 1.57 ms |

The #377 example, 101 x 99 at density 0.01, columns 5:11: view within 10% of the copy for `sum`, `maximum` and `count` along either dimension, versus the element-wise fallback before.

**Tests** compare every reduction with the dense result over shapes, densities and `dims`, for the matrix, its column-range view and with `sparse = true`, real and complex; `init`; the stored pattern; element types for small integers and `Bool`; empty dimensions; an allocation bound on the hypersparse path; `@which` for the view kernels; and the errors. Full suite, whitespace and Aqua pass locally on 1.14-DEV. Docs: a paragraph in `docs/src/index.md`. Not for backport.

**Left out.** Two pre-existing corners: `fill!(spzeros(0, n), x)` with `x != 0` throws in `_fillnonzero!`, and `minimum(spzeros(3, 0); dims = 1)` returns a sparse `1 x 0` result through Base's `map`.

https://claude.ai/code/session_01DGaTC2P39YrE2aAU5es55x

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -186,6 +186,12 @@ In many cases it may be better to convert the sparse matrix into `(I,J,V)` forma
 manipulate the values or the structure in the dense vectors `(I,J,V)`, and then reconstruct
 the sparse matrix.
 
+Reductions along a dimension, such as `sum(S; dims = 2)`, return a dense `Matrix`, as for dense
+input. To keep the result sparse, reduce into a sparse destination: `sum!(spzeros(size(S, 1), 1), S)`
+stores an entry only for the rows of `S` that store one, at a cost proportional to the number of
+stored entries rather than to the number of rows. `Base.mapreducedim!` and the other in-place
+reductions accept a sparse destination in the same way.
+
 ## Correspondence of dense and sparse methods
 
 The following table gives a correspondence between built-in methods on sparse matrices and their
--- a/src/sparsematrix.jl
+++ b/src/sparsematrix.jl
@@ -2495,11 +2495,20 @@ Base.isequal(A::Transpose{<:Any,<:AbstractSparseMatrixCSCInclAdjointAndTranspose
 
 ## Reductions
 
-# In general, output of sparse matrix reductions will not be sparse,
-# and computing reductions along columns into SparseMatrixCSC is
-# non-trivial, so use Arrays for output. Array element type is given by `R`.
-function Base.reducedim_initarray(A::AbstractSparseMatrixCSC, region, v0, ::Type{R}) where {R}
-    fill!(Array{R}(undef, Base.to_shape(Base.reduced_indices(A, region))), v0)
+# Reductions along a dimension of a sparse matrix return a sparse matrix (issue #43): the
+# result only stores an entry for the rows or columns that store one themselves, unless
+# the reduction of a structurally empty slice is nonzero. The initial array is therefore
+# structurally empty when the initial value is zero, and fully stored otherwise.
+function Base.reducedim_initarray(A::AbstractSparseMatrixCSC{<:Any,Ti}, region, v0, ::Type{R}) where {R,Ti}
+    m, n = Base.to_shape(Base.reduced_indices(A, region))
+    if !applicable(zero, R)
+        # no structural zero for the result, e.g. the tuples of `extrema`: reduce densely
+        return fill!(Array{R}(undef, m, n), v0)
+    elseif isequal(v0, zero(R))
+        return spzeros(R, Ti, m, n)
+    else
+        return SparseMatrixCSC(m, n, Ti[1 + m*j for j in 0:n], repeat(Ti.(1:m), n), fill!(Vector{R}(undef, m*n), v0))
+    end
 end
 
 # General mapreduce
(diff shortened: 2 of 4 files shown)


rasmushenningsson:
I'm coming from a field where the sparse matrices are not hyper-sparse and this obviously feels like a strange default to me.

The only case where I see a big benefit of the proposal is when you have many more rows than columns, very low density, and reduce with `dims=2`. In all other cases a dense output is the natural choice because when we lose when dimension, we have orders of magnitude fewer elements, but the fraction of non-zeros easily goes to 1. And for **downstream** operations, you don't want to end up with sparse arrays used as input when the data is dense.

Instead, the natural choice to me would be to opt-in to this behavior, because in a limited number of situations, it is super useful.

ViralBShah:
Agree - and the CSC storage format is not a good fit for hypersparse anyways.

ViralBShah:
Made the sparse result opt-in, as suggested. `sum(A; dims)` and friends return a dense `Matrix` again. The sparse result is a keyword: `sum(A; dims = 2, sparse = true)`, and likewise for `prod`, `maximum`, `minimum`, `count`, `any`, `all` and `mapreduce`. That keeps the hypersparse kernels for the cases where they pay off (the 10^7-row example goes from 15 ms and 153 MiB to 2.4 µs) without changing the default anyone relies on. Column-range views now return the same dense result as their copy, where Base's `similar` gave them a sparse one before. Description updated.

Issue #301, opened 2022-11-27
A \ b doesn't return the minimum norm solution for sparse least-norm problems


amontoison:
If `A` has more columns than rows, we should compute the `QR` factorization of `A'` such that we recover an LQ factorization of `A`.

stevengj:
Unfortunately, we don't yet have an `ldiv!` implementation for the transpose of a sparse QR factorization, which is needed for this feature: #656

PR #804, opened 2026-09-09
Solve with the adjoint of a sparse QR factorization

### LINKED ISSUE #656: missing ldiv for transpose of sparse QR
This would be useful to find the minimum-norm solution of a "wide" (underdetermined) system of equations:
```jl
julia> qr(sprand(3,100,0.1)')' \ rand(3)
ERROR: MethodError: no method matching ldiv!(::LinearAlgebra.AdjointFactorization{Float64, SparseArrays.SPQR.QRSparse{Float64, Int64}}, ::Vector{Float64})
The function `ldiv!` exists, but no method is defined for this combination of argument types.
```
Here is a simple example implementation of this (see [this explanation](https://www.netlib.org/lapack/lug/node41.html)):
```jl
function myldiv(Fadj::LinearAlgebra.AdjointFactorization{<:Number, <:SparseArrays.SPQR.QRSparse}, b::AbstractVector{<:Number})
    F = parent(Fadj)
    m, n = size(F)
    y = [UpperTriangular(F.R)' \ b[F.pcol]; zeros(m - n)]
    xperm = F.Q * y
    x = similar(xperm); x[F.prow] = xperm
    return x
end
```
(Needed for #301.)

Example usage:
```jl
julia> A = sprand(3,100,0.1)
3×100 SparseMatrixCSC{Float64, Int64} with 33 stored entries:
⎡⠀⠀⠀⠁⠈⠀⠀⠀⠀⠀⡀⡀⠀⢀⠀⠀⢀⢁⠀⠀⠀⢁⠈⠀⢈⡀⠀⢀⠀⠀⠀⠀⠀⠀⠈⠀⠀⠈⠀⣁⎤
⎣⠀⢀⠀⠀⢀⠀⠀⠀⠀⡀⠀⡀⠀⠀⠀⢀⠀⠀⠀⠀⢀⠀⠀⠀⠀⠀⢀⠀⠀⡀⠀⢀⠀⠀⠀⢀⡀⠀⡀⠀⎦

julia> b = randn(3);

julia> myldiv(qr(A')', b) ≈ Matrix(A) \ b
true
```

ViralBShah:
Fixes #656, and completes the remaining half of #115 (`qr(A')` already works).

`qr(A)' \ b` previously threw

```
ERROR: MethodError: no method matching ldiv!(::AdjointFactorization{Float64, QRSparse{Float64, Int64}}, ::Vector{Float64})
```

This adds `ldiv!(X, ::AdjointFactorization{<:Any,<:QRSparse}, B)` plus the corresponding `\` methods, so a single factorization of `A` can be used to solve with both `A` and `A'` — the use case from #115.

### What it computes

With `A[prow, pcol] == Q*R` we have `A' == Pcol*R'*Q'*Prow`, so solving `A'x = b` is a forward substitution with `R'` followed by a multiply by `Q` and the inverse row permutation. It reuses the permutations, workspace and lock of the existing `ldiv!`, so a solve allocates only the result.

For a tall `A` the system `A'x = b` is underdetermined and the **minimum-norm** solution is returned (the trailing components of `Q'x` are zeroed), matching `qr(A)' \ b` for dense `A` and `pinv(A') * b`:

```jl
julia> A = sprand(3, 100, 0.1);

julia> b = randn(3);

julia> qr(sparse(A'))' \ b ≈ Matrix(A) \ b
true
```

When `A` is rank deficient, zeroing the free variables also drops the equations that the leading rank block of `R` cannot represent — the counterpart of the basic solution that `qr(A) \ b` already returns.

Wide `A` is rejected with the same `DimensionMismatch("overdetermined systems are not supported")` that dense QR throws, since that solve needs a least-squares factorization of `A'` rather than of `A`.

### Also fixed

`\(::Adjoint{<:Any,<:AbstractSparseMatrixCSC}, B)` and the `Transpose` variant already routed non-square matrices through `adjoint(qr(A)) \ B`, so `A' \ b` and `transpose(A) \ b` threw a `MethodError` for any non-square sparse `A`. They now work for every shape: a tall `A` uses the new method above and returns the minimum-norm solution, while a wide `A` factorizes the sparse transpose and returns the least squares solution, like the dense path.

### Relation to #301

#301 wants `A \ b` to return the minimum-norm solution for wide `A`, which needs exactly this method (`qr(sparse(A'))' \ b`). That is a behavior change to `\` and is left for a separate PR; this one supplies the missing piece.

### Tests

Added to the existing `spqr.jl` testsets: minimum-norm agreement with the dense `Array(A)' \ B` over the element-type matrix (real/complex factorization × `Int`/real/complex rhs, vector and matrix), the `Transpose` and `A' \ b` paths, coverage across all `SPQR.ORDERINGS` (which exercises the empty-`cpiv` `ORDERING_FIXED` case), and the dimension-mismatch errors.

https://claude.ai/code/session_01CSC6hQqcWDvAuvdkXJstBM


ViralBShah:
This change was accidentally included in the squash merge of #810. #817 reverts that portion from main; once it lands, this PR merges cleanly against main again and can be reviewed as intended.

PR #805, opened 2026-09-10
Define getrowval and getnzval for sparse vectors

### LINKED ISSUE #172: `getrowval` and `getnzval` are defined on SparseMatrixCSC but not on SparseVector
`rowval` is defined on both, but curiously `getrowval` is not. I'm not sure exactly why we have two identical accessor functions.

ViralBShah:
Fixes #172.

`rowvals(x::SparseVectorUnion)` already existed, but `getrowval` and `getnzval` (the accessors used throughout `linalg.jl`) only had `AbstractSparseMatrixCSC` methods. This adds the `SparseVectorUnion` methods, forwarding to `nonzeroinds` and `nonzeros`, so both accessor families cover both matrices and vectors (including column views and whole-vector views).

It also documents and exports `getrowval` and `getnzval`, whose names match the `rowval` and `nzval` fields of `SparseMatrixCSC`:

- New docstrings with matrix and vector examples, added to the API section of the docs.
- Both names exported.
- Triangular-wrapper forwards added so the new names cover everything `rowvals` and `nonzeros` do.
- Existing uses of `rowvals` and `nonzeros`, including the `nzrange` example, are left as they are, and no deprecation of `rowvals` or `nonzeros` is announced here.

Tests added to the existing `basic properties` testset in `test/sparsevector.jl`. Doctests and the Documenter build pass locally.

https://claude.ai/code/session_01Lbrk9jqFT4HCx4ENQdtx6f

### DIFF
--- a/src/sparsematrix.jl
+++ b/src/sparsematrix.jl
@@ -194,10 +194,80 @@ const SparseMatrixCSCUnion2{Tv,Ti} = Union{AbstractSparseMatrixCSC{Tv,Ti}, Spars
 getcolptr(S::SorF)     = getfield(S, :colptr)
 getcolptr(S::SparseMatrixCSCView) = view(getcolptr(parent(S)), first(S.indices[2]):(last(S.indices[2]) + 1))
 getcolptr(S::SparseMatrixCSCColumnSubset) = error("getcolptr not well-defined for $(typeof(S))")
+"""
+    getrowval(A)
+
+Return the vector of row indices of the structural nonzeros of sparse array `A`.
+For a [`SparseMatrixCSC`](@ref) this is the `rowval` field; for a
+[`SparseVector`](@ref) it is the `nzind` field. Any modifications to the returned
+vector will mutate `A` as well. Providing access to how the row indices are
+stored internally can be useful in conjunction with iterating over structural
+nonzero values. See also [`getnzval`](@ref) and [`nzrange`](@ref).
+
+`getrowval` is the preferred name for this accessor; [`rowvals`](@ref) is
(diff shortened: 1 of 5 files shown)


matbesancon:
@ViralBShah I just fixed a conflict with the same implementation of rowvals but I realize that it would be preferable to homogenize all methods to use a `SparseVectorOrView` argument?

ViralBShah:
@matbesancon agreed, done in 947cb73: the new vector methods and the triangular forwards now dispatch on `SparseVectorOrView` / `SparseMatrixCSCOrView`, matching `rowvals` and `nonzeros` (the `*Union` names are only kept for downstream). Also merged `main` (the `cscimport` doctest conflict resolves to main's version) and added a `LowerTriangular` case for the lines Codecov flagged.

_Written by Claude Code._

src/sparsematrix.jl
```diff
@@ -279,12 +347,12 @@ rowvals(S::LowerTriangular{<:Any,<:SparseMatrixCSCOrView}) = rowvals(S.data)
...
-[`rowvals`](@ref), this allows for convenient iterating over a sparse matrix :
+of sparse array `A`. In conjunction with [`getrowval`](@ref) and
+[`getnzval`](@ref), this allows for convenient iterating over a sparse matrix :
 
     A = sparse(I,J,V)
-    rows = rowvals(A)
-    vals = nonzeros(A)
+    rows = getrowval(A)
+    vals = getnzval(A)
     m, n = size(A)
```
ViralBShah, line 356: Let's keep these as rowvals and nonzeros and not make changes just yet.

PR #816, opened 2026-09-11
Rationalize the dispatch type aliases

ViralBShah:
The `Union` aliases used for dispatch had grown to 36 across 7 files, with duplicates, near-duplicates that hid semantic differences, and no naming convention. This collects the cross-file ones into one section of `abstractsparse.jl`, drops the duplicates, reuses LinearAlgebra's aliases where they already exist, and gives the survivors one scheme. Solver scalar-type aliases stay with their solvers.

### Before and after

Counting every `const X = Union{...}` in `src`:

| | `main` | this PR |
|---|---|---|
| All `Union` aliases | 44 | 32 |
| Solver scalar / C-struct types (`ITypes`, `VTypes`, `UMFITypes`, ...) | 10 | 9 |
| Dispatch aliases | 34 | 23 |

The 23 dispatch aliases are now: 9 in the shared section of `abstractsparse.jl` (`AbstractSparseVecOrMat`, `SparseVecOrMat`, `SparseMatrixCSCOrView`, `SparseMatrixCSCOrColumnSubset`, `SparseVectorOrView`, two `<X>MaybeAdjOrTrans`, `SparseOrTri`, `LinAlgLeftQs`), 8 local to `linalg.jl` (`DenseMatrixUnion`, `DenseInputVector`, `MatrixWrappers`, `MatrixWrappersOrView`, `QuasiSparseMatrix`, `QuasiStridedMatrix`, the two kron groups), 3 in CHOLMOD (`StridedVecOrMatMaybeAdjOrTrans`, `SparseVectorOrMatrixCSC`, `FactorComponentRHS`), `UMFAdjOrTransLU` in UMFPACK, and `SparseVecOrMatStyle`, `SparseOrStructuredMatrix` and the two concat groups. Placement follows one rule: used from more than one file goes in the shared section, used across a file goes at the top of that file, used by one section sits at the head of that section. A further three aliases (`SparseTriangular`, `SparseMatrixCSCSymmHerm`, CHOLMOD's `RealHermSymComplexHermSSL`) no longer spell out a `Union` because they are built on a LinearAlgebra alias.

### Consolidation

- `SparseVecOrMat` / `_SparseArraysCSC`, `StructuredMatrix` / `_SpecialArrays`, `_SparseVectorUnion` / `SparseVectorUnion`, `DenseViewWrappers` / `WrapperMatrixTypes`, and `SparseTriangular` / `AbstractTriangularSparse` each collapse to one alias.
- "X or `Adjoint`/`Transpose` of X" is spelled `<X>MaybeAdjOrTrans` throughout (after LinearAlgebra's `StridedMaybeAdjOrTransMat`) and built from `AdjOrTrans`.
- "X or a view of X" is spelled `<X>OrView` / `<X>OrColumnSubset`. Alias names use the `Sparse` family name and accept the abstract tier of their base type; only the concrete CHOLMOD alias says otherwise in its name: `SparseMatrixCSCUnion` becomes `SparseMatrixCSCOrView`, `SparseMatrixCSCUnion2` becomes `SparseMatrixCSCOrColumnSubset`, `SparseVectorUnion` becomes `SparseVectorOrView`, and `AdjOrTransSparseVectorUnion` becomes `AdjOrTransSparseVectorOrView`. The two-tier CSC split stays: the `getcolptr`-based kernels need a unit-range column view, the `nzrange`-based ones accept any column subset (#476).
- `SorF` is gone; `SparseMatrixCSC` and `FixedSparseCSC` each define their accessors.
- The seven one-off kron intermediates are folded into `_SparseKronGroup` / `_DenseKronGroup`.
- The inline spellings of the vec-or-mat union in the broadcast style rules, `fkeep!` and `fill!` use `SparseVecOrMat`; the two adjoint/transpose `BroadcastStyle` rules become one.
- CHOLMOD's concrete `SparseVecOrMat` is renamed `SparseVectorOrMatrixCSC` so it no longer shares a name with the top-level abstract alias. It stays concrete because `Sparse(B)` only accepts those two types.
- The remaining cryptic or misleading local names are renamed: UMFPACK's `ATLU` becomes `UMFAdjOrTransLU`, `SPVM` becomes `SparseVecOrMatStyle`, and `WrapperMatrixTypes` (which is `MatrixWrappers` plus `SubArray`) becomes `MatrixWrappersOrView`. The single-use `SuiteSparseStruct` is inlined.
- `SparseOrStructuredMatrix` is built on `SparseMatrixCSCOrView` instead of listing `FixedSparseCSC`, `SparseMatrixCSC` and `SparseMatrixCSCView` explicitly, so `map` over any `AbstractSparseMatrixCSC` subtype mixed with structured matrices takes the sparse path. The pre-existing all-sparse `map`/`map!` disambiguation methods widen from `Vararg{SparseMatrixCSC}` to `Vararg{AbstractSparseMatrixCSC}` to stay more specific than both competing methods.

### Reusing LinearAlgebra's names

- `BiTriSym` was redefined identically; it is now imported from LinearAlgebra.
- `HigherOrderFns.StructuredMatrix` shadowed the larger `LinearAlgebra.StructuredMatrix` in the same file. Its set (`Diagonal`, `Bidiagonal`, `Tridiagonal`, `SymTridiagonal`) is exactly `LinearAlgebra.BandedMatrix`, which is used instead.
- CHOLMOD's `RealHermSymComplexHermSSL` is now `RealHermSymComplexHerm{Tr,<:SparseMatrixCSC{<:Any,Ti}}`; dispatch is unchanged.

### Downstream-facing names

`SparseMatrixCSCView`, `SparseColumnView` and `SparseVectorView` are unchanged and used both internally and by downstream packages (LowRankOpt, IntervalMDP, Distances, and unregistered ones). `SparseMatrixCSCUnion` (Enzyme, SparseLinearAlgebra) and `SparseVectorUnion` (FillArrays, ArrayLayouts, Distances, FrankWolfe, SparseLinearAlgebra) are no longer used internally but are kept as aliases of the new names, with a note that they may be deprecated in a future release. `AdjOrTransSparseVectorUnion` has no external user on GitHub or JuliaHub and is dropped.

### Behavior changes

- `SparseTriangular` now uses `UpperOrLowerTriangular`, so `UnitUpperTriangular`/`UnitLowerTriangular` wrappers of sparse matrices take the Gustavson product path like the non-unit ones (they are materialized via `sparse(A)` first; the temporary widens to `Int` indices when the added diagonal would not fit the parent's index type, and the result keeps the operands' index type). Tri × Tri products keep LinearAlgebra's result wrapper, including `UnitUpper * UnitUpper` staying unit.
- The sparse-vector fast path dispatches on `SparseVectorOrView` rather than `AbstractSparseVector`, so custom `AbstractSparseVector` types outside the kernel's supported set take the generic product. This also fixes `UpperTriangular(S) * v` for such vectors, a `MethodError` on `main`.
- `nnz` and `indtype` are extended to triangular wrappers of column-range views. On `main`, `UpperTriangular(view(S, :, 1:n)) * S` and 14 similar combinations throw `MethodError`; they now return correct sparse results.

A scratchpad probe of 1652 product, kron, dot, broadcast, conversion and solve combinations across sparse, view, wrapper and dense operands shows no result-type or correctness change for anything that worked on `main`, plus the 15 former `MethodError`s now passing. Full suite and Aqua/ambiguity check pass locally, and CI is green on all platforms after rebasing onto `main` past #813, #817 and #819 (the all-sparse `map!` disambiguation now carries #813's `_unaliasargs`).

https://claude.ai/code/session_01MFPoSXGFJAGgJrHT98z2AY

### DIFF
--- a/src/abstractsparse.jl
+++ b/src/abstractsparse.jl
@@ -81,6 +81,8 @@ const SparseMatrixCSCSymmHerm{Tv,Ti} = HermOrSym{Tv,<:SparseMatrixCSCOrView{Tv,T
 
 # LinearAlgebra's BiTriSym (Bidiagonal/Tridiagonal/SymTridiagonal) and BandedMatrix (those
 # plus Diagonal) are imported for the banded special matrices.
+# LinearAlgebra's Q types that multiply sparse arrays via densification
+const LinAlgLeftQs = Union{HessenbergQ,QRCompactWYQ,QRPackedQ}
 
 # Former names of the above, not used here but relied on by downstream packages together
 # with SparseMatrixCSCView, SparseColumnView and SparseVectorView. May be deprecated in a
--- a/src/abstractsparse.jl
+++ b/src/abstractsparse.jl
@@ -79,16 +79,14 @@ const SparseTriangular{Tv,Ti} = UpperOrLowerTriangular{Tv,<:SparseMatrixCSCOrVie
 const SparseOrTri{Tv,Ti} = Union{SparseMatrixCSCOrView{Tv,Ti}, SparseTriangular{Tv,Ti}}
 const SparseMatrixCSCSymmHerm{Tv,Ti} = HermOrSym{Tv,<:SparseMatrixCSCOrView{Tv,Ti}}
 
-# LinearAlgebra's banded special matrices; Diagonal is often handled separately
-const BiTriSym = Union{Bidiagonal,Tridiagonal,SymTridiagonal}
-const DiagBiTriSym = Union{Diagonal,BiTriSym}
+# LinearAlgebra's BiTriSym (Bidiagonal/Tridiagonal/SymTridiagonal) and BandedMatrix (those
+# plus Diagonal) are imported for the banded special matrices.
 
 # Former names of the above, not used here but relied on by downstream packages together
 # with SparseMatrixCSCView, SparseColumnView and SparseVectorView. May be deprecated in a
 # future release.
 const SparseMatrixCSCUnion{Tv,Ti} = SparseMatrixCSCOrView{Tv,Ti}
 const SparseVectorUnion{Tv,Ti} = SparseVectorOrView{Tv,Ti}
-const AdjOrTransSparseVectorUnion{Tv,Ti} = AdjOrTransSparseVectorOrView{Tv,Ti}
 
 """
     issparse(S)
--- a/src/abstractsparse.jl
+++ b/src/abstractsparse.jl
@@ -40,6 +40,8 @@ Supertype for matrix with compressed sparse column (CSC).
 abstract type AbstractSparseMatrixCSC{Tv,Ti<:Integer} <: AbstractSparseMatrix{Tv,Ti} end
 
 # ---- Type aliases used for dispatch across files ----
+# Alias names use the Sparse family name and accept the abstract tier of the type they
+# are built on; an alias restricted to the concrete types says so in its name.
 # (Aliases for solver scalar types live with the solvers.)
 
 const AbstractSparseVecOrMat = Union{AbstractSparseVector,AbstractSparseMatrix}
@@ -70,8 +72,7 @@ const AdjOrTransSparseVectorOrView{Tv,Ti} = AdjOrTrans{Tv, <:SparseVectorOrView{
 const SparseVectorPartialView{Tv,Ti} = SubArray{Tv,1,<:AbstractSparseVector{Tv,Ti},<:Tuple{AbstractUnitRange},false}
 
 # `X` or an `Adjoint`/`Transpose` of `X`, named after LinearAlgebra's StridedMaybeAdjOrTransMat
-const AbstractSparseMatrixCSCMaybeAdjOrTrans = Union{AbstractSparseMatrixCSC, AdjOrTrans{<:Any,<:AbstractSparseMatrixCSC}}
-const AbstractSparseVecOrMatMaybeAdjOrTrans = Union{AbstractSparseVecOrMat, AdjOrTrans{<:Any,<:AbstractSparseVecOrMat}}
+const SparseMatrixCSCMaybeAdjOrTrans = Union{AbstractSparseMatrixCSC, AdjOrTrans{<:Any,<:AbstractSparseMatrixCSC}}
 const SparseVecOrMatMaybeAdjOrTrans = Union{SparseVecOrMat, AdjOrTrans{<:Any,<:SparseVecOrMat}}
 
 # LinearAlgebra wrappers around CSC storage
--- a/src/abstractsparse.jl
+++ b/src/abstractsparse.jl
@@ -41,6 +41,9 @@ abstract type AbstractSparseMatrixCSC{Tv,Ti<:Integer} <: AbstractSparseMatrix{Tv
 
 # ---- Type aliases used for dispatch across files ----
 # (Aliases for solver scalar types live with the solvers.)
+# SparseMatrixCSCView, SparseMatrixCSCUnion, SparseColumnView, SparseVectorView,
+# SparseVectorUnion and AdjOrTransSparseVectorUnion are not exported but are used by
+# downstream packages, so their names and definitions should stay stable.
 
 const AbstractSparseVecOrMat = Union{AbstractSparseVector,AbstractSparseMatrix}
 # types exposing compressed storage via nonzeros, rowvals/nonzeroinds and nzrange
--- a/src/abstractsparse.jl
+++ b/src/abstractsparse.jl
@@ -32,8 +32,6 @@ of type `Tv` and index type `Ti`. Alias for `AbstractSparseArray{Tv,Ti,2}`.
 """
 const AbstractSparseMatrix{Tv,Ti} = AbstractSparseArray{Tv,Ti,2}
 
-const AbstractSparseVecOrMat = Union{AbstractSparseVector,AbstractSparseMatrix}
-
 """
     AbstractSparseMatrixCSC{Tv,Ti<:Integer} <: AbstractSparseMatrix{Tv,Ti}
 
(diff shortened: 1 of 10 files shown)


src/linalg.jl
```diff
@@ -343,8 +341,13 @@ end
 # depending on expected execution speed the sorting of the result column is
 # done by a quicksort of the row indices or by a full scan of the dense result vector.
 # The last is faster, if more than ≈ 1/32 of the result column is nonzero.
-# TODO: extend to SparseMatrixCSCUnion to allow for SubArrays (view(X, :, r)).
+# TODO: extend to SparseMatrixCSCOrView to allow for SubArrays (view(X, :, r)).
+# Unit triangular wrappers keep their diagonal implicitly, so they are materialized first.
+_explicitdiag(A) = A
+_explicitdiag(A::UnitUpperOrUnitLowerTriangular{<:Any,<:SparseMatrixCSCOrView}) = sparse(A)
```
ViralBShah: [P2] Diagonal materialization can overflow the index type

```julia
A = spzeros(Float64, Int8, 127, 127)
UnitUpperTriangular(A) * A
```

The base returns a zero matrix; this PR throws `InexactError: trunc(Int8, 128)`. Materializing the implicit diagonal requires a final column pointer of 128, although the operands and result fit in Int8 storage. `UnitLowerTriangular` and sparse-vector multiplication are also affected (for example, `UnitUpperTriangular(A) * spzeros(127)`). Preserve implicit-diagonal handling or use a sufficiently wide temporary index type.

src/abstractsparse.jl
```diff
@@ -32,15 +32,63 @@ of type `Tv` and index type `Ti`. Alias for `AbstractSparseArray{Tv,Ti,2}`.
...
+# view of a unit range of a sparse vector's indices
+const SparseVectorPartialView{Tv,Ti} = SubArray{Tv,1,<:AbstractSparseVector{Tv,Ti},<:Tuple{AbstractUnitRange},false}
+
+# `X` or an `Adjoint`/`Transpose` of `X`, named after LinearAlgebra's StridedMaybeAdjOrTransMat
+const AbstractSparseMatrixCSCMaybeAdjOrTrans = Union{AbstractSparseMatrixCSC, AdjOrTrans{<:Any,<:AbstractSparseMatrixCSC}}
+const AbstractSparseVecOrMatMaybeAdjOrTrans = Union{AbstractSparseVecOrMat, AdjOrTrans{<:Any,<:AbstractSparseVecOrMat}}
+const SparseVecOrMatMaybeAdjOrTrans = Union{SparseVecOrMat, AdjOrTrans{<:Any,<:SparseVecOrMat}}
+
+# LinearAlgebra wrappers around CSC storage
+const SparseTriangular{Tv,Ti} = UpperOrLowerTriangular{Tv,<:SparseMatrixCSCOrView{Tv,Ti}}
```
ViralBShah: [P2] The expanded alias breaks custom sparse-vector multiplication

Including unit-triangular wrappers in `SparseOrTri` redirects `A * v`, for every `AbstractSparseVector`, through `spmatmulv`. But `spmatmul` only accepts compressed vectors and supported views. A custom `AbstractSparseVector` implementing the sparse accessors succeeds on the base and now throws `MethodError: spmatmul(::UnitUpperTriangular, ::V)`:

```julia
struct V <: AbstractSparseVector{Float64,Int}
    x::SparseVector{Float64,Int}
end
Base.size(v::V) = size(v.x)
Base.getindex(v::V, i::Int) = v.x[i]
SparseArrays.nonzeros(v::V) = nonzeros(v.x)
SparseArrays.nonzeroinds(v::V) = SparseArrays.nonzeroinds(v.x)
A = UnitUpperTriangular(sparse([1. 2.; 0. 1.]))
A * V(sparsevec([0., 3.]))  # base: [6.0, 3.0]; PR: MethodError
```

Keep a fallback for vectors outside the kernel's supported types.

ViralBShah (commented):
Found two P2 regressions in the expanded unit-triangular multiplication path, reproduced against base revision 6096fc0.

Validation: the linalg, higherorderfns, and fixed suites passed 9,033 tests, with one marked broken. Additional alias and custom-CSC mapping checks passed.

src/linalg.jl
ViralBShah: Fixed in 4b46016. `_explicitdiag` now widens the temporary to `Int` indices when `nnz + n` would not fit the parent's index type, and `spmatmul` takes the result index type from the original operands, so `UnitUpperTriangular(A) * A` with `Int8` indices returns an `Int8` `SparseMatrixCSC` again (and `* spzeros(127)` a `SparseVector`). Regression test added to the triangular-multiply testset. Note that `UnitUpper * UnitUpper` at `Int8` capacity still errors, as it does on `main` (`ArgumentError` there): the product's full unit diagonal genuinely does not fit.

src/abstractsparse.jl
ViralBShah: Fixed in 4b46016. The sparse-vector fast path now dispatches only on `SparseVectorOrView` (compressed vectors and whole column/vector views), split so it does not overlap the plain CSC × `AbstractSparseVector` method in `sparsevector.jl`. Other `AbstractSparseVector`s take the generic product, which also fixes `UpperTriangular(S) * v` for such vectors (a `MethodError` on `main`). Your repro is now a regression test.

src/linalg.jl
```diff
@@ -343,8 +341,13 @@ end
 # depending on expected execution speed the sorting of the result column is
 # done by a quicksort of the row indices or by a full scan of the dense result vector.
 # The last is faster, if more than ≈ 1/32 of the result column is nonzero.
-# TODO: extend to SparseMatrixCSCUnion to allow for SubArrays (view(X, :, r)).
+# TODO: extend to SparseMatrixCSCOrView to allow for SubArrays (view(X, :, r)).
```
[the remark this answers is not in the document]
ViralBShah: Removed in 4b46016.

PR #810, opened 2026-09-10
Update version number and julia_compat to 1.14

ViralBShah:

PR #817, opened 2026-09-11
Revert the SPQR adjoint changes that landed with #810

ViralBShah:
#810 was meant to only bump the version and julia compat to 1.14, but it was accidentally based on the branch of #804 and squash-merged with the "Solve with the adjoint of a sparse QR factorization" change included (see https://github.com/JuliaSparse/SparseArrays.jl/pull/810#issuecomment-5633278266). This reverts the `src/solvers/spqr.jl` and `test/spqr.jl` portion of a248d20 so that #804 can be reviewed and merged on its own. The `Project.toml` bump is kept.

Both files are restored byte-for-byte to their state before a248d20, and `test/spqr.jl` passes locally on the 1.14-DEV build.

https://claude.ai/code/session_01RFzXKGsJuSogARQpDGipPr

### DIFF
--- a/src/solvers/spqr.jl
+++ b/src/solvers/spqr.jl
@@ -5,7 +5,7 @@ module SPQR
 import Base: \, *
 using Base: require_one_based_indexing
 using LinearAlgebra
-using LinearAlgebra: AbstractQ, AdjointQ, AdjointAbsVec, AdjointFactorization, copy_similar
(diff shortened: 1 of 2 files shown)


ViralBShah:
Superseded by #818, which reverts the whole commit.

ViralBShah:
Reopened: this is the intended fix. It reverts only the SPQR adjoint change from a248d20 and keeps the Project.toml bump.

PR #825, opened 2026-09-16
Multiply sparse operands with a sparse QR's Q through its reflectors

### LINKED ISSUE #121: Sparse QR: thin Q*R fails with sparse R; sparse*Q slow and missing adjoint methods
https://github.com/JuliaLang/julia/blob/b56a9f07948255dfbe804eef25bdbada06ec2a57/stdlib/SuiteSparse/src/spqr.jl#L207
The code should be 
```julia
     size(A, 1) !=size(Q.factors,2)
```
as in https://github.com/JuliaLang/julia/blob/80516ca20297a67b996caa08c38786332379b6a5/stdlib/LinearAlgebra/src/qr.jl#L533

ViralBShah:
Fixes #121.

Products of a sparse QR's `Q` with a **sparse** operand were either slow or threw:

- `F.Q * B` and `A * F.Q` with sparse `B` or `A` went through `sparse(Q)`, which materializes the `m x m` matrix column by column even though `Q` is a product of Householder reflectors and dense in general. For a 4000 x 50 factorization that is 1.1 s against 8 ms for a dense operand.
- `F.Q' * B`, `A' * F.Q`, `transpose(A) * F.Q` and `A * F.Q'` with a sparse operand threw a `MethodError`. They fell through to LinearAlgebra's generic product, which allocates the result with `similar` on the operand, gets a sparse array back, and calls `lmul!` or `rmul!` on it, for which only strided methods exist.
- The thin `F.Q * F.R` from the issue failed on the size check of the `sparse(Q)` method for any tall `A`, while `F.Q * Array(F.R)` worked. The only existing test of that identity used a square factorization.

### What changes

A sparse operand, including the `Adjoint` and `Transpose` of one, column-range views, sparse vectors, their views and a transposed sparse vector, is copied to a dense array and multiplied by the existing dense-operand methods. So the thin shapes those accept (`Q * B` with `size(A, 2)` rows, `A * Q'` with `size(A, 2)` columns, matching dense QR) are accepted here too, the adjoint of `Q` works, and nothing materializes `Q`. The result is a dense `Matrix` or `Vector`, as for a dense operand; it is dense in general, and callers who want the `Q * R` identity as a sparse matrix can wrap it in `sparse`.

Two tests asserted a `SparseMatrixCSC` result from `sparse(I, 2, 2) * Q` (#26367) and `F.Q * F.R` (#26368). They now assert the dense type and still check the values.

### Measured

4000 x 50 factorization, 1.14-DEV:

| product | before | after | dense operand |
|---|---|---|---|
| `C * Q`, sparse `C` 60 x 4000 | 1143 ms | 7.6 ms | 7.5 ms |
| `Q * B`, sparse `B` 4000 x 50 | 1169 ms | 14.9 ms | 12.9 ms |
| `C' * Q`, `Q' * B`, `B' * Q`, `transpose(B) * Q`, `C * Q'` | `MethodError` | work | |
| `F.Q * F.R`, 27 x 2 | `DimensionMismatch` | equals `A[F.prow, F.pcol]` | |

This touches dispatch: six new `*` methods on the aliases from `abstractsparse.jl`, and the two old `sparse(Q)` methods are gone. Since `Q * SparseMatrixCSC` now returns a dense matrix instead of a sparse one, it is a behaviour change and not a backport candidate. The `qr` docstring says nothing about product result types, so no docs change.

### Tests

A new testset over tall, square, wide, complex and `Int32`-indexed factorizations checks the `Q * R` identity, every sparse-operand product (both orders, adjoint and transpose operands, adjoint `Q`, views, sparse vectors, thin shapes) against the same product with a dense copy of the operand, the dense result type, and `DimensionMismatch` for the shapes the dense-operand methods reject.

Tested on 1.14-DEV with `test/spqr.jl`, `test/ambiguous.jl` and the whitespace check.

The local `_densecopy` helper exists because `Matrix` of a sparse adjoint or view goes through generic indexing; #828 fixes those constructors, and a follow-up can replace the helper with plain `Matrix` and `Vector` calls once it merges.

Not addressed here: a real `QRSparseQ` cannot multiply a complex operand of either kind, because `convert(AbstractQ{ComplexF64}, ::QRSparseQ{Float64})` is not defined. That is pre-existing and separate.


ViralBShah:
Closing to move this off the stack; GitHub does not allow changing the base of a stacked PR. Reopened against main with the same branch.

PR #828, opened 2026-09-17
Convert views, adjoints and transposes of sparse arrays to dense in O(length + nnz)

ViralBShah:
### Issue

`Matrix(A')`, `Matrix(transpose(A))`, `Matrix(view(A, :, r))`, `Vector(view(A, :, j))`, `Vector(view(x, r))` and `Matrix(x')` for sparse `A` and `x` all fell through to Base's generic element-wise `copyto!`, which indexes the sparse parent once per element. Only a plain `SparseMatrixCSC` or `SparseVector` had an O(nnz) path, and the vector one was a hand-written `Vector` constructor in `sparsevector.jl`.

```julia
A = sprandn(4000, 4000, 1e-4)
Matrix(A')   # 22 ms on main, 2.3 ms here
```

### Mechanism and fix

The `Array` constructors copy through `copyto!`, so that is the one hook. `sparseconvert.jl` now holds all sparse-to-dense copies in one section: the existing matrix kernel, widened to column-range views through `getrowval` and `getnzval`, plus kernels for sparse vectors and their views, adjoints and transposes of sparse matrices, and adjoints and transposes of sparse vectors. Each zeroes the destination only when the source has structural zeros, and takes the zero from the parent eltype through the wrapper op, so element types without a `zero` still convert when every entry is stored, and signed zeros match what the element-wise path produced.

The hand-written `Matrix`, `Vector` and `Array` constructors for sparse matrices and vectors are removed; the Base constructors reach the same kernels. The buffer check the `Matrix` constructor ran now sits in the two kernels that read the buffers under `@inbounds`, so every entry point runs it. This touches dispatch: `copyto!` gains methods for the wrapped sources with `AbstractMatrix` destinations, and `PermutedDimsArray` resolutions as the existing sparse method has. Vector sources are specialized on `Vector` destinations only, to stay clear of the sparse-destination `copyto!` methods.

### Measured

1.14-DEV, minimum of five runs:

| conversion | main | this PR |
|---|---|---|
| `Matrix(A')`, 4000 x 4000, 1600 stored | 22.2 ms | 2.3 ms |
| `Matrix(transpose(A))` | 18.5 ms | 2.4 ms |
| `Matrix(view(A, :, 1:2000))` | 8.7 ms | 1.1 ms |
| `Vector(view(x, :))`, 10^6, 100 stored | 6.9 ms | 0.06 ms |
| `Vector(view(x, 1:10^5))` | 0.73 ms | 0.04 ms |
| `Matrix(x')` | 6.9 ms | 0.13 ms |
| `Matrix(A)`, `Vector(x)` | unchanged | unchanged |

### Tests

In the existing conversion testsets: each wrapped form compared with `collect`, the result type, and `which` proving the `copyto!` method is this package's. Run on 1.14-DEV with the full suite, `test/ambiguous.jl` and the whitespace check.

Performance only, no behaviour change. PR #830 (sparse QR products) is independent and can drop its local dense-copy helper once this merges.


dkarrasch:
Uh, I started working on the same, but my coding by hand is too slow for the "AI competitor". Anyway, this looks very much like my attempt.

ViralBShah:
@dkarrasch Happy to see if @keno can get you an open source license (should you be interested).

I just saw a lot of noise in #799 with code duplication, and hence did this. I am glad to see your comment here and would love to know  if you have any comments - otherwise I was planning to merge it. Also happy to use your PR instead if it is almost ready.

dkarrasch:
Does this mean we can delete https://github.com/JuliaSparse/SparseArrays.jl/blob/ce0013b7a6cc8d36b9c1bcab483684f9a5d91f5d/src/sparsematrix.jl#L985-L992

or should the buffer check be somehow included here?

dkarrasch:
> @dkarrasch Happy to see if @Keno can get you an open source license (should you be interested).

I already got one, thanks, I just didn't get around to setting up an agent and all of that.

ViralBShah:
Yes, deleted in b5c1c18: Base's `Matrix(S)` allocates and calls the same `copyto!`, so the constructors only added the buffer check. That check now sits at the top of the two kernels that read the buffers under `@inbounds` (the shared one in `sparseconvert.jl` and the fast `Array` path in `sparsematrix.jl`), so every entry point runs it, including `Matrix{T}(S)`, `copyto!(dense, S)` and the wrapped forms, which skipped it before. Full suite passes on 1.14-DEV.

PR #830, opened 2026-09-17
Multiply sparse operands with a sparse QR's Q through its reflectors

### LINKED ISSUE #121: Sparse QR: thin Q*R fails with sparse R; sparse*Q slow and missing adjoint methods

ViralBShah:
Fixes #121.

Products of a sparse QR's `Q` with a **sparse** operand were either slow or threw:

- `F.Q * B` and `A * F.Q` with sparse `B` or `A` went through `sparse(Q)`, which materializes the `m x m` matrix column by column even though `Q` is a product of Householder reflectors and dense in general. For a 4000 x 50 factorization that is 1.1 s against 8 ms for a dense operand.
- `F.Q' * B`, `A' * F.Q`, `transpose(A) * F.Q` and `A * F.Q'` with a sparse operand threw a `MethodError`. They fell through to LinearAlgebra's generic product, which allocates the result with `similar` on the operand, gets a sparse array back, and calls `lmul!` or `rmul!` on it, for which only strided methods exist.
- The thin `F.Q * F.R` from the issue failed on the size check of the `sparse(Q)` method for any tall `A`, while `F.Q * Array(F.R)` worked. The only existing test of that identity used a square factorization.

### What changes

A sparse operand, that is a sparse matrix or a column-range view of one, with or without an `Adjoint` or `Transpose`, a transposed sparse vector as a one-row matrix, or a sparse vector or a view of one, is copied to a dense array and multiplied by the existing dense-operand methods. So the thin shapes those accept (`Q * B` with `size(A, 2)` rows, `A * Q'` with `size(A, 2)` columns, matching dense QR) are accepted here too, the adjoint of `Q` works, and nothing materializes `Q`. The result is a dense `Matrix` or `Vector`, as for a dense operand; it is dense in general, and callers who want the `Q * R` identity as a sparse matrix can wrap it in `sparse`.

Two tests asserted a `SparseMatrixCSC` result from `sparse(I, 2, 2) * Q` (#26367) and `F.Q * F.R` (#26368). They now assert the dense type and still check the values.

### Measured

4000 x 50 factorization, 1.14-DEV:

| product | before | after | dense operand |
|---|---|---|---|
| `C * Q`, sparse `C` 60 x 4000 | 1143 ms | 7.6 ms | 7.5 ms |
| `Q * B`, sparse `B` 4000 x 50 | 1169 ms | 14.9 ms | 12.9 ms |
| `C' * Q`, `Q' * B`, `B' * Q`, `transpose(B) * Q`, `C * Q'` | `MethodError` | work | |
| `F.Q * F.R`, 27 x 2 | `DimensionMismatch` | equals `A[F.prow, F.pcol]` | |

### LinearAlgebra's Q types

As @dkarrasch pointed out, the existing methods for `HessenbergQ`, `QRCompactWYQ`, `QRPackedQ` and `LQPackedQ` had the same shape of problem: an adjoint or transposed sparse operand went through `copy(B)`, which builds a sparse transpose and then dispatches a second time to the plain sparse method that calls `Matrix`. They now share the two operand unions with `QRSparseQ`, defined in `abstractsparse.jl`, and call `Matrix` or `Vector` directly. That also gives the dense Q types column-range views of sparse matrices, adjoints and transposes of those views, transposed sparse vectors, and views of sparse vectors, which threw a `MethodError` before. An adjoint sparse vector stays with LinearAlgebra's generic `(Q' * u')'`, so `b' * Q` keeps returning an `Adjoint` as for a dense `b`. The separate sparse-vector loop in `sparsevector.jl` is folded into the same loop. Result types are unchanged there. A testset in `test/sparsematrix_ops.jl` builds the four Q types (plain and pivoted QR, Hessenberg, LQ) from one matrix and checks the same operand kinds as the SPQR testset.

This touches dispatch: eight `*` methods per Q type on the two operand unions from `abstractsparse.jl`, and the two old `sparse(Q)` methods are gone. Since `Q * SparseMatrixCSC` now returns a dense matrix instead of a sparse one, it is a behaviour change and not a backport candidate. The `qr` docstring says nothing about product result types, so no docs change.

### Tests

A new testset over a tall real and a wide complex factorization checks the `Q * R` identity, one sparse operand of each kind (both orders, adjoint and transpose operands, adjoint `Q`, views, sparse vectors, thin shapes) against the same product with a dense copy of the operand, the dense result type, and `DimensionMismatch` for the shapes the dense-operand methods reject.

Tested on 1.14-DEV with `test/spqr.jl`, `test/sparsematrix_ops.jl`, `test/sparsevector.jl`, `test/ambiguous.jl` and the whitespace check.

The dense copies are plain `Matrix` and `Vector` calls, which are O(length + nnz) for sparse wrappers and views since #828. Rebased on `main` after that merge.

Not addressed here: a real `QRSparseQ` cannot multiply a complex operand of either kind, because `convert(AbstractQ{ComplexF64}, ::QRSparseQ{Float64})` is not defined. That is pre-existing and separate.


dkarrasch:
The same direct densification approach should (perhaps afterward?) be applied to https://github.com/JuliaSparse/SparseArrays.jl/blob/be805914ebde95481b2e35039646202d0b788529/src/sparsematrix.jl#L480-L490.

ViralBShah:
Done in 024a6dc. The dense Q types now share the operand unions with `QRSparseQ` and call `Matrix` directly instead of `copy`, which also gives them column-range views of sparse matrices, adjoints and transposes of those views, and sparse vector views. The sparse-vector loop in `sparsevector.jl` is folded into the same loop, and there is a testset over the four Q types in `test/sparsematrix_ops.jl`.

src/abstractsparse.jl
```diff
@@ -86,9 +86,9 @@ const SparseMatrixCSCSymmHerm{Tv,Ti} = HermOrSym{Tv,<:SparseMatrixCSCOrView{Tv,T
 const LinAlgLeftQs = Union{HessenbergQ,QRCompactWYQ,QRPackedQ}
 # Sparse operands that a Q multiplies through a dense copy: Q is dense in general, so
 # applying it to a dense copy of the operand is cheaper than materializing Q or a sparse
-# transpose of the operand. Matrix operands, including adjoints, transposes and column-range
-# views, and a transposed sparse vector as a one-row matrix; and vector operands.
-const SparseQMatOperand = Union{SparseMatrixCSCMaybeAdjOrTrans, SparseMatrixCSCView, Transpose{<:Any,<:SparseVectorOrView}}
+# transpose of the operand. Sparse matrices and their views, with or without an adjoint or
+# transpose; and sparse vectors and their views.
+const SparseQMatOperand = Union{SparseMatrixCSCOrView, AdjOrTrans{<:Any,<:SparseMatrixCSCOrView}}
```
dkarrasch: Should we include here `AdjOrTrans{<:Any,<:SparseVectorOrView}`? Depends if we have nice corresponding `Matrix` constructor for that case.

ViralBShah: I was trying to see if I could avoid adding new type unions - but it should be possible to do that.

ViralBShah: Added `Transpose{<:Any,<:SparseVectorOrView}` in 05defd5; `Matrix` of it is a fast path since #828. `Adjoint` is left out on purpose: `b' * Q` already works through LinearAlgebra's generic `(Q'u')'` and returns an `Adjoint` as for a dense `b`, and a more specific method here would turn that into a one-row `Matrix`.

PR #836, opened 2026-09-17
Build the CSC arrays directly when stacking sparse vectors

### LINKED ISSUE #498: Base.stack is underperforming for SparseArrays
**Describe the bug**
Stacking arrays of SparseArrays is slow.
Similar to JuliaGPU/CUDA.jl#2248.

**To reproduce**

The Minimal Working Example (MWE) for this bug:

```julia
using BenchmarkTools, SparseArrays;
N=100;
M=1000;
x=spzeros(N);
x[1]=10.0;
x_dense=collect(x);
@btime stack(fill($x,$M));
@btime stack(fill($x_dense,$M));
@btime sparse(stack(fill(collect($x),$M)));
```
As timing I am getting:
```
2.211 ms (15 allocations: 75.20 KiB)
74.900 μs (3 allocations: 789.23 KiB)
194.600 μs (7 allocations: 813.98 KiB)
```

**Version info**

Details on Julia: 1.10

```
Julia Version 1.10.0
Commit 3120989f39 (2023-12-25 18:01 UTC)
Build Info:
  Official https://julialang.org/ release
Platform Info:
  OS: Windows (x86_64-w64-mingw32)
  CPU: 8 × Intel(R) Core(TM) i7-8550U CPU @ 1.80GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-15.0.7 (ORCJIT, skylake)
  Threads: 1 on 8 virtual cores
```

ViralBShah:
Fixes #498.

```julia
using SparseArrays, BenchmarkTools
x = spzeros(100); x[1] = 10.0
@btime stack(fill($x, 1000))            # before: 634 μs, after: 7 μs
@btime stack(fill($(sprand(10_000, 0.001)), 1000))  # before: 75 ms, after: 8 μs
```

`stack` on a collection of sparse vectors had no sparse path, so Base's generic `_typed_stack` loop ran: `similar` gave an empty `SparseMatrixCSC`, and the fallback `copyto!(B, offset, x)` then copied every element of every slice, zeros included, through `getindex` on the vector and `setindex!` on the matrix. That is O(m·n) element operations with a binary search each, and it grows with the vector length regardless of sparsity, which is why the dense round trip was faster. This adds a `Base._typed_stack` method for sparse vector slices that assembles `colptr`, `rowval` and `nzval` in one pass. The slices may be sparse vectors or views of them and of sparse matrix columns (`SparseVectorOrView`), so `stack(eachcol(S))` is covered (16 ms to 3 μs for a 10,000×200 matrix). It applies to `stack(A)`, `stack(A; dims=2)` and `stack(f, A)`. A `Base._dim_stack` method handles `dims=1` as `permutedims` of that result (15 ms to 13 μs for 200 vectors of length 10,000) and now returns a `SparseMatrixCSC`. A container with more than one axis stacked without `dims` still falls through to Base and produces a dense 3-d array, since there is no sparse type for it; the same holds for `stack` of `SparseMatrixCSC` slices.

`hcat` of sparse vectors now converts its arguments to `SparseVector{Tv,Ti}` and calls `stack` on them, so the near-identical `_absspvec_hcat` is removed and no new helper is added. `hcat` timings and allocations are at parity with `main` (2 to 64 vectors, homogeneous and mixed `Tv`/`Ti`).

This touches dispatch: it adds methods to the Base internals `_typed_stack` and `_dim_stack`.

Tests cover the reproducer against `hcat`, `dims=1` and an out-of-range `dims`, a generator, `eachcol` views, promotion across element and index types, the dense multi-axis fallback, and the empty and mismatched-length errors. `test/sparsevector.jl` passes on nightly.


src/sparsevector.jl
```diff
@@ -1224,6 +1224,42 @@ function _absspvec_hcat(X::AbstractSparseVector{Tv,Ti}...) where {Tv,Ti}
...
+# `copyto!(B, offset, x)` per slice, which copies every element, including the stored
+# zeros, through `getindex` and `setindex!` on sparse arrays. Building the CSC arrays
+# directly, as `_absspvec_hcat` does, costs O(nnz + n) instead of O(m * n).
+function Base._typed_stack(::Colon, ::Type{T}, ::Type{S}, A, Aax::Tuple{Any}) where {T,S<:AbstractSparseVector}
+    X = A isa AbstractArray ? A : collect(A)
+    isempty(X) && return Base._empty_stack(:, T, S, A)
+    Ti = mapreduce(indtype, promote_type, X)
+    return _absspvec_stack(T, Ti, X)
+end
+function _absspvec_stack(::Type{Tv}, ::Type{Ti}, X) where {Tv,Ti}
```
dkarrasch, line 1207: Why do we need the extra function here? As a function barrier? Or could this be reused elsewhere?

ViralBShah: It wants it for the function barrier, but I am able to get more reuse as well (as in the next commit).

PR #837, opened 2026-09-18
Show how to import external CSC and CSR arrays in the manual

ViralBShah:
The CSC storage section of the manual described importing CSC data from other libraries and the double-transpose trick for sorting row indices, but showed no code. This replaces that paragraph with a short passage covering the following.
- `SparseMatrixCSC(m, n, colptr, rowval, nzval)` aliases its arguments.
- The storage invariants, and which of them the constructor checks: `sparse_check` and `_goodbuffers` validate `colptr` and the buffer lengths, but never look at the row indices.
- Converting 0-based arrays.
- Sorting row indices with `copy(transpose(copy(transpose(A))))` or `sparse(findnz(A)..., size(A)...)`.
- Importing CSR arrays as the CSC arrays of the transpose.

The material comes from reading `src/sparsematrix.jl` and running each case. The examples are doctests, checked with the `docs/make.jl` build on Julia nightly. Documentation only.


dkarrasch:
This introduced an error in the docs generation. The new docstring doesn't define `A`. Maybe this `cscimport` is supposed to import an `A` that is defined elsewhere?

ViralBShah:
I am fixing in a minute. Merged a couple of doc PRs without the CI run. Learnt my lesson!

PR #840, opened 2026-09-18
Add a performance tips section to the sparse arrays manual

ViralBShah:
Adds a "Performance tips" section to `docs/src/index.md`, before the dense and sparse correspondence table. It covers:
- column-major iteration over stored entries with `nzrange`, `rowvals` and `nonzeros`
- building from `(I, J, V)` with `sparse` instead of repeated `setindex!` (and what `sizehint!` does and does not save)
- column versus row slicing
- `Int32` indices and their size limits
- which operations work through the lazy `transpose` and `adjoint` wrappers and which copy or fall back to generic methods
- stored zeros and `dropzeros!`, `droptol!` and `fkeep!`

The material comes from reading `src/sparsematrix.jl`, `src/sparsevector.jl`, `src/linalg.jl` and `src/higherorderfns.jl`, with each dispatch claim checked using `@which` on nightly. All examples are doctests, and `docs/make.jl` passes on `julia +nightly`. Documentation only, no dispatch changes.

Adjacent to the insertion point of #839, so whichever merges second will need a trivial rebase.


docs/src/index.md
```diff
@@ -186,6 +186,150 @@ In many cases it may be better to convert the sparse matrix into `(I,J,V)` forma
...
+           return y
+       end;
+
+julia> mymul(A, [1.0, 10.0, 100.0]) == A * [1.0, 10.0, 100.0]
+true
+```
+
+In contrast, scalar indexing `A[i, j]` has to search column `j` for row `i`, and a loop over all
+`(i, j)` costs time proportional to `length(A)` rather than `nnz(A)`.
+
```
ViralBShah, line 220: It seems this will be proportional to `prod(size(A))` rather than `length(A)`.

PR #857, opened 2026-09-19
Lower indices with `to_indices` in sparse `getindex`

### LINKED ISSUE #42: Sparse array index lowering
When implementing custom array types it's useful to independently resolve indices using `to_indices`, and subsequently index into an underlying array with them. Right now this works with `Array`, but fails when the array is sparse:

```julia
using SparseArrays
let x = sparse([1 2; 3 4])
    x[1, [true, false]] # Works
    
    I′ = to_indices(x, (1, [true, false]))
    x[I′...] # => getindex not defined for Base.LogicalIndex{Int64,Array{Bool,1}}
end
```

This also comes up when indexing into e.g a custom array with user-defined index types, or with `Not` from [InvertedIndices.jl](https://github.com/mbauman/InvertedIndices.jl) (this originally came up in an [issue](https://github.com/mbauman/InvertedIndices.jl/issues/9) there.)

I think it has to do with the way that sparse arrays hook into the `to_indices` machinery.

ViralBShah:
Fixes #42. Sparse `getindex` is defined on raw index types and never lowers them, unlike `setindex!`, which calls `Base.ensure_indexable(to_indices(...))`. Four symptoms, plus a fallback on adjoints:

```julia
using SparseArrays
x = sparse([1 2; 3 4; 5 6])
x[to_indices(x, (1, [true, false]))...]  # CanonicalIndexError: getindex not defined for Base.LogicalIndex
x[[true, false], 1]                      # wrong-length mask accepted; dense throws BoundsError
m = x .> 2
x[vec(m)] == Array(x)[vec(m)]            # false: wrong values
x[Not(1), Not(2)]                        # InvertedIndices: Base's elementwise fallback, O(m*n)
```

Mechanism and fix, one commit each:

1. `_logical_index` reads `I[row, col]` under `@inbounds`, an out-of-bounds read for a `Vector{Bool}` mask (which `checkbounds` allows when `length(I) == length(A)`). The mask is reshaped to `size(A)`. Wrong-results bug, candidate for backport on its own.
2. `to_indices` lowers a mask to `Base.LogicalIndex`, an `AbstractVector` that iterates but has no `getindex`, so it matched the generic kernels and failed on random access.
3. The generic kernels `(Integer, Vector)`, `(Range, Vector)`, `(Vector, Vector)` now lower through `_lower_indices`: `to_indices`, then `checkbounds` while a mask still knows its length, then `ensure_indexable`. That replaces the `findall` block of six `Bool` methods, which skipped the mask length check, and the hand-written `extrema` bounds checks. A range of `Bool` is treated as a mask, as dense does, instead of erroring. A catch-all `getindex(A, I...)` for `AbstractSparseMatrixCSC` and `AbstractSparseVector` lowers anything no method matches (custom index types, `CartesianIndex`) and dispatches again; indices that lowering leaves unchanged (`A[i]`, `A[i, j, 1]`) go to Base via `invoke`, which is the only non-internal way to reach Base's generic method from a more specific one.

4. `Adjoint`/`Transpose` of a sparse matrix had only `(i, :)` methods; every other nonscalar index (`A'[1:2, 1:2]`, `A'[:, j]`, masks, custom indices) took Base's elementwise fallback. A two-index method now lowers the indices and indexes the parent with them swapped, replacing the two `Colon` methods. It needs one `(i::Int, j::Int)` method to resolve the ambiguity with LinearAlgebra's scalar `getindex`, with the same body. Linear indexing of an adjoint stays with Base.

**Touches dispatch**: removes ten `getindex` methods and adds two varargs catch-alls, one `AbstractRange{Bool}` method, and two methods on `AdjOrTrans` of a sparse matrix. The Aqua ambiguity job passes locally.

Custom index on 20000×20000 with 40k stored entries, `A[AllBut(1), AllBut(2)]`, and scalar/nonscalar paths on 1000×1000 at 1% density (min of 5, nightly, M-series):

| | before | after |
|---|---|---|
| custom index pair | 2370 ms | 5.6 ms |
| `A[i]` loop over all | 5.03 ms | 5.02 ms |
| `A[CartesianIndex]` loop | 3.52 ms | 3.53 ms |
| `A[i,j]` loop | 3.58 ms | 3.57 ms |
| `A[r,r]`, 300 random | 0.039 ms, 41072 B | 0.038 ms, 41072 B |
| `A[m,m]` masks | 0.050 ms, 61840 B | 0.051 ms, 61840 B |
| `A[2:900,r]` | 0.005 ms, 51952 B | 0.005 ms, 51952 B |

`T = A'`, 5000×5000 `ComplexF64` at 0.1% density (min of 3):

| | before | after |
|---|---|---|
| `T[1:2500,1:2500]` | 1265 ms, 619 MB | 0.083 ms, 0.4 MB |
| `T[r,r]`, 1000 random | 104 ms, 78 MB | 0.17 ms, 0.17 MB |
| `T[m,m]` masks | 1280 ms, 377 MB | 0.34 ms, 0.43 MB |
| `T[:,7]` | 0.127 ms, 332 kB | 0.005 ms, 752 B |
| `T[i,j]` loop, `T[7,:]`, `T[r,:]` | unchanged | unchanged |

Left out: linear (single-index) nonscalar indexing of an adjoint or transpose still takes Base's fallback, and InvertedIndices itself was not tested (no new test dependency); the test uses an equivalent index type lowered through `to_indices`.

Tested on nightly 1.14.0-DEV.3298: full suite passes (28135 pass, 9 broken as on main), Aqua/ambiguity job and whitespace check pass. Tests cover real and complex, matrix and vector, pre-lowered masks, wrong-length masks, and assert custom indices dispatch into SparseArrays.


dkarrasch:
The OP says something about backporting some of this?

ViralBShah:
This fixes lots of issues with getindex (including an out of bounds) - but I think it is best not to backport unless someone needs it. I would much rather let it simmer on master and make its way into 1.14.

Should be easy to isolate the exact case and build a backportable PR. Let me know if you think otherwise.

Issue #627, opened 2025-05-19
Missing dot product specializations for `SparseArray`s with `Adjoint{<:SparseArray}`s


MasonProtter:
Here's an MWE where there's a 100x performance drop due to hitting the generic `dot` fallback when taking the inner product of a sparse array with the adjoint of a sparse array:
```julia
julia> using SparseArrays, LinearAlgebra, BenchmarkTools

julia> let N = 1000
           A = sprand(N, N, 0.1)
           At = spzeros(N, N)
           adjoint!(At, A)
           @btime $At ⋅ $A
           @btime $A' ⋅ $A
       end;
  509.345 μs (0 allocations: 0 bytes)
  32.945 ms (0 allocations: 0 bytes)
```

ViralBShah:
This is currently:

```
julia> VERSION
v"1.13.0"

julia> using SparseArrays, LinearAlgebra, BenchmarkTools

julia> let N = 1000
           A = sprand(N, N, 0.1)
           At = spzeros(N, N)
           adjoint!(At, A)
           @btime $At ⋅ $A
           @btime $A' ⋅ $A
       end;

  663.838 μs (0 allocations: 0 bytes)
  2.196 ms (0 allocations: 0 bytes)
```

ViralBShah:
#770 added an O(nnz) `dot(::AdjOrTrans{<:Any,<:AbstractSparseMatrixCSC}, ::AbstractSparseMatrixCSC)` kernel, so the case reported here no longer reaches the generic fallback. On `main` with nightly (1.14.0-DEV.3300), minimum of 20 runs:

| N, density | `At ⋅ A` | `A' ⋅ A` | `A ⋅ A'` |
|---|---|---|---|
| 1000, 0.1 | 0.67 ms | 1.05 ms | 1.04 ms |
| 10⁴, 1e-3 | 0.66 ms | 1.12 ms | 1.13 ms |
| 10⁵, 1e-4 | 6.8 ms | 13.2 ms | 13.3 ms |

The remaining 1.6-2x over the materialized adjoint is the cost of matching entries at transposed positions rather than merging two sorted columns. The 1.13.0 numbers above predate #770.

Mixed wrappers (`transpose(A) ⋅ C'`) still hit the generic method (37 ms at N=1000); a follow-up PR for that is coming.

*Written by Claude Code.*

PR #858, opened 2026-09-19
Specialize `dot` for mixed adjoint and transpose sparse matrices

ViralBShah:
Follow-up to #770 for #627. `dot` of an `Adjoint` with a `Transpose` of a sparse matrix (either order) still reaches LinearAlgebra's generic O(mn) `getindex` loop, because LinearAlgebra only strips matching wrappers.

```julia
A = sprand(ComplexF64, 10^4, 10^4, 1e-3); C = sprand(ComplexF64, 10^4, 10^4, 1e-3)
dot(A', transpose(C))
```

Both wrappers transpose positions, so the stored entries of the parents line up column by column and only the elementwise operation differs. The column-merge loop of `dot(::AbstractSparseMatrixCSC, ::AbstractSparseMatrixCSC)` moves into `_dot_walk(f, A, B)`, and the two new methods call it on the parents with `dot(adjoint(a), transpose(b))` and its mirror. This touches dispatch: two new `dot` methods, no new ambiguities with LinearAlgebra.

Minimum of 10 runs, nightly 1.14.0-DEV.3300, 0 allocations before and after:

| N, density | `A' ⋅ transpose(C)` main | PR | `A ⋅ C` |
|---|---|---|---|
| 1000, 0.1 | 36.6 ms | 0.66 ms | 0.66 ms |
| 10⁴, 1e-3 | 1469 ms | 0.65 ms | 0.65 ms |

Tested against dense with complex entries, stored zeros and a non-square shape, plus a multiplication count proving the kernel is reached; `test/linalg_products.jl` passes on nightly. Not covered: `Symmetric`/view operands against a lazy adjoint, which remain on the fallback.

Written by Claude Code.


dkarrasch:
Would we want something like that in LinearAlgebra as well? Like indexing into the parents and tweak the elementwise `dot` computation?

ViralBShah:
The combinatorial explosion of types is much easier to specialize with agents. In many cases the performance gains are very nice - so I would think it is worth trying.

In SparseArrays, generic fallbacks end up being way more expensive than in dense, where it is possible that the gap is not much. Best to measure.

PR #859, opened 2026-09-19
Implement the 2-norm in `opnorm` for sparse matrices

### LINKED ISSUE #119: Operator norm (p = 2) for sparse matrices is not implemented
Using `opnorm` on a sparse matrix (of type `SparseMatrixCSC`) gives this error:
```
ERROR: ArgumentError: 2-norm not yet implemented for sparse matrices. Try opnorm(Array(A)) or opnorm(A, p) where p=1 or Inf.
```

Here is a piece of code which produces that error:
```
using LinearAlgebra
using SparseArrays
A = SparseMatrixCSC([1.0 2.0 0.0; 0.0 1.0 0.0; 0.0 0.0 0.3])
opnorm(A)
```

I am interested in working on that issue, but I would need some guidance.

The proper way to solve this (in my opinion) is to implement `svdvals` for sparse matrices, so that `opnorm` can be modifed to return the largest singular value of `A` by using `svdvals` for sparce matrices. If this is the way to go, which algorithm should be used to implement `svdvals` for sparse matrices?
There have been many discussions about implementing SVD for sparse matrices (e.g. JuliaLang/LinearAlgebra.jl#106) a long time ago...

If the proposed solution is not the desired way to resolve this, why can't we just replace the error message by returning `opnorm(Array(A))` as it is suggested in the error message itself?

ViralBShah:
Fixes #119. `opnorm` defaults to `p = 2`, which threw for sparse matrices:

```julia
julia> opnorm(sparse([1.0 2.0 0.0; 0.0 1.0 0.0; 0.0 0.0 0.3]))
ERROR: ArgumentError: 2-norm not yet implemented for sparse matrices. Try opnorm(Array(A)) or opnorm(A, p) where p=1 or Inf.
```

It now returns an estimate of the largest singular value from Golub-Kahan-Lanczos bidiagonalization (internal helper `opnorm2est`), using only `A*v` and `A'*u`. It stops once the residual of the leading Ritz pair is below `1e-10` relative to it on two checks in a row, or after `max(100, 2*min(m, n))` steps. The estimate converges from below with error about `min(r, r^2/gap)`. Each check takes one eigenpair of the Golub-Kahan tridiagonal, at O(k). The recurrence runs on `A/s`, with `s` the largest stored magnitude, so its `Float64` coefficients stay in range for any element type; a non-finite entry returns `Inf` or `NaN`. The start vector is seeded, so repeated calls agree; `ones` would lie in the null space of every graph Laplacian. The new `opnorm` docstring states that the result is an estimate, computed at `Float64` accuracy, and that tightly clustered spectra can need about `n` iterations.

On nightly the result matches `opnorm(Array(A))` to 1e-15 for real and complex, square and rectangular matrices, and is correct for `spdiagm([ones(9999); 1 + 1e-6])`, for entries of `1e-307`, and for `BigFloat` entries of `1e±400`. A 1-D Laplacian with n = 8000 takes 2.3 s and a 90000×90000 2-D Laplacian 0.8 s. `Pkg.test(test_args=["linalg"])` passes.

Left out: `opnorm` of adjoints, transposes and views of sparse matrices, a follow-up touching dispatch, and full `BigFloat` precision. Not backportable, since it is new behaviour.


ViralBShah:
@dkarrasch Would be great for your review on this - I don't want to merge this one without review.

PR #736, opened 2026-07-16
use normal entrypoints into base's display pipeline

rokke-git:
Base and SparseArrays had some funky action-at-a-distance stuff going on, this fully disentangles the printing functions. this is what I really came over here for, the type thing was just me hoping to simplify things a bit. didn't end up effecting much, but does add the `circular_reference` line.

it also makes SparseArrays a little smarter about when to swap to braille output, and actually uses the whole screen instead of just the left half. it's also theoretically doing less work now, but it's just a printing function so whatever. the width of displayed zeros also doesn't depend on the type anymore, and is always centered.

since this removes the error that used to come up during printing, I added in a new warning to replace it.

fixes #21, fixes #22, fixes #233, fixes #420, fixes #618

PR #863, opened 2026-09-19
Scale the braille `show` of sparse matrices to `displaysize` only when `:limit` is set

### LINKED ISSUE #233: don't use displaysize in show unless limit=true is set
It was [pointed out on discourse](https://discourse.julialang.org/t/show-and-displaysize/86126/5?u=stevengj) that the Braille display of sparse matrices unconditionally limits the output to `displaysize` here: https://github.com/JuliaSparse/SparseArrays.jl/blob/3b1ce54fe4575913140cc4f3b001d456da6091e9/src/sparsematrix.jl#L304-L305

Arguably, this should be conditional on whether `get(io, :limit, false) == true`, something like:
```jl
if get(io, :limit, false)
    maxHeight = displaysize(io)[1] - 4 # -4 from [Prompt, header, newline after elements, new prompt]
    maxWidth = displaysize(io)[2] ÷ 2
else
    maxHeight = maxWidth = typemax(Int)
end
```

ViralBShah:
Fixes #233.

```julia
julia> using SparseArrays

julia> S = sparse(trues(200, 16));

julia> count(==('\n'), repr("text/plain", S))  # 20 on main (squeezed into the default 24×80 displaysize), 50 with this PR
```

`_show_with_braille_patterns` already consulted `:limit`, but defaulted it to `true`, so an `IOContext` without the key (a plain `IOBuffer`, `repr("text/plain", S)`) was still scaled down to `displaysize`. The default is now `false`, matching Base and the `SparseVector` `show`. The REPL, Documenter and doctests set `:limit => true`, so their output is unchanged. I did not use the `typemax(Int)` form suggested in the issue, because `4maxHeight` would overflow in the existing guard.

Behaviour change, so not for backport: without `:limit` the braille grid takes about `m*n/4` bytes, so `repr("text/plain", S)` of a 10⁵×10⁵ matrix now allocates roughly 2.5 GB instead of printing a thumbnail. Dense arrays print in full the same way. #736 touches the same display code and may overlap.

Tested on nightly: `sparsematrix_constructors_indexing` passes, and the new test gives 5 lines with the fix and 2 on `main`. Full suite not run locally.


ViralBShah:
This one is a smaller PR - but ideally is closed in favour of getting #736 through.

PR #865, opened 2026-09-19
Take the same dense factors on either side of sparse-dense `*` and `mul!`

ViralBShah:
Follow-up to #861 and #864. The sparse-dense kernels were reached through different hooks with different type restrictions depending on which side the sparse factor was on, and anything outside `StridedMatrix`/`BitMatrix` fell to LinearAlgebra's generic product, which sweeps the sparse factor with O(n²) `getindex` calls.

```julia
S = sprand(3000, 3000, 2/3000); D = rand(3000, 3000); T = rand(3000, 5)
S * view(T, :, [1, 2, 3, 4])             # non-strided dense factor
S * (1.0:3000)
S * UpperHessenberg(D)                   # 31 s
Symmetric(D) * Hermitian(complex.(S))    # 6 s
```

- Both orders now use the same hook, the 7-argument `mul!(C, tA, tB, A, B, α, β)` that `generic_matmatmul_wrapper!` falls back to, with `spdensemul!`/`densespmul!` as mirror images. The two `generic_matmatmul_wrapper!` methods for dense × sparse are gone.
- The dense factor is `AbstractMatrix` (`AbstractVector` for mat-vec) on both sides. It arrives with its `Adjoint`/`Transpose`/`Symmetric`/`Hermitian` wrapper already stripped, and structured types with their own product (`Diagonal`, banded, triangular) are intercepted by LinearAlgebra's `_mul!` methods before this hook, so what is left is non-strided arrays, ranges, reshapes and `UpperHessenberg`. The kernels only need scalar `getindex`. `DenseInputVector` had no other user and is removed.
- `LinearAlgebra.mul(::HermOrSym, ::HermOrSym)` materializes its second factor with `copyto!(similar(parent(B)), B)`, which for a sparse `B` is an elementwise `setindex!` loop. Since the kernels take both wrappers as they are, dense-sym × sparse-sym (both orders) goes straight to `_mul`.

This touches dispatch. Widening needs two disambiguating methods: sparse × sparse into a strided `C` (keeps the generic product it had), and `mul!(::StridedVector, tA, ::AbstractSparseMatrixCSC, ::AbstractSparseVector, …)`, for which the existing body moved to `_spmatspvecmul!`. Aqua/ambiguity job passes.

Nightly, n = 3000, time and allocations:

| product | main | PR |
|---|---|---|
| `view(W, idx, :) * S` (4 rows) | 18.3 ms, 0.09 MiB | 0.048 ms, 0.09 MiB |
| `S * view(T, :, idx)` (4 cols) | 76.4 ms, 0.09 MiB | 0.038 ms, 0.09 MiB |
| `S * view(v, idx)` | 26.1 ms | 0.040 ms |
| `S * (1.0:n)` | 26.6 ms | 0.056 ms |
| `Symmetric(S) * reshape(1.0:4n, n, 4)` | 121 ms | 0.081 ms |
| `S * UpperHessenberg(D)` | 31200 ms, 68.7 MiB | 42.9 ms, 68.7 MiB |
| `UpperHessenberg(D) * S` | 51.0 ms | 14.0 ms |
| `Symmetric(D) * Hermitian(Sc)` | 6250 ms, 563 MiB | 53.5 ms, 137 MiB |
| `Hermitian(Sc) * Symmetric(D)` | 75.2 ms, 206 MiB | 62.7 ms, 137 MiB |

Tests: each sparse factor (plain, adjoint, column view, `Symmetric`, `Hermitian`) against each new dense factor on both sides, `*` and 5-arg `mul!`, real and complex, swept against dense; sparse × sparse into a dense `C`. Dispatch is proven with stored zeros (kernels give `Inf * 0 = NaN`, the generic fallbacks skip zeros); all five proofs fail on main. Full `Pkg.test()` passes on nightly.

Left out: a dense factor times a *triangular* or `UpperHessenberg` sparse matrix, and row-sliced sparse views, still take generic paths (and `Matrix * LowerTriangular(S)'` returns a sparse matrix where the mirrored product returns dense). Those need a `generic_mattrimul!` kernel mirroring the existing `generic_trimatmul!` and are a separate change.


dkarrasch:
It would be great if `generic_matmatmul_wrapper!` was gone. I "accidentally" renamed it recently in LinearAlgebra, and it caused issues over here. This might have been the only usage of this internal function outside of `LinearAlgebra`.

ViralBShah:
I will try to see what it takes to remove `generic_matmatmul_wrapper` in the set of PRs coming up.

PR #866, opened 2026-09-19
Specialize `dot` for column views of sparse matrices

ViralBShah:
Every `dot` kernel is typed on `AbstractSparseMatrixCSC`, which a `SubArray` is not, so a column view of a sparse matrix falls to LinearAlgebra's elementwise loop in all of the Frobenius, three-argument and `Symmetric`/`Hermitian` forms.

```julia
A = sprand(5000, 5100, 2e-3); V = view(A, :, 1:5000); x = rand(5000)
dot(V, V); dot(V', V); dot(x, V, x); dot(x, Symmetric(V), x)
```

The kernels only need `rowvals`, `nonzeros` and the per-column bounds, all of which column views (ranges and arbitrary subsets) already provide into the parent's storage, so the signatures widen to `SparseMatrixCSCOrColumnSubset`. `getcolptr` is not defined for arbitrary column subsets, so `_dot_walk` and `_dot_transposed_walk` read the bounds through small `_colbounds`/`_colstarts`/`_colstops` helpers; they avoid `nzrange` and a view of `colptr` because both cost 5-10% on plain matrices. This touches dispatch: the view methods for the two-argument forms are spelled out per operand so each stays more specific than the dense-operand methods, and `detect_ambiguities` reports nothing new for `dot`.

Minimum times in ms on nightly 1.14.0-DEV.3300 for the sizes above, 0 allocations for `V ⋅ W` before and after; plain `P ⋅ Q` and `P' ⋅ Q` are unchanged within 1%:

| | `V ⋅ W` | `V' ⋅ W` | `dot(x, V, x)` | `dot(x, Symmetric(V), x)` |
|---|---|---|---|---|
| main | 169 | 421 | 78 | 40 |
| PR | 0.32 | 0.52 | 0.040 | 0.047 |

Tests compare against dense for range and permuted-subset views with complex entries and stored zeros, and count multiplications to prove each kernel is reached; `linalg` and `linalg_products` pass. A local sweep of 7200 random cases against dense found no mismatches. Views that also restrict rows are not sparse view types and stay on the fallback.

Written by Claude Code.


IanButterworth:
xref https://github.com/JuliaLang/julia/pull/63297#issuecomment-5767981554

PR #870, opened 2026-09-20
Add a cacheless `getindex` kernel for sorted row vectors with few columns or hypersparse matrices

### LINKED ISSUE #12: Avoid large cache array in getindex_I_sorted_bsearch_I
https://github.com/JuliaLang/julia/pull/42647 improves sparse matrix slicing when the matrix isn't too sparse and the number of requested indices isn't too large relative to the average number of non-zeros in a column (the linear search case). However, the overhead from a large cache vector is also present in the method used when the number of requested indices is large relative to the average number of nonzeros, (the binary search in the index vector case, i.e. `getindex_I_sorted_bsearch_I`). The problematic allocation is https://github.com/JuliaLang/julia/blob/7c5e28c4c1ff4bf5c3c9f38450158833a366b4f6/stdlib/SparseArrays/src/sparsematrix.jl#L2438-L2440 and it would be nice to get rid of. In https://github.com/JuliaLang/julia/pull/42647, it was possible to show a 1000000x speedup in a simple case and I suspect something similar is possible here.

ViralBShah:
Fixes #12. `A[I, J]` with a sorted vector `I` allocates and sweeps a cache of length `size(A, 1)` in `getindex_I_sorted_bsearch_I` and `getindex_I_sorted_linear`, however little is requested, so a loop over columns costs O(n·m) instead of O(nnz). JuliaLang/julia#42647 addressed the linear kernel but was closed unmerged when the stdlib moved.

```julia
julia> A = spdiagm(ones(10^7)); I = collect(1:258);

julia> @time A[I, 1];   # 7 ms, 76 MiB before; 0.2 μs, 304 bytes after
```

The cache is what makes these kernels fast when many columns are selected (each distinct row is located in `I` once rather than once per stored entry; removing it outright is 10x+ slower for `S[I, :]`), so it stays. This adds `getindex_I_sorted_nocache`, a two-pass count-and-fill kernel with no scratch space that locates each stored row in `I` by galloping from the previous hit (short brackets finish with a linear scan), so a search costs the log of the distance moved rather than of `length(I)` and one kernel serves both the bsearch and the linear regime. `getindex_I_sorted` picks it when `32 * nnzJ * log2(gap) < m`, with `gap` the length of `I` per stored entry of an average selected column; `nnzJ` is read off `colptr` in O(nJ). The factor is set at the cache's best case, every selected column storing the same rows with `I = 1:m` for m up to 5e7, where one search per distinct row is reused by all columns: just under the gate the cacheless kernel takes 0.4–0.55x the cached time there and 0.03–0.8x on `sprand`, so it is not slower than the current code on any input I could construct. It builds the same result with no scratch, so memory never goes up. On a 1000x1000 9-point Laplacian, 5-point and 3D 7-point Laplacians, tridiagonal, banded, arrowhead and block-diagonal matrices (single columns, red-black splits, subdomain blocks, random subsets), every call above 1 ms keeps its current kernel and timing, and few-column calls drop from 0.07–0.9 ms to microseconds. The choice between the two cached kernels is unchanged.

The last commit covers the third kernel. `getindex_I_sorted_bsearch_A`, picked when `m > nnz(A)` or when the columns are much longer than `I`, walks `I` for every column, so hypersparse matrices pay O(nJ·nI). The cacheless kernel now takes over there when `nI > 16 + avgJ ÷ 16`, `avgJ` being the average stored entries of the selected columns, and skips each column's entries above `I[1]` with one search. The threshold comes from a sweep over columns of 1 to 1e4 entries and `nI` from 0.01x to 100x that, with `I` spread out or a block at the top, middle or bottom: above it the cacheless kernel takes 0.02–1.1x the `bsearch_A` time, below it `bsearch_A` is kept (it wins by up to 8x on a 5-row `I` against long columns).

| nightly 1.14.0-DEV, M-series | before | after |
|---|---|---|
| diag 1e7: `A[[1], 1]` | 0.64 ms, 76 MiB | 0.0001 ms, 304 B |
| diag 1e7: `A[collect(1:258), 1]` | 7.0 ms, 76 MiB | 0.0002 ms, 304 B |
| diag 1e7: `A[collect(1:10^6), collect(1:1000)]` | 6.9 ms, 76 MiB | 0.46 ms, 24 KiB |
| diag 1e7: `A[I, j]` for `j = 1:1000` | 3198 ms, 74.5 GiB | 0.20 ms, 258 KiB |
| sprand 1e5, d=1e-4: `S[1:2:end, [7]]` (vector `I`) | 0.12 ms, 784 KiB | 0.02 ms, 352 B |
| sprand 1e5, d=1e-4: `S[1:2:end, :]` (vector `I`, cached path) | 14.0 ms, 9.2 MiB | 14.3 ms, 9.2 MiB |
| 9-point Laplacian 1e6: `A[R, R[k:k+19]]`, R random N/10 | 0.67 ms | 0.004 ms |
| 9-point Laplacian 1e6: `A[r, r]`, r random N/1000 | 0.90 ms | 0.11 ms |

| `m > nnz(A)` branch | before | after |
|---|---|---|
| hypersparse 1e6 (nnz 5e5): `H[collect(1:5e5), 1:100]` | 23.1 ms | 0.23 ms |
| hypersparse 1e6: `H[collect(1:5e5), 1:10^4]` | 1970 ms | 1.1 ms |
| hypersparse 1e6: `H[R, :]`, R 1e5 random rows | 27337 ms | 77 ms |
| hypersparse 1e6: `H[[i], :]` (short `I`, still `bsearch_A`) | 11.4 ms | 11.1 ms |
| tall 1e7x100, 1e4 per column: `T[collect(1e6:2e6), :]` | 137 ms | 14.5 ms |
| tall: `T[I, :]`, 5 rows (still `bsearch_A`) | 0.08 ms | 0.08 ms |

Tests: the new kernel runs in the existing `test_getindex_algs` grid against dense (stored zeros, duplicates and unsorted `I`/`J`), plus an allocation bound on a 10^6-row matrix showing scalar, single and few-column `J` reach it in both regimes. For the `bsearch_A` branch, where allocations do not tell the kernels apart, an `I` that counts its reads shows `I` is no longer walked per column (201 reads per element on main, 3.8 here). Full suite passes on nightly. No dispatch or API change; performance only, so no backport.


ViralBShah:
@andreasnoack This is a really old one. Please do chime in (even after merging) with thoughts if any.

PR #883, opened 2026-09-21
Make the `Base.USE_GPL_LIBS` conditionals `@static`

ViralBShah:
`Base.USE_GPL_LIBS` is a build-time constant, and the solver test files already branch on it with `@static if`. The remaining plain `if Base.USE_GPL_LIBS` conditionals now do the same: the solver `include`s in `src/SparseArrays.jl`, the inventory in `test/runtests.jl`, and the two in `test/threads_suite.jl`. No behaviour change; the untaken branch is simply never lowered.

The plain conditionals in `test/fixed.jl` and `test/linalg.jl` are left alone here because #882 moves those testsets into `test/solvers/solvers.jl`, which is already under an `@static` guard. This PR and #882 touch neighbouring lines in `test/runtests.jl` and the renamed `threads_suite.jl`, so whichever merges second needs a small rebase.

Tested on nightly (1.14.0-DEV): `Pkg.test(test_args=["fixed", "linalg", "threads_suite", "spqr"])` passes.

Written with Claude Code.


ViralBShah:
Folded into #882, which now carries this change.

PR #884, opened 2026-09-21
Make the last write win for repeated indices in nonscalar `setindex!`

### LINKED ISSUE #811: Repeated index in `A[I, J] = B` corrupts the matrix (duplicate row) or drops values (duplicate column)
When assigning to a repeated index, a `SparseMatrixCSC` instead stores multiple assigned values. The last write should win, but with a sparse matrix the assignment corrupts the matrix:

```julia
julia> using SparseArrays

julia> A = spzeros(2, 2);

julia> A[[1, 1], 1] = [5.0, 6.0];

julia> Matrix(A) 
2×2 Matrix{Float64}:
 6.0  0.0
 0.0  0.0

julia> A # incorrect value and number of stored entries
2×2 SparseMatrixCSC{Float64, Int64} with 2 stored entries:
 5.0   ⋅
  ⋅    ⋅

julia> sum(A) # incorrect sum
11.0

julia> sum(Matrix(A))
6.0

julia> B = zeros(2, 2);

julia> B = zeros(2, 2);^C

julia> B[[1, 1], 1] = [5.0, 6.0];

julia> B
2×2 Matrix{Float64}:
 6.0  0.0
 0.0  0.0
```

## Possible cause (AI-diagnosed)

`setindex!(A::AbstractSparseMatrixCSC, V::AbstractVecOrMat, I, J)` in `sparsematrix.jl` sorts `I` and `J` but assumes they contain no repeats. In the column merge

```julia
rowB = I[rowvalB[ptrB]]          # two rows of B map to the same row of A: both are emitted
...
colptrA[col+1] = ptrS
colB += 1                        # one column of B per distinct column of A: later duplicates in J are never visited
```

and the final `_checkbuffers(A)` only checks buffer lengths. A minimal fix is to fall back to the generic element-wise loop when `!allunique(I) || !allunique(J)` (which gives the `Matrix` semantics), or to throw an `ArgumentError` for repeated indices.

ViralBShah:
Fixes #811. Assigning through a repeated index corrupted a `SparseMatrixCSC`: a repeated row stored the same row twice in a column, and a repeated column dropped values. Dense arrays let the last write win.

```julia
julia> A = spzeros(2, 2); A[[1, 1], 1] = [5.0, 6.0]; A
2×2 SparseMatrixCSC{Float64, Int64} with 2 stored entries:   # now 1 stored entry, 6.0
 5.0   ⋅
  ⋅    ⋅

julia> sum(A), sum(Matrix(A))
(11.0, 6.0)                                                  # now (6.0, 6.0)
```

`setindex!(A, V, I, J)` sorts `I` and `J` and merges `V` into each column, assuming the sorted indices are strictly increasing. A new helper, `_setindex_lastwrites`, returns the stable sort permutation with all but the last position of each repeated index removed, so `V[pI, pJ]` holds exactly the writes that survive; it returns `nothing` when the index is already strictly increasing, which leaves the common path with one extra `issorted` pass. The linear-index method `setindex!(A, x, I)` already tried to skip repeats but compared `I[S[k]]` with its neighbour in the original order rather than the sorted one, so `A[[1, 2, 1]] = x` also stored a duplicate; it now compares with `I[S[k+1]]`. The scalar-fill, logical-mask and `SparseVector` methods were already correct.

Found by sweeping 3000 random `A[I, J] = V`, `A[I, j] = v`, `A[L] = x`, `A[I, J] .= c` and `v[L] = x` assignments with repeated, sorted and unsorted indices against `Matrix`/`Vector`: about two thirds of the matrix cases and a quarter of the linear ones disagreed before, none after. The new tests in `test/indexing.jl` fail on `main`. Tested with the `indexing`, `sparsematrix`, `fixed`, `issues` and `higherorderfns` suites on nightly (1.14).

No dispatch or public API changes. A bug fix, so a candidate for `backport 1.x`.


ViralBShah:
Timing against `main`, 20000×20000 `A` with density 1e-3, 5000×5000 (or 20000×5000) sparse `V`, best of 15 on nightly, time in ms / allocations in MiB:

| case | `main` | this PR |
|---|---|---|
| `A[Is, Js] = V`, sorted vectors | 5.47 / 16.56 | 5.33 / 16.55 |
| `A[1:n, Js] = V` | 2.06 / 17.69 | 2.04 / 17.69 |
| `A[:, 1:k] = V` | 1.92 / 17.69 | 1.98 / 17.69 |
| `A[n:-1:1, Js] = V` | 2.84 / 19.25 | 2.71 / 19.25 |
| `A[Iu, Ju] = V`, unsorted vectors | 2.75 / 17.46 | 2.80 / 17.46 |
| `A[Iu, Js] = V` | 2.52 / 17.36 | 2.54 / 17.35 |
| 10×10 block, sorted | 0.86 / 16.13 | 0.85 / 16.14 |
| `A[L] = x`, 200000 unsorted linear indices | 15.63 / 20.40 | 15.35 / 20.40 |

The first version converted the range permutation of a decreasing range to a `Vector`, which cost that case about 30% (3.65 ms, 20.47 MiB); the second commit keeps the range, as `main` did.

PR #885, opened 2026-09-21
Define the dense-operand `dot` methods one per type to cut load time

IanButterworth:
@ViralBShah I'm not sure if this is 1) over-optimization 2) something that should be fixed in base 3) just reasonably slower and shouldn't worry about it.

Claude:

---

Since #866, `using SparseArrays` takes ~80 ms longer on nightly (0.143 s → 0.233 s here), which JuliaLang/julia#63297 carried into Base and shows up as +0.13 s `load` on every SparseArrays dependent in the Julia TTFX CI. It is not invalidations; the time is method insertion during package load, `jl_method_table_activate` computing type intersections:

```julia
using SparseArrays: DenseMatrixUnion, MatrixWrappersOrView, AbstractSparseMatrixCSC, SparseMatrixCSCOrColumnSubset
const LHS = Union{DenseMatrixUnion,MatrixWrappersOrView{<:Any,<:Union{DenseMatrixUnion,AbstractSparseMatrix}}}
@time @eval LinearAlgebra.dot(A::LHS, B::AbstractSparseMatrixCSC) = 1         # 0.6 ms
@time @eval LinearAlgebra.dot(A::LHS, B::SparseMatrixCSCOrColumnSubset) = 1   # 84 ms
```

Most of the 84 ms is `typeintersect`/`morespecific` against LinearAlgebra's `dot(x::StridedVecLike{T}, y::StridedVecLike{T}) where T<:BlasReal/BlasComplex`: a `Union` of `SubArray`s on the left times a `SubArray` union on the right. Defining the dense-operand methods one per union member instead makes each insert in 0.4–7 ms. `StridedMatrix` and the `SubArray` member overlap (strided views of dense matrices), so their `typeintersect` gets its own method to keep dispatch unambiguous; `detect_ambiguities` is clean. The mirror methods keep re-dispatching through `dot(B, A)` so wrapped sparse operands still reach the sparse kernels. `using SparseArrays` on nightly 1.14.0-DEV.3309, ABBA x3: main 0.231–0.238 s, this PR 0.151–0.152 s. The #866 benchmarks (`V ⋅ W`, `V' ⋅ W`, `dot(x, V, x)`, `dot(x, Symmetric(V), x)`) and the dense-operand paths (`M ⋅ V`, `view(M, :, r) ⋅ V`, `Symmetric(M) ⋅ V`, `M ⋅ P`) are unchanged within noise, same allocations.

Tested on nightly: `ambiguous`, `linalg`, `matmul`, `sparsematrix` and `issues` pass. Left out: the other #866 signatures cost ~8 ms together, not worth splitting. This unblocks the Base TTFX regression, so a bump after merge would be good.

Written with Claude Code (Opus 5); the human author has not yet read all of it.


ViralBShah:
I have been working through lots and lots of combinations of operations involving adjoints, views, symmetric, which have added lots of methods in SparseArrays. So, definitely there may be patterns that are overboard here.

Is it possible to add a CI test to keep a lid on load times? And should AGENTS.md be updated to give it the right guidance?

I fully expect to walk back or correct some of these things. So far I have mostly been focussed on running correctness tests on downstream packages.

IanButterworth:
> Is it possible to add a CI test to keep a lid on load times?

That's what the TTFX CI job does. On PRs that touch /base /Compiler it runs to detect regressions before merge, and on every master commit benchmarks. [https::/perf.julialang.org/ttfx](http://3.82.159.74/?tab=ci-ttfx&tt=7). You can see the uptick on the bump PR.

But for individual repos.. I'd just ask claude to make a little CI script with ABBA repeats

ViralBShah:
In general, I found `using SparseArrays` is quick enough that this wasn't an issue I even considered. Let's merge any savings we find, and then accept the rest - since we are adding a lot of functionality.

IanButterworth:
https://github.com/JuliaLang/julia/pull/63306 seems like the proper fix. Can reopen if it gets rejected

